Issue #7, opened 2021-10-25
Bob's claim transaction can be frontrun by anyone


alxs:
Anyone can see Bob's secret in the mempool and nothing prevents them from calling claim instead.

The most straightforward way to prevent this would be for Bob to provide the address from which he'll call the contract along with his Monero keys and for Alice to include it in the contract I think.

Then we can restrict claim to only be callable by Bob.

PR #15, opened 2021-10-27
Prevent Bob being frontrun

Lederstrumpf:
Swap is instantiated with Bob's ETH address and msg.sender checked in `claim` call to match said address to prevent Bob being frontrun.

Addresses #7.


alxs:
@Lederstrumpf is this still a draft?

noot:
hey @Lederstrumpf going to close this for now and I think it was already done

PR #27, opened 2021-11-17
Solidity contracts + testing revamp

alxs:
The main goal of this was to abstract the specific cryptographic implementation for the swap from the swap logic itself in order to minimise code duplication, which is now the case. `Swap.sol` is an abstract contract that can be instantiated needing only a `verifySecret` function that contains the verification logic.

In between a bunch of other stuff happened and in this PR are now also:
- consistent error reporting
- testing logic in the smart contracts is now also independent of the impl files
- scalable hardhat tests
- simplified crypto - at least for secp256k1 now only the bit that actually matters is included

I was halfway through doing this when I realised the DLEQ approach isn't actually integrated with the rest of the code - that is still the idea though, no? It's way faster, works in principle and I for one would like to play around with @Lederstrumpf's stuff.

TODO:
- propagate new error messages to Go tests
- rename all instances of `Swap.sol` to `SwapOnChain.sol` in Go code
- run integration tests (the new script is awesome, thanks :D)
- integrate the DLEQ option into the actual program (not for this PR...)

Other than that, here are some preliminary gas measurements for the two approaches:
```
·----------------------------------------|----------------------------|-------------|-----------------------------·
|          Solc version: 0.8.5           ·  Optimizer enabled: false  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·········································|····························|·············|······························
|  Methods                                                                                                        │
····················|····················|··············|·············|·············|···············|··············
|  Contract         ·  Method            ·  Min         ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
····················|····················|··············|·············|·············|···············|··············
|  SwapDLEQMock     ·  testVerifySecret  ·       30304  ·      30316  ·      30310  ·            2  ·          -  │
····················|····················|··············|·············|·············|···············|··············
|  SwapOnChainMock  ·  testVerifySecret  ·      930673  ·     950395  ·     938818  ·            5  ·          -  │
····················|····················|··············|·············|·············|···············|··············
|  Deployments                           ·                                          ·  % of limit   ·             │
·········································|··············|·············|·············|···············|··············
|  SwapDLEQMock                          ·           -  ·          -  ·    1350801  ·        4.5 %  ·          -  │
·········································|··············|·············|·············|···············|··············
|  SwapOnChainMock                       ·           -  ·          -  ·    1855645  ·        6.2 %  ·          -  │
·----------------------------------------|--------------|-------------|-------------|---------------|-------------·
```

So that'd indeed be roughly a 30x improvement.


noot:
> rename all instances of Swap.sol to SwapOnChain.sol in Go code
I would rather not do this, as I find it confusing.. the contract is always on chain :p

> I was halfway through doing this when I realised the DLEQ approach isn't actually integrated with the rest of the code - that is still the idea though, no? It's way faster, works in principle and I for one would like to play around with @Lederstrumpf's stuff.
yep, the gas benefits look pretty significant, so it would be nice to add. the contract deployment is still a lot, but hopefully #28 will address that

alxs:
Feel free to suggest another name :)

Personally I find it confusing to call one Swap and the other SwapDLEQ, giving the impression that the first is the default. IMO both should be treated equally.

Another argument is that now `Swap.sol` contains the abstract swap logic, and the other two the crypto. We can also just call them by the curve names.

noot:
hey @alxs I made some protocol updates in this PR https://github.com/noot/atomic-swap/pull/37 and it seems to improve the gas a lot - feel free to take a look and let me know what you think!

alxs:
@noot thanks for the ping, one of Farcaster's team members and I left some comments on the associated issue (#36). I just reopened it.

noot:
@alxs just responded, I changed back the contract, I got a bit too excited with gas optimizations haha. let's continue to go down the DLEQ route, let me know if you need any help with the go code.

in terms of contract naming, I think there should be always 1 contract called Swap.sol which is the canonical contract that's currently used by the protocol. then, if we decide to move to the DLEQ version as the main contract, then we can replace the current Swap.sol with the DLEQ version. that way we won't have to rename a bunch of things when we switch the contracts over, since if the API is the same, it should be pretty straightforward.

alxs:
I mean, it would have been nice haha. Thanks, yeah alright I'll give it a go.

The API is not exactly the same - in Robert's version you pass the x and y coordinates of the public key instead of the encoded point, with my abstraction it's the hash of the ABI-encoded coordinates. Not a major change, but just wanted to point it out.

In any case, I actually thought we would offer both options. Maybe you see where I was coming from with the renaming now. Though of course if we just drop the classic one there's no point in any renaming in the first place. And if both are equally secure I guess there's no point in maintaining one that is 30x more expensive. What do you think, shall we just switch?

noot:
@alxs sorry for the late reply! yeah, I think we can just switch over when the DLEq version is complete, that'll be cleaner to maintain.

alxs:
Closing from my side, feel free to steal it like the rest of my work.

PR #81, opened 2022-01-20
add CORS support

noot:
@Tbaut pls try and let me know if this works!


Tbaut (approved):
:rocket: :rocket: :rocket: :rocket:

PR #87, opened 2022-01-31
Some UI to get started

Tbaut:
I use yarn, but it should work with npm.
When launched, the UI will talk to a node listening on `5001` and query the offers for all its peers

note that the tooltip showing the peer address truncated seems to be an issue with the hot reloading server I'm using.. so I'd ignore this for now :D

https://user-images.githubusercontent.com/33178835/151891075-aacd7ee6-37a8-45d5-831e-517a52e5122c.mp4


noot (approved):
works perfectly for me! amazing :D

PR #91, opened 2022-02-10
Swaping modal

Tbaut:
one can take an offer by clicking on the take button. The user can then type in a value. They will be warned if the amount requested is too high or too low.
Once a swap is executed, the list of offers is refreshed, because the offer that was just taken isn't available any more.

It looks like this:

https://user-images.githubusercontent.com/33178835/153770547-b13d3918-20a9-48a2-adde-12dcc5cd897b.mp4


scripts/setup-devnet.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# open Bob's wallet (must have funds)
+# sleep 5
+# curl http://localhost:18083/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"open_wallet","params":{"filename":"stagenet-wallet","password":""}}' -H 'Content-Type: application/json'
+
+# check balance
+# curl http://localhost:18083/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":get_accounts","params":{}}' -H 'Content-Type: application/json'
+
+# start mining (if synced)
+# curl http://localhost:18083/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":start_mining","params":{"threads_count":1,"do_background_mining":true,"ignore_battery":false}}' -H 'Content-Type: application/json'
```
Tbaut, line 51:
```suggestion
```

noot (approved):
looks good to me! my only suggestion is to add an error message if the swap fails for whatever reason, i noticed that if the swap fails the UI currently doesn't log anything.

PR #99, opened 2022-02-28
Devnet setup script and showing errors in the UI

Tbaut:
If you want to try it out quickly with some errors, you can comment one part of the UI and send a too large offer. See my comment on Github

https://user-images.githubusercontent.com/33178835/156044306-0d005cea-6e91-43af-bfde-684308d4cfa3.mp4


Tbaut:
I can't comment on Github because I didn't touch those lines, but you should comment out this section for instance:
```
} else if (
    willReceive !== 0 &&
    $selectedOffer &&
    willReceive > $selectedOffer.maxAmount
  ) {
    error = `The amount of ${getCorrespondingToken(
      $selectedOffer.provides
    )} to swap is too high`
```

so that you are able to use the UI to send too large amount.

ui/src/components/TakeDealDialog.svelte
```diff
@@ -55,25 +54,31 @@
         if (result?.status === 'success') {
           isSuccess = true
           getPeers()
+        } else if (result?.status === 'aborted') {
```
noot, line 57: could you add the `refunded` case as well?

scripts/setup-devnet.sh
```diff
@@ -1,35 +1,52 @@
 #!/bin/bash
 
+# useful dir relative to this script
 MONERO_DIR="../monero-x86_64-linux-gnu-v0.17.3.0"
+LOG_DIR="../log"
```
noot, line 5: could you make this `$TMPDIR/swapd_log` or something like that? so it automatically goes into the temp directory?

ui/src/components/TakeDealDialog.svelte
noot: yeah, it happens if the swap fails after funds get locked. maybe you can say something like "swap funds refunded, please check logs for more info"

Tbaut: done

scripts/setup-devnet.sh
Tbaut: found an elegant way with TMPDIR and a default `/tmp` in case nothing's set

scripts/setup-devnet.sh
```diff
@@ -1,35 +1,53 @@
...
 
-# bash ./scripts/install-monero-linux.sh
-echo "========== start ganache-cli"
-ganache-cli -d &
+echo "cleanup"
+pkill -e -f monero;
+pkill -e -f ganache-cli;
+killall -v swapd;
+pkill -e -f swapcli;
+rm -rf $LOG_DIR/*;
```
Tbaut, line 14: No need to rm the logs
```suggestion
```

noot (approved):
looks good to me!

PR #103, opened 2022-03-20
Use WS to take an offer

Tbaut:


Tbaut:
done in https://github.com/noot/atomic-swap/pull/126

PR #129, opened 2022-06-10
Metamask store cleanup and bring back success message

Tbaut:
- connects to MM automatically if you've ever been connected
- only shows the connect button if not connected
- some cleanup in the MM store
- Shows the success message if the swap was successful


noot (approved):
works for me!

PR #135, opened 2022-06-16
Test cleanup

dimalinux:
Primary changes are:
* Fixes race conditions in unit tests where tests in different packages were concurrently producing transactions with the same Ganache test key (most common error was a nonce mismatch).
* Fixes race conditions were unit tests were concurrently using monero-wallet-rpc services which can only have one wallet open at a time
* Fixes race conditions from different tests sharing the same test file names in `/tmp`
* Removes websocket connection leaks by closing the ganache connection when the test exits
* Provided a fixed path to the Monero install directory so numerous scripts will not need to be updated on every release
* `monerod` and `monero-wallet-rpc` now bind to `127.0.0.1` instead of `0.0.0.0` in the tests (ganache-cli already did this by default)
* Minor fixes for the lint target when GOPATH is not defined
* Renamed `install_lint.sh` to `install-lint.sh` to make the naming consistent with the other scripts

If you want anything done differently, additional changes, something reverted, smaller more targeted PRs or whatever, just let me know.

UPDATE:  Unit tests and integration tests are now working on my 12-core host. There are still some flaky tests that sometimes fail, but the problems are probably not related to tests running concurrently.


Makefile
```diff
@@ -1,8 +1,10 @@
 .PHONY: lint test install build build-dleq mock
 all: build-dleq install
 
+GOPATH ?= $(shell go env GOPATH)
```
dimalinux, line 4: This will only set the GOPATH if it was not already set

common/consts.go
```diff
@@ -10,9 +10,25 @@ const (
...
+	// (0) 0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
+	// (1) 0x6cbed15c793ce57650b9877cf6fa156fbef513c4e6134f022a85b1ffdd59b2a1
+	// (2) 0x6370fd033278c143179d81c5526140625662b8daa446c22ee2d73db3707e620c
+	// (3) 0x646f1ce2fdad0e6deeeb5c7e8e5543bdde65e86029e2fd9fc169899c440a7913
+	// (4) 0xadd53f9a7e588d003326d1cbf9e4a43c061aadd9bc938c843a79e7b4fd2ad743
+	// (5) 0x395df67f0c2d2d9fe1ad08d1bc8b6627011959b79c53d7dd6a3536a33ab8a4fd
+	// (6) 0xe485d098507f54e7733a205420dfddbe58db035fa577fc294ebd14db90767a52
+	// (7) 0xa453611d9419d0e56f499079478fd72c37b251a94bfde4d19872c44cf65386e3
+	// (8) 0x829e924fdf021ba3dbbc4225edfece9aca04b929d6e75613329ca6f1d31c0bb4
+	// (9) 0xb0057716d5917badaf911b193b12b910811c1497b5bada8d7711f758981c3773
```
dimalinux, line 24: I added all the keys in the comments, as new tests may require additional keys and it makes it more clear what the indices are.  I changed the `ganache-cli` invocations to use the long flag (`--deterministic`), because most daemons that take a `-d` or `-D` flag use it to mean `--detach` or `--daemonize` making the long form more clear in scripts.

I wanted to move the key constants to a *_test.go file, but they are used in the non-test swapd code.

protocol/backend/backend_test.go
```diff
@@ -14,19 +15,19 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const defaultXMRTakerAddress = "0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1"
```
dimalinux, line 17: I updated the code below to derive this address. My thinking was that it was better to have all the Ganache key related values defined in one place.

scripts/install-lint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+GOBIN="$(go env GOPATH)/bin"
```
dimalinux, line 3: I was hoping github would show the diff even though the file was renamed.  The command above works equally well for defaulted and custom GOPATH valus.

scripts/install-monero-linux.sh
```diff
@@ -1,9 +1,17 @@
 #!/bin/bash
 
-if [[ -d "./monero-x86_64-linux-gnu-v0.17.3.2" ]]; then 
-    echo "monero-x86_64-linux-gnu-v0.17.3.2 already installed"
+arch=linux64
+
+if [[ -d "monero-bin" ]]; then
+    echo "$(dirname $(realpath monero-bin)) already installed"
     exit 0
 fi
```
dimalinux, line 8: `monero-bin` is a symbol link.  All scripts now reference it instead of the arch/release dependent name.  In the test above, both the symbolic link and the directory it refers to must exist in order to take the early exit path (you can delete either one to update).

scripts/run-integration-tests.sh
```diff
@@ -1,31 +1,32 @@
...
+./monero-bin/monero-wallet-rpc --rpc-bind-ip 127.0.0.1 --rpc-bind-port 18084 --disable-rpc-login --wallet-dir ./alice-test-keys &> monero-wallet-cli-alice.log &
 MONERO_WALLET_CLI_ALICE_PID=$!
 
 # install ganache and run 
 echo "installing and starting ganache-cli..."
-if ! command -v golangci-lint &> /dev/null; then
-	npm i -g ganache-cli
+if ! command -v ganache-cli &> /dev/null; then
+	echo "Please install ganache-cli and place it in your path"
 fi
```
dimalinux, line 27: I made this same change in 3 different scripts.  The check was looking at `golangci-lint` instead of `ganache-cli` and the nmp install will fail unless you have write access to /usr/local/*.

Any opinion on how to better handle the error?  I could add an `exit` after the error message, but to we'd need start `ganache-cli` before any of the other daemons to make sure any started daemons get cleaned up.

scripts/run-unit-tests.sh
```diff
@@ -1,31 +1,32 @@
...
 
 echo "starting monero-wallet-rpc on port 18083..."
 mkdir bob-test-keys
-./monero-x86_64-linux-gnu-v0.17.3.2/monero-wallet-rpc --rpc-bind-port 18083 --disable-rpc-login --wallet-dir ./bob-test-keys &> monero-wallet-cli-bob.log &
+./monero-bin/monero-wallet-rpc --rpc-bind-ip 127.0.0.1 --rpc-bind-port 18083 --disable-rpc-login --wallet-dir ./bob-test-keys &> monero-wallet-cli-bob.log &
 MONERO_WALLET_CLI_BOB_PID=$!
 
 sleep 5
 curl http://localhost:18083/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"create_wallet","params":{"filename":"test-wallet","password":"","language":"English"}}' -H 'Content-Type: application/json'
+echo
```
dimalinux, line 16: The returned JSON isn't newline terminated, so I added an echo.  I didn't try passing `-w "\n" ` to `curl`, but that would probably work equally as well.

scripts/run-integration-tests.sh
noot: good catch! can you re-add the `npm i -g ganachi-cli` line? the CI now fails as it doesn't have ganache-cli installed. alternatively I think you can update the CI files to install ganache-cli there.

noot (commented):
looks good! only issue is CI failing due to ganache-cli not being run.

tests/alice.key
```diff
@@ -1 +0,0 @@
-b0d2d5dac3cc02d8db12e4a9703d6175cd4e9531027268b0566f50b85ca2e9410add5985d336d208fcb66c429f3abab7ae4d42f94714b54ea6fb8558d3a72df9
```
dimalinux, line 1: I removed this file from source control by accident.  Was it supposed to be here?  `*.key` is listed in the top-level `.gitignore`.

tests/ganache_test_keys.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	// returns the package and function name from the program counter
+	// example: "github.com/noot/atomic-swap/protocol/xmrtaker.newBackend"
+	packageName := minPackageName(t, runtime.FuncForPC(pc).Name())
+
+	for i, name := range testPackageNames {
+		if name == packageName {
+			return i * keysPerPackage
+		}
+	}
```
dimalinux, line 90: I'm not a huge fan of my own solution, but I couldn't come up with a simpler idea.  The original tests were written with the concept that they have access to 2 Ethereum test keys, so the best idea for how to update the code without modifying lots of tests that I don't fully understand yet, was to hand out each unique Go package 2 unique keys.

cmd/daemon/contract_test.go
```diff
@@ -13,15 +14,18 @@ import (
...
 func TestGetOrDeploySwapFactory(t *testing.T) {
-	pk, err := ethcrypto.HexToECDSA(common.DefaultPrivKeyXMRTaker)
+	pk, err := ethcrypto.HexToECDSA(tests.GetTakerTestKey(t))
 	require.NoError(t, err)
 
 	ec, err := ethclient.Dial(common.DefaultEthEndpoint)
 	require.NoError(t, err)
+	defer ec.Close()
+
+	tmpDir := t.TempDir()
```
dimalinux, line 24: The `TempDir()` function from `testing.T` creates a 2nd-level temporary directory in `/tmp` and removes it when the test is complete. The name includs the test name if you are ever debugging.

monero/client_test.go
```diff
@@ -51,13 +49,10 @@ func TestClient_Transfer(t *testing.T) {
 	kpABPub := mcrypto.SumSpendAndViewKeys(kpA.PublicKeyPair(), kpB.PublicKeyPair())
 	vkABPriv := mcrypto.SumPrivateViewKeys(kpA.ViewKey(), kpB.ViewKey())
 
-	r, err := rand.Int(rand.Reader, big.NewInt(10000))
-	require.NoError(t, err)
-
 	cXMRTaker := NewClient(common.DefaultXMRTakerMoneroEndpoint)
 
 	// generate view-only account for A+B
-	walletFP := fmt.Sprintf("test-wallet-%d", r)
```
dimalinux, line 60: @noot What does the `FP` in `walletFP` stand for?

monero/client_test.go
```diff
@@ -84,20 +79,20 @@ func TestClient_Transfer(t *testing.T) {
...
-	_ = cXMRTaker.callGenerateFromKeys(skAKPriv, vkABPriv, kpABPub.Address(common.Mainnet),
-		fmt.Sprintf("test-wallet-%d", r), "")
+	_ = cXMRTaker.callGenerateFromKeys(skAKPriv, vkABPriv, kpABPub.Address(common.Mainnet), walletFP, "")
 
 	err = cXMRTaker.refresh()
 	require.NoError(t, err)
 
 	balance, err = cXMRTaker.GetBalance(0)
 	require.NoError(t, err)
-	require.NotEqual(t, 0, balance.Balance)
```
dimalinux, line 100: I was surprised that this didn't give a mismatched types error.  I guess the test above always passes regardless of the value.  I created this example to understand it better and it passes:
```
func Test_FloatNotZero(t *testing.T) {
	var f = 0.0
	require.NotEqual(t, 0, f)
}
```

protocol/xmrmaker/swap_state_test.go
```diff
@@ -66,7 +62,7 @@ func newTestXMRMaker(t *testing.T) *Instance {
 		EthereumClient:       ec,
 		EthereumPrivateKey:   pk,
 		Environment:          common.Development,
-		ChainID:              big.NewInt(common.MainnetConfig.EthereumChainID),
```
dimalinux, line 69: The current version of `ganache` is just called `ganache` instead of `ganache-cli`.  @noot assuming you agree, I'll create a PR that puts us on the current version of ganache in a subsequent PR after the multi-core issues are all solved. The newer versions of ganache throw an error when the chain ID is incorrect.

scripts/run-integration-tests.sh
dimalinux: I added it back and I also added instructions to `docs/local.md` on how to set the install prefix to a directory that doesn't require root.  (For anyone else like myself that installed node with `snap` on Ubuntu.)

noot:
@dimalinux thanks so much for the work on this, apologies for the CI not running automatically. I've updated the settings so it should run automatically for you now.

docs/build.md
```diff
@@ -1,14 +1,13 @@
...
+1. Install Golang
 
-For Linux 64-bit:
+On Ubuntu, the easiest way to keep up-to-date with the latest stable version of
+Go is with snap:
 ```bash
-wget https://go.dev/dl/go1.18.linux-amd64.tar.gz
-rm -rf /usr/local/go && tar -C /usr/local -xzf go1.18.linux-amd64.tar.gz
-echo "export PATH=$PATH:/usr/local/go/bin" >> .profile
-source .profile
```
noot, line 10: can you leave the previous instructions as well? many users are anti-snap :P

docs/local.md
```diff
@@ -3,23 +3,29 @@
...
 
 #### Set up development environment
 
 Note: the `scripts/install-monero-linux.sh` script will download the monero binaries needed for you. You can also check out the `scripts/run-unit-tests.sh` script for the commands needed to setup the environment.
 
-Start ganache-cli with determinstic keys:
+Start ganache-cli with deterministic keys:
 ```bash
-ganache-cli -d
+ganache-cli --deterministic --accounts=20
```
noot, line 22: why do we need 20 accounts?

monero/client_test.go
noot: filepath!

monero/client_test.go
noot: interesting that it doesn't have a type error actually, good to know!

protocol/xmrmaker/swap_state_test.go
noot: sounds good!

protocol/xmrmaker/swap_state_test.go
```diff
@@ -46,11 +45,8 @@ var (
 	defaultTimeoutDuration, _ = time.ParseDuration("86400s") // 1 day = 60s * 60min * 24hr
 )
 
-func newTestXMRMaker(t *testing.T) *Instance {
-	pk, err := ethcrypto.HexToECDSA(common.DefaultPrivKeyXMRMaker)
-	require.NoError(t, err)
-
-	ec, err := ethclient.Dial(common.DefaultEthEndpoint)
```
noot, line 53: curious why you changed the ethclient to be an argument instead of creating it here? seems like basically every test case creates one now instead and passes it in

protocol/xmrtaker/net_test.go
```diff
@@ -1,18 +1,21 @@
...
+	"github.com/noot/atomic-swap/common"
 	"github.com/noot/atomic-swap/common/types"
 
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/stretchr/testify/require"
 )
 
-func newTestXMRTaker(t *testing.T) *Instance {
-	b := newBackend(t)
+func newTestXMRTaker(t *testing.T, ec *ethclient.Client) *Instance {
```
noot, line 14: same here!

protocol/xmrtaker/recovery_test.go
```diff
@@ -1,17 +1,19 @@
...
 	"github.com/noot/atomic-swap/common"
 
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/stretchr/testify/require"
 )
 
-func newTestRecoveryState(t *testing.T) *recoveryState {
-	s := newTestInstance(t)
+func newTestRecoveryState(t *testing.T, ec *ethclient.Client) *recoveryState {
```
noot, line 15: same here!

docs/local.md
noot: nevermind, the tests/ package explains it :)

noot (commented):
looks good and CI is passing! just have a few final comments/questions.

docs/build.md
dimalinux: @noot I'll put them back.  My concern with leaving them in, was that we'll need to update these instructions frequently if the release version is included. The URL we include has the `/usr/loca/go` instructions and it automatically gets updated on every new release.

protocol/xmrmaker/swap_state_test.go
dimalinux: The ethclient is creating a persistent websocket connection.  To get it properly closed by the unit tests even if there is an error, we need to have a `defer ec.close()` statement at a scope that is far enough out to not close the connection too early.

I thought about creating a test suite to reduce the boiler plate code for creating the connection, but decided that it would be too big of a change to take on in this PR.  I can do it in a future PR if you think it will make the code easier to maintain.

protocol/xmrtaker/net_test.go
dimalinux: See my reply on `newTestXMRMaker` above and let me know if you have an alternative solution.

protocol/xmrmaker/swap_state_test.go
noot: you can keep it in the `newTestXMRMaker` function and use `t.Cleanup()` within it to call `ec.Close()` when each unit test is done. that should have the same functionality I believe

protocol/xmrtaker/swap_state_test.go
```diff
@@ -308,7 +313,10 @@ func TestSwapState_NotifyClaimed(t *testing.T) {
 	require.NoError(t, err)
 
 	// mine some blocks to get xmr first
-	_ = maker.GenerateBlocks(xmrmakerAddr.Address, 60)
+	err = maker.GenerateBlocks(xmrmakerAddr.Address, 60)
+	require.NoError(t, err)
+	err = maker.Refresh()
```
dimalinux, line 318: After I switched the code to use dynamic `monero-wallet-rpc` instances, this test started failing with insufficient balance.  Before the change, the wallet would have had an existing balance created by other tests.  The change here that made the difference was the `Refresh`.  Mining 60 blocks works, but the wallet isn't syncing its balance often enough to know that its balance changed without the `Refresh`.  We probably need to audit the tests for every call to `GenerateBlocks` and see if a wallet refresh after it is warranted.  There are probably places in the tests that are mining more blocks than needed, because if you mine enough blocks the mining operation will overlap with a non-explicit wallet refresh.

PR #138, opened 2022-06-29
update to current version of ganache

dimalinux:
Updates the repo to install `ganache` instead of `ganache-cli`.  When installing `ganache-cli` (before this PR), you get the following messages:
```
npm WARN deprecated ganache-cli@6.12.2: ganache-cli is now ganache; visit https://trfl.io/g7 for details
[...]
3 vulnerabilities (1 moderate, 2 high)
```
Interestingly, when you install `ganache`, you get both `ganache-cli` and `ganache` (identical executables) presumably for backwards compatibility.  The current method installing `ganache-cli` gives `v6.12.2`.  Installing via `ganache` gives version `v7.3.2`. There were some code fixes to use the correct Ethereum ChainID with v7, but there were already merged.

============ Code changes needed when upgrading to newer Ganache ============

One of the things that changed in the newer version of Ganache, is gas fee estimation uses the timestamp from the last mined block instead of the local computer's current timestamp. This mirrors what Geth does, so the Ganache change is helpful, but it causes a few problems:

1. If we want to do dynamic fee estimation for a transaction that can not happen before time T0 or T1, we need to wait until there is one mined block after T0 or T1 to create our transaction. Otherwise the transaction will be reverted during fee estimation, returning an error when creating the transaction, and we'll never send the transaction out to the network.
2. Ganache, by default, only mines blocks when you create a transaction. In order to have a mined block after T0 or T1 to calculate transaction fees, we need to stop using instamine. Instead of using instamine, we now mine a block every second. This PR chose the minimum block interval for speedier tests.
3. With instamine enabled, transactions are mined as soon as they are created. You never have to wait for receipts, because there is no delay between the creation of a transaction and it being mined. While this feature can be useful when testing, it leads to code that doesn't wait for blocks to be mined and does not handle reverted transactions that are only reverted after they have gone out to the network and mined into a block that does not meet their requirements. This PR addresses those issues.

Apologies for the huge PR. I originally thought upgrading ganache was going to be a simple task. That said, making the changes to work with the newer ganache puts us closer to working reliably on real networks.


Makefile
```diff
@@ -1,29 +1,37 @@
-.PHONY: lint test install build build-dleq mock
```
dimalinux, line 1: These were out of sync with the rest of the file. It's uglier, but putting the `.PHONY` label above each non-filesystem target is the only way I've found to correctly maintain them.

cmd/daemon/contract_test.go
```diff
@@ -1,32 +1,28 @@
...
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
-	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/stretchr/testify/require"
 )
 
 func TestGetOrDeploySwapFactory(t *testing.T) {
 	pk, err := ethcrypto.HexToECDSA(tests.GetTakerTestKey(t))
 	require.NoError(t, err)
 
-	ec, err := ethclient.Dial(common.DefaultEthEndpoint)
```
dimalinux, line 20: I created the `tests.NewEthClient` function to eliminate some boilerplate code and reduce the code points that need changing if you want to test on a different chain.

cmd/recover/main.go
```diff
@@ -184,7 +184,7 @@ func (inst *instance) recover(c *cli.Context) error {
 		}
 
 		if res.Claimed {
-			log.Info("claimed ether from contract! transaction hash=%s", res.TxHash)
```
dimalinux, line 187: I wonder if we can teach the linter to catch these errors automatically like it would for the system `fmt` and `log` packages.

protocol/backend/eth_util.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// WaitForEthBlockAfterTimestamp returns the header of the first block received where the block's timestamp
+// is >= ts.
+func WaitForEthBlockAfterTimestamp(ctx context.Context, ec *ethclient.Client, ts int64) (*ethtypes.Header, error) {
+	timeDelta := time.Duration(ts-time.Now().Unix()) * time.Second
+
+	// The sleep is safe even if timeDelta is negative. We only optimise for timestamps in the future, but if
+	// the timestamp had already passed for some reason, nothing bad happens other than waiting for an additional
+	// block to be mined.
```
dimalinux, line 30: The last part of the comment might be wrong.  It felt like I was getting the current block header as soon as subscribing (verses not getting any headers until the next block is mined).  The documentation on this is unhelpful.

protocol/xmrmaker/swap_state.go
```diff
@@ -334,6 +334,7 @@ func (s *swapState) tryClaim() (ethcommon.Hash, error) {
 	}
 
 	if untilT0 > 0 && stage != swapfactory.StageReady {
+		// TODO: t0 could be 24 hours from now. Don't we want to poll the stage periodically?
```
dimalinux, line 337: @noot Fixing this is probably out of scope for this PR, but do you agree that this is a important issue to fix?

protocol/xmrmaker/swap_state_test.go
```diff
@@ -183,10 +177,6 @@ func TestSwapState_GenerateAndSetKeys(t *testing.T) {
 }
 
 func TestSwapState_ClaimFunds(t *testing.T) {
-	if testing.Short() {
-		t.Skip() // TODO: randomly fails on CI with "no contract code at given address"
-	}
```
dimalinux, line 188: I can put this back if it is still happening.  It sounds like the error you might get if you don't wait for the contract creation to get mined into a block.

recover/recovery_test.go
```diff
@@ -44,29 +43,26 @@ func newSwap(
 	pk, err := ethcrypto.HexToECDSA(tests.GetTakerTestKey(t))
 	require.NoError(t, err)
 
-	txOpts, err := bind.NewKeyedTransactorWithChainID(pk, big.NewInt(common.GanacheChainID))
+	ec, chainID := tests.NewEthClient(t)
+
+	txOpts, err := bind.NewKeyedTransactorWithChainID(pk, chainID)
 	require.NoError(t, err)
 
-	ec, err := ethclient.Dial(common.DefaultEthEndpoint)
```
dimalinux, line 50: This client wasn't used anywhere.

recover/recovery_test.go
```diff
@@ -162,7 +154,7 @@ func TestRecoverer_RecoverFromXMRMakerSecretAndContract_Claim_afterTimeout(t *te
 	require.NoError(t, err)
 
 	claimKey := keys.Secp256k1PublicKey.Keccak256()
-	addr, contract, swapID, swap := newSwap(t, claimKey, [32]byte{}, true)
```
dimalinux, line 165: This puts the logic back to the way it was before my previous PR. The `setReady` value here is the only thing that differentiates this test from the previous test.

scripts/cleanup-test-processes.sh
```diff
@@ -0,0 +1,11 @@
...
+# This script can be handy to kill test processes launched manually
+# for debugging or still hanging around for other reasons.
+
+pkill --uid "${UID}" --full '/monerod .* --regtest '
+pkill --uid "${UID}" --full '/ganache.* --deterministic '
+
+# If you have monero-wallet-rpc or swapd processes owned by the current user
+# that you don't want to kill, don't use this script!
+pkill --uid "${UID}" --full '/monero-wallet-rpc '
+pkill --uid "${UID}" --full '/swapd '
```
dimalinux, line 11: This script is helpful for me at least. If you don't think it is useful, I can get rid of it and just use a bash function from my startup scripts.

swapfactory/utils.go
```diff
@@ -26,13 +26,13 @@ func GetSecretFromLog(log *ethtypes.Log, event string) (*mcrypto.PrivateSpendKey
 		return nil, errors.New("invalid event name, must be one of Claimed or Refunded")
 	}
 
-	abi, err := abi.JSON(strings.NewReader(SwapFactoryABI))
```
dimalinux, line 29: I'm surprised the linter didn't flag a local variable named the same as an imported package.

tests/integration_test.go
```diff
@@ -98,8 +99,8 @@ func TestXMRTaker_Discover(t *testing.T) {
 	_, err := bc.MakeOffer(xmrmakerProvideAmount, xmrmakerProvideAmount, exchangeRate)
 	require.NoError(t, err)
 
-	c := rpcclient.NewClient(defaultXMRTakerDaemonEndpoint)
-	providers, err := c.Discover(types.ProvidesXMR, defaultDiscoverTimeout)
```
dimalinux, line 102: Just trying to make the tests easier to understand.  Bob was getting a `b` prefix, so I gave Alice an `a` prefix.

tests/integration_test.go
```diff
@@ -127,9 +128,16 @@ func TestXMRTaker_Query(t *testing.T) {
...
+	var respOffer *types.Offer
+	for _, offer := range resp.Offers {
+		if offerID == offer.ID.String() {
+			respOffer = offer
+		}
+	}
+	require.NotNil(t, respOffer)
+	require.Equal(t, xmrmakerProvideAmount, respOffer.MinimumAmount)
+	require.Equal(t, xmrmakerProvideAmount, respOffer.MaximumAmount)
+	require.Equal(t, exchangeRate, float64(respOffer.ExchangeRate))
```
dimalinux, line 140: If you run this test multiple times, the new offer that we just added is in the last index, not the first index.  I wasn't sure if I should hard-code that though, so I search the the offer ID.  I was running these tests by hand out-of-order and the values `resp.Offers[0]` where not matching up, because the offer in index zero wasn't added by this test.

tests/integration_test.go
```diff
@@ -156,51 +164,48 @@ func TestSuccess(t *testing.T) {
...
 				}
 
 				if status != types.CompletedSuccess {
 					errCh <- fmt.Errorf("swap did not complete successfully: got %s", status)
 				}
-
 				return
 			case <-time.After(testTimeout):
 				errCh <- errors.New("make offer subscription timed out")
+				return
```
dimalinux, line 181: Without this, we can't get past the waigroup to print the error message.

tests/integration_test.go
```diff
@@ -225,7 +230,7 @@ func TestRefund_XMRTakerCancels(t *testing.T) {
 
 	const (
 		testTimeout = time.Second * 60
-		swapTimeout = 5 // 5s
```
dimalinux, line 228: 7s is now the fastest we can do with 1s block intervals instead of instamine.

tests/integration_test.go
```diff
@@ -253,54 +259,66 @@ func TestRefund_XMRTakerCancels(t *testing.T) {
...
-				if status != types.CompletedRefund {
-					errCh <- fmt.Errorf("swap did not refund successfully for XMRMaker: got %s", status)
+				switch status {
+				case types.CompletedRefund:
+					// Do nothing, desired outcome
+				case types.CompletedSuccess:
+					t.Log("XMRMaker completed swap before XMRTaker's cancel took affect")
+					swapSucceeded = true
+				default:
+					errCh <- fmt.Errorf("swap did not succeed or refund for XMRMaker: status=%s", status)
```
dimalinux, line 273: We're in `TestRefund_XMRTakerCancels` here.  I wasn't sure how to fix this. The existing goal, as best as I can tell, was to have Alice get a refund, but on my system the swap is going to complete successfully before she cancels it.  That is what should happen the majority of the time, isn't it?

tests/integration_test.go
```diff
@@ -318,15 +336,24 @@ func TestRefund_XMRTakerCancels(t *testing.T) {
...
 // TestRefund_XMRMakerCancels_untilAfterT1 tests the case where XMRTaker and XMRMaker
 // both lock their funds, but XMRMaker goes offline
 // until time t1 in the swap contract passes. This triggers XMRTaker to refund, which XMRMaker will then
 // "come online" to see, and he will then refund also.
 func TestRefund_XMRMakerCancels_untilAfterT1(t *testing.T) {
-	testRefundXMRMakerCancels(t, 5, types.CompletedRefund)
+	// TODO: This test is broken. It is supposed to have types.CompletedRefund below, but I don't see
+	//       an easy way to make the documented above behaviour happen (i.e. take the XMRMaker offline
+	//       until after T1.
+	testRefundXMRMakerCancels(t, 7, types.CompletedSuccess)
```
dimalinux, line 356: @noot This is similar to the issue I was having in `TestRefund_XMRTakerCancels`.  The swap is going to succeed before Bob ever gets his refund request in. There is no easy way here on the front-end to have the backend for XMRMaker go offline until T1.

protocol/txsender/sender.go
```diff
@@ -64,12 +65,9 @@ func (s *privateKeySender) SetContractAddress(_ ethcommon.Address) {}
 func (s *privateKeySender) NewSwap(_ types.Hash, _pubKeyClaim [32]byte, _pubKeyRefund [32]byte,
 	_claimer ethcommon.Address, _timeoutDuration *big.Int, _nonce *big.Int,
 	value *big.Int) (ethcommon.Hash, *ethtypes.Receipt, error) {
-	s.txOpts.Value = value
-	defer func() {
-		s.txOpts.Value = nil
-	}()
-
-	tx, err := s.contract.NewSwap(s.txOpts, _pubKeyClaim, _pubKeyRefund, _claimer, _timeoutDuration, _nonce)
+	txOpts := *s.txOpts // make a copy, so we don't modify the original
```
dimalinux, line 68: The `bind.TransactOpts` object is usually created fresh for every transaction. We don't have the information to do that here, so instead I changed the code to never modify `s.txOpts` and always use a copy in transactions.  There are many fields in this object that are initialized on-demand when creating the transaction, but only if they don't already have a value set.  We were running into errors, because fees had been calculated on a different transaction and then were being reused against a transaction that cost more.

cmd/recover/main.go
noot: that would be handy!

protocol/backend/backend.go
```diff
@@ -337,6 +339,26 @@ func (b *backend) WaitForReceipt(ctx context.Context, txHash ethcommon.Hash) (*e
...
+func (b *backend) WaitForTimestamp(ctx context.Context, ts time.Time) error {
+	hdr, err := WaitForEthBlockAfterTimestamp(ctx, b.ethClient, ts.Unix())
+	if err == nil {
+		log.Debug("Wait complete for block %d with ts=%s >= %s",
+			hdr.Number.Uint64(),
+			time.Unix(int64(hdr.Time), 0).Format(common.TimeFmtSecs),
+			ts.Format(common.TimeFmtSecs),
+		)
+	}
+	return err
```
noot, line 351:
```suggestion
	hdr, err := WaitForEthBlockAfterTimestamp(ctx, b.ethClient, ts.Unix())
	if err != nil {
	    return err
	}
	
        log.Debug("Wait complete for block %d with ts=%s >= %s",
              hdr.Number.Uint64(),
              time.Unix(int64(hdr.Time), 0).Format(common.TimeFmtSecs),
              ts.Format(common.TimeFmtSecs),
        )
	return nil
```

is there an `err != nil` case?

protocol/backend/eth_util.go
```diff
@@ -0,0 +1,54 @@
...
+			<-timer.C
+		}
+	case <-timer.C:
+	}
+}
+
+// WaitForEthBlockAfterTimestamp returns the header of the first block received where the block's timestamp
+// is >= ts.
+func WaitForEthBlockAfterTimestamp(ctx context.Context, ec *ethclient.Client, ts int64) (*ethtypes.Header, error) {
+	timeDelta := time.Duration(ts-time.Now().Unix()) * time.Second
```
noot, line 26:
```suggestion
func WaitForEthBlockAfterTimestamp(ctx context.Context, ec *ethclient.Client, ts time.Time) (*ethtypes.Header, error) {
	timeDelta := time.Until(ts).Seconds()
```

I think this should be equivalent?

protocol/backend/eth_util.go
noot: https://pkg.go.dev/github.com/ethereum/go-ethereum/ethclient#Client.SubscribeNewHead

the docs seem to suggest it sends the current head over the channel immediately

protocol/backend/eth_util.go
```diff
@@ -0,0 +1,54 @@
...
+	defer sub.Unsubscribe()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case err := <-sub.Err():
+			return nil, err
+		case header := <-headers:
+			if header.Time >= uint64(ts) {
```
noot, line 49: if you change `ts` to `time.Time`:
```suggestion
			if header.Time >= ts.Unix() {
```

protocol/xmrmaker/message_handler.go
```diff
@@ -182,23 +183,40 @@ func (s *swapState) handleNotifyETHLocked(msg *message.NotifyETHLocked) (net.Mes
...
-		case <-s.readyCh:
+	select {
+	case <-s.ctx.Done():
+		return
+	case <-s.readyCh:
+		return
+	case err := <-waitCh:
+		if err != nil {
+			// TODO: Do we propagate this error? If we retry, the logic should probably be inside WaitForTimestamp.
+			log.Errorf("Failure waiting for T0 timeout: err=%s", err)
```
noot, line 215: the only errors would be if the subscription fails for some reason (bad, should probably retry), or if the context is cancelled (which is fine)

do you know in what cases the subscription might error? i guess if the ethclient connection goes down?

protocol/xmrmaker/swap_state.go
noot: yeah, this should probably be fixed. in the ideal case, the eth-provider should notify us when they call ready, but we can't necessarily count on that, so would be good to poll it here.

protocol/xmrmaker/swap_state_test.go
noot: if it's not happening anymore, that's great :D

protocol/xmrtaker/message_handler.go
```diff
@@ -148,17 +150,26 @@ func (s *swapState) handleSendKeysMessage(msg *net.SendKeysMessage) (net.Message
...
-
-		log.Debugf("time until refund: %vs", until.Seconds())
+		// Current algorithm is to trigger the timeout when only 15% of the allotted
+		// time is remaining. If the block interval is 1 second on a test network and
+		// and T0 is 7 seconds after swap creation, we need the refund to trigger more
+		// than one second before the block with a timestamp exactly equal to T0 to
+		// satisfy the strictly less than requirement. 7s * 15% = 1.05s. 15% remaining
+		// may be reasonable even with large timeouts on production networks, but more
+		// research is needed.
+		t0Delta := s.t1.Sub(s.t0) // time between swap start and T0 is equal to T1-T0
```
noot, line 160: it doesn't make a difference now since the time between initiation and t0 is the same as between t0 and t1, but this should probably be the time difference between t0 and initiation

protocol/xmrtaker/message_handler.go
```diff
@@ -290,23 +302,40 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) (net.Message
...
+	select {
+	case <-s.ctx.Done():
+		return
+	case <-s.claimedCh:
+		return
+	case err := <-waitCh:
+		if err != nil {
+			// TODO: Do we propagate this error? If we retry, the logic should probably be inside WaitForTimestamp.
+			log.Errorf("Failure waiting for T1 timeout: err=%s", err)
 			return
```
noot, line 335: same comment as above

protocol/xmrtaker/swap_state.go
```diff
@@ -317,22 +318,51 @@ func (s *swapState) doRefund() (ethcommon.Hash, error) {
...
+	log.Debugf("tryRefund isReady=%v untilT0=%vs untilT1=%vs",
+		isReady, ts.Sub(s.t0).Seconds(), ts.Sub(s.t1).Seconds())
 
-	if (untilT0 > 0 && isReady) && untilT1 > 0 {
-		// we've passed t0 but aren't past t1 yet, so we need to wait until t1
-		log.Infof("waiting until time %s to refund", s.t1)
-		<-time.After(untilT1)
+	if ts.Before(s.t0) && !isReady {
+		// using errT0 to avoid shadow warnings
+		txHash, errT0 := s.refund()
```
noot, line 347:
```suggestion
		txHash, err := s.refund() //nolint:govet
```

i *think* govet is the linter that complains? either way, i'd just throw a nolint in there

scripts/cleanup-test-processes.sh
noot: this is helpful!

swapfactory/utils.go
```diff
@@ -20,19 +21,35 @@ const (
 	StageCompleted
 )
 
+// StageToString converts a contract Stage enum value to a string
+func StageToString(stage byte) string {
+	switch stage {
```
noot, line 26:
```suggestion
type Stage byte
const (
	StageInvalid Stage = iota //nolint:revive
	StagePending
	StageReady
	StageCompleted
)

// String converts a contract Stage enum value to a string
func (stage Stage) String() string {
	switch stage {
	// rest of code here
```

this will auto-format `Stage` into a string for printlns, logs, and such, may be overkill but could be handy later

tests/integration_test.go
noot: not sure why the swap would be successful here unless the time between initiation and t0 is too small, since the xmr-taker cancels right after `types.ETHLocked`, so the contract is never set to ready and the xmr-maker shouldn't be able to claim until t0. maybe just increase the `swapTimeout`?

tests/integration_test.go
```diff
@@ -318,15 +337,22 @@ func TestRefund_XMRTakerCancels(t *testing.T) {
...
+	}
 }
 
 // TestRefund_XMRMakerCancels_untilAfterT1 tests the case where XMRTaker and XMRMaker
 // both lock their funds, but XMRMaker goes offline
 // until time t1 in the swap contract passes. This triggers XMRTaker to refund, which XMRMaker will then
 // "come online" to see, and he will then refund also.
 func TestRefund_XMRMakerCancels_untilAfterT1(t *testing.T) {
-	testRefundXMRMakerCancels(t, 5, types.CompletedRefund)
+	t.Skip() // @noot, this test is a giant race condition, and I need your help on how to fix it.
```
noot, line 354: the problem is that the swap program might end up completing the swap before the cancel call goes through right? i wonder if it would help to have a test endpoint for initiating swaps that cancel after a certain stage. however i'm also somewhat opposed to that as it requires changing the non-test code somewhat significantly for the purpose of tests. i'll give it some thought

noot (commented):
changes look good overall, really nice work! let me know if you need help with any of the test failures, i'll have time next week to look into it.

protocol/xmrmaker/message_handler.go
dimalinux: If you are using a local ethereum node (which users should), it is probably something that will never error.

I think the way to figure the answer out, will be to run some tests against a very-unreliable public node like wss://goerli-light.eth.linkpool.io/ws and see what errors happen in practice.

protocol/xmrtaker/message_handler.go
dimalinux: I would have done that, but I didn't have the information to compute it. We'll need to store the timestamp of the block that created the swap in `swapState` to compute it that way.

ethereum/block/testdata/UTContract.sol
```diff
@@ -0,0 +1,19 @@
...
+// will be greater than what we passed in.
+//
+contract UTContract {
+    uint256 private stamp;
+
+    function check_stamp(uint256 _stamp) external {
+        require(block.timestamp <= _stamp, "block.timestamp was not less than stamp");
+        stamp = _stamp; // Prevent the function from being view-only
+    }
+}
```
dimalinux, line 19: This is the contract I used when developing `errorFromBlock` and it seemed like the best way to unit test it. Even if we could test it using the `SwapFactory` contract, that package depends on this one, so we couldn't use it from here. Go tools ignore folders named `testdata`. Normally, people place input data for Unit tests in these directories, but it was the best idea I could come up with for where to put data used to generate code that test cases link against.

protocol/backend/eth_util.go
dimalinux: I adjusted the comment.  `WaitForReceipt` and `WaitForEthBlockAfterTimestamp` are now in a new package named `block` under `ethereum`.

ethereum/block/wait_for_receipt.go
```diff
@@ -0,0 +1,51 @@
...
+	receiptSleepDuration = time.Second * 10
+)
+
+var (
+	log               = logging.Logger("ethereum/block")
+	errReceiptTimeOut = errors.New("failed to get receipt, timed out")
+)
+
+// WaitForReceipt waits for the transaction to be mined into a block. If the transaction was reverted when mined,
+// we return an error describing why.
```
dimalinux, line 27: It actually waits until the transaction has been mined into a block plus some number of confirmations. The documentation for `TransactionReceipt` is not clear on what that number of confirmations is. Empirically, it appears to be 9 confirmations. If I get the current block after waiting for the receipt, it is 9 blocks ahead of the value in `receipt.BlockNumber`.

swapfactory/utils.go
dimalinux: That is the way I originally wanted/tried to implement it. The problem was that the generated SwapFactory code is using `byte` for the stage. Explicitly calling a function seemed 10% more obvious than doing a cast to enable the String() method on the type.

protocol/txsender/sender.go
```diff
@@ -42,6 +34,7 @@ type privateKeySender struct {
 	ec       *ethclient.Client
 	contract *swapfactory.SwapFactory
 	txOpts   *bind.TransactOpts
+	txLock   sync.Mutex // locks from TX start until receipt so we don't reuse ETH nonce values
```
dimalinux, line 37: This is a new and important change.  If we are doing multiple current swaps, we can't have multiple pending transactions from the same wallet running at the same time.  If there is a use case for doing multiple swaps simultaneously with short timeouts, the user is going to want to be able to use multiple wallets so one swap doesn't block an important timeout for a different swap while a transaction is getting processed.

protocol/xmrtaker/errors.go
```diff
@@ -12,14 +13,16 @@ var (
...
+	errRefundSwapCompleted     = fmt.Errorf("can not refund, %w", errSwapCompleted)
 	errNilMessage              = errors.New("message is nil")
 	errIncorrectMessageType    = errors.New("received unexpected message")
 	errNoLockedXMRAddress      = errors.New("got empty address for locked XMR")
 	errClaimTxHasNoLogs        = errors.New("claim transaction has no logs")
 	errNoPublicKeysSet         = errors.New("our public keys aren't set")
 	errCounterpartyKeysNotSet  = errors.New("counterparty's keys aren't set")
 	errSwapInstantiationNoLogs = errors.New("expected 1 log, got 0")
-	errSwapCompleted           = errors.New("swap has already completed")
+	errSwapCompleted           = errors.New("swap is already completed")
```
dimalinux, line 25: Matching the wording of the error to the wording of the revert in the contract.

noot (approved):
great work, lgtm!

PR #139, opened 2022-07-06
Fix double locks

dimalinux:
* Fixes an issue in `handleNotifyXMRLock` where the `swapState` lock could be grabbed twice after T1 expires
* Changes the lock in both `swapState` structs to be explicit (making the code easier to audit)
* Moves the T0 and T1 expiration code into dedicated functions for readability (verses having them in long code blocks inside a switch of an inline go process)
* Increased mined blocks by `TestSwapState_NotifyClaimed` (was still flaky and occasionally running out of money with only 60 blocks)
* Reaps exited monero-wallet-rpc processes immediately instead of when testing for a package completes
* Change to `TestRecoverer_RecoverFromXMRMakerSecretAndContract_Claim_afterTimeout` to call `set_ready`

### DIFF
--- a/protocol/xmrmaker/message_handler.go
+++ b/protocol/xmrmaker/message_handler.go
@@ -202,14 +202,8 @@ func (s *swapState) handleNotifyETHLocked(msg *message.NotifyETHLocked) (net.Mes
 }
 
 func (s *swapState) handleT0Expired() {
-	doExit := false
 	s.lockState()
-	defer func() {
-		s.unlockState()
-		if doExit {
-			_ = s.Exit()
-		}
-	}()
+	defer s.unlockState()
 
 	if !s.info.Status().IsOngoing() {
 		// swap was already completed, just return
@@ -221,7 +215,9 @@ func (s *swapState) handleT0Expired() {
 	if err != nil {
 		log.Errorf("failed to claim: err=%s", err)
 		// TODO: retry claim, depending on error
-		doExit = true
+		if err = s.exit(); err != nil {
+			log.Errorf("exit failed: err=%s", err)
+		}
 		return
 	}
 
--- a/protocol/xmrmaker/message_handler.go
+++ b/protocol/xmrmaker/message_handler.go
@@ -23,8 +23,8 @@ func (s *swapState) HandleProtocolMessage(msg net.Message) (net.Message, bool, e
 		return nil, true, errNilSwapState
 	}
 
-	s.Lock()
-	defer s.Unlock()
+	s.lockState()
+	defer s.unlockState()
 
 	if s.ctx.Err() != nil {
 		return nil, true, fmt.Errorf("protocol exited: %w", s.ctx.Err())
@@ -184,39 +184,14 @@ func (s *swapState) handleNotifyETHLocked(msg *message.NotifyETHLocked) (net.Mes
 
 	go func() {
 		until := time.Until(s.t0)
-
 		log.Debugf("time until t0: %vs", until.Seconds())
 
 		select {
 		case <-s.ctx.Done():
 			return
+		// TODO: Document why we add one second
 		case <-time.After(until + time.Second):
-			s.Lock()
-			defer s.Unlock()
-
-			if !s.info.Status().IsOngoing() {
-				// swap was already completed, just return
-				return
-			}
-
-			// we can now call Claim()
-			txHash, err := s.claimFunds()
-			if err != nil {
-				log.Errorf("failed to claim: err=%s", err)
-				// TODO: retry claim, depending on error
-				_ = s.exit()
-				return
-			}
-
-			log.Debug("funds claimed!")
-			s.clearNextExpectedMessage(types.CompletedSuccess)
-
-			// send *message.NotifyClaimed
-			if err := s.SendSwapMessage(&message.NotifyClaimed{
-				TxHash: txHash.String(),
-			}, s.ID()); err != nil {
-				log.Errorf("failed to send NotifyClaimed message: err=%s", err)
-			}
+			s.handleT0Expired()
 		case <-s.readyCh:
 			return
 		}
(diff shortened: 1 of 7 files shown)


protocol/xmrmaker/swap_state.go
```diff
@@ -40,9 +40,9 @@ var (
 type swapState struct {
 	backend.Backend
 
-	ctx    context.Context
-	cancel context.CancelFunc
-	sync.Mutex
```
dimalinux, line 45: I changed the swapState<->lock relationship from `is-a` to a `has-a`, because without a variable the code is hard to audit, as you can't search for everywhere the lock is used.  You can search on the `Lock` and `Unlock` methods, but then you'll end up finding every lock use, not the use on a specific maker or taker swapState.

protocol/xmrtaker/message_handler.go
```diff
@@ -296,31 +296,9 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) (net.Message
...
-			if err = s.SendSwapMessage(&message.NotifyRefund{
-				TxHash: txhash.String(),
-			}, s.ID()); err != nil {
-				log.Errorf("failed to send refund message: err=%s", err)
-			}
-
+		// TODO: document why we add one second
+		case <-time.After(until + 5*time.Second):
+			s.handleT1Expired()
 			_ = s.Exit()
```
dimalinux, line 302: This was the problem line right here. `Exit` grabs the lock, but we already have the lock on what was formerly line 300.

protocol/xmrmaker/message_handler.go
```diff
@@ -184,39 +184,14 @@ func (s *swapState) handleNotifyETHLocked(msg *message.NotifyETHLocked) (net.Mes
...
-				// swap was already completed, just return
-				return
-			}
-
-			// we can now call Claim()
-			txHash, err := s.claimFunds()
-			if err != nil {
-				log.Errorf("failed to claim: err=%s", err)
-				// TODO: retry claim, depending on error
-				_ = s.exit()
```
dimalinux, line 207: I'm just realizing with the diff view of the PR, that this `xmrmaker` version of `swapState` is calling `exit` with a lower case `e`, while the `xmrtaker` version was calling `Exit` with a capital `E`.  The lower-case version does not grab the lock, so this `xmrmaker` version was actually safe.

I would still break the 2 timeout handlers into separate functions the way this PR does to make the code readable though.

protocol/xmrmaker/message_handler.go
```diff
@@ -226,6 +201,40 @@ func (s *swapState) handleNotifyETHLocked(msg *message.NotifyETHLocked) (net.Mes
...
+		// swap was already completed, just return
+		return
+	}
+
+	// we can now call Claim()
+	txHash, err := s.claimFunds()
+	if err != nil {
+		log.Errorf("failed to claim: err=%s", err)
+		// TODO: retry claim, depending on error
+		doExit = true
```
dimalinux, line 224: @noot Based on the comments I left above, I can change this to call `s.exit()`, which does not grab the lock, here and then simplify the `defer` code at the top.

protocol/xmrtaker/message_handler.go
dimalinux: @noot Based on other comments above, I can change this line from `s.Exit()` to `s.exit()` and move this line to the end `handleT1Expired()` below.

recover/recovery_test.go
```diff
@@ -158,15 +158,11 @@ func TestRecoverer_RecoverFromXMRMakerSecretAndContract_Claim(t *testing.T) {
...
 func TestRecoverer_RecoverFromXMRMakerSecretAndContract_Claim_afterTimeout(t *testing.T) {
-	// if testing.Short() {
-	// 	t.Skip() // TODO: fails on CI with "no contract code at address"
-	// }
-
 	keys, err := pcommon.GenerateKeysAndProof()
 	require.NoError(t, err)
 
 	claimKey := keys.Secp256k1PublicKey.Keccak256()
-	addr, contract, swapID, swap := newSwap(t, claimKey, [32]byte{}, false)
```
dimalinux, line 169: @noot The boolean field at the end is `setReady`. If it's false, the test was failing with `VM Exception while processing transaction: revert too early to claim!`.  Here's the relevant solidity code:
```
    function claim(Swap memory _swap, bytes32 _s) public {
[...]
        require((block.timestamp >= _swap.timeout_0 || swapStage == Stage.READY), "too early to claim!");
```
... so the test was failing because the `swapStage` was never set to `READY`. Should this test have worked before my change?

tests/monero_wallet_rpc.go
```diff
@@ -28,6 +28,7 @@ func CreateWalletRPCService(t *testing.T) string {
 	t.Cleanup(func() {
 		_ = outPipe.Close()
 		_ = cmd.Process.Kill()
+		_ = cmd.Wait()
```
dimalinux, line 31: The defunct processes were getting cleaned up when the test of a particular package ended, so  this change isn't critical, but I should have been reaping the exited children immediately.

protocol/xmrtaker/message_handler.go
dimalinux: Upon further inspection, `xmrtaker` only has an upper-case version of `Exit`. To make the change above, we would need to create a lower-case version (as is done in `xmrmaker`) that does not grab the lock.

protocol/xmrmaker/message_handler.go
```diff
@@ -226,6 +201,36 @@ func (s *swapState) handleNotifyETHLocked(msg *message.NotifyETHLocked) (net.Mes
...
+	if !s.info.Status().IsOngoing() {
+		// swap was already completed, just return
+		return
+	}
+
+	// we can now call Claim()
+	txHash, err := s.claimFunds()
+	if err != nil {
+		log.Errorf("failed to claim: err=%s", err)
+		// TODO: retry claim, depending on error
```
dimalinux, line 217: Not in this PR, but I think it is time to fix this (already existing) TODO.

protocol/xmrtaker/message_handler.go
```diff
@@ -296,32 +296,10 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) (net.Message
 		select {
 		case <-s.ctx.Done():
 			return
+		// TODO: document why we add one second
```
dimalinux, line 299: @noot What comment can I add here to explain why we add a second?

protocol/xmrmaker/swap_state.go
```diff
@@ -109,6 +109,14 @@ func newSwapState(b backend.Backend, offer *types.Offer, om *offerManager, statu
 	return s, nil
 }
 
+func (s *swapState) lockState() {
+	s.stateMu.Lock()
+}
+
+func (s *swapState) unlockState() {
+	s.stateMu.Unlock()
+}
```
dimalinux, line 118: If there is are alternative names that you like better, just let me know and I'll rename these.

protocol/xmrtaker/message_handler.go
```diff
@@ -296,32 +296,10 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) (net.Message
...
-			s.clearNextExpectedMessage(types.CompletedRefund) // TODO: duplicate?
-
-			// send NotifyRefund msg
-			if err = s.SendSwapMessage(&message.NotifyRefund{
-				TxHash: txhash.String(),
-			}, s.ID()); err != nil {
-				log.Errorf("failed to send refund message: err=%s", err)
-			}
-
-			_ = s.Exit()
```
dimalinux, line 324: This was the problem line grabbing the lock a second time.  I separated out an `exit` function (does not grab lock) from `Exit` (grabs lock) to match the existing implementation in `xmrmaker`. The code above was moved into `handleT1Expired` and that function now calls `s.exit` instead of `s.Exit`.

protocol/xmrtaker/message_handler.go
noot: a second is added here so that we don't refund at the *exact* moment we are able to (ie. exactly at t1) as the block time of the network might vary and potentially be slightly behind, so our tx might get rejected

recover/recovery_test.go
noot: I think it should have, because the recoverer should have waited until past t0 to claim. I don't really see anywhere in your changes that would cause this to fail though

noot (approved):
looks good! not sure if the CI failing is a fluke or not, re-ran it to see. feel free to merge when it passes

PR #143, opened 2022-07-30
fix incorrect padding in NewPublicKeyFromBigInt

dimalinux:
On occasion we were getting transactions reverted with `provided secret does not match the expected public key`. It was a padding issue when we were converting X/Y in big.Int format when X or Y had less than 32 bytes.  The array was in big endian format and we were shifting the value left by however many high order bytes were all zero. This PR fixes the issue.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -7,12 +7,14 @@ import (
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 )
 
-// Reverse reverses the byte slice and returns it.
+// Reverse returns a copy of the slice with the bytes in reverse order
 func Reverse(s []byte) []byte {
-	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
-		s[i], s[j] = s[j], s[i]
+	l := len(s)
+	rs := make([]byte, l)
+	for i := 0; i < l; i++ {
+		rs[i] = s[l-i-1]
 	}
-	return s
+	return rs
 }
 
 // EthereumPrivateKeyToAddress returns the address associated with a private key
--- a/crypto/secp256k1/secp256k1.go
+++ b/crypto/secp256k1/secp256k1.go
@@ -16,7 +16,7 @@ var (
 
 // PublicKey represents a secp256k1 public key
 type PublicKey struct {
-	x, y [32]byte
+	x, y [32]byte // points stored in big-endian format
 }
 
 // NewPublicKey returns a new public key from the given (x, y) coordinates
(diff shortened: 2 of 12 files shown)


swapfactory/swap_factory_test.go
```diff
@@ -422,3 +423,39 @@ func TestSwap_MultipleSwaps(t *testing.T) {
...
+
+func TestSwapFactory_MultVerifyForClaim(t *testing.T) {
+	auth, conn, _ := setupXMRTakerAuth(t)
+	defer conn.Close() // remove when merging into update-ganache branch
+
+	_, _, contract, err := DeploySwapFactory(auth, conn)
+	require.NoError(t, err)
+	// WaitForReceipt goes here when merging into update-ganache branch
+
+	dleq := &dleq.CGODLEq{}
```
dimalinux, line 435: My IDE doesn't like it when we name variables using the same name as an imported package. Would it be a good thing if I made a future pull request that removes the `CGODLEq` object  and we would just call `Prove` and `Verify` directly on the package name?

common/utils.go
```diff
@@ -7,12 +7,14 @@ import (
...
-	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
-		s[i], s[j] = s[j], s[i]
+	l := len(s)
+	rs := make([]byte, l)
+	for i := 0; i < l; i++ {
+		rs[i] = s[l-i-1]
 	}
-	return s
+	return rs
 }
```
dimalinux, line 18: I wasn't expecting `Reverse([]byte) []byte` to be modifying the backing array of the passed in slice and was getting burned by the behavior.  There were not any callers that wanted the backing array modified, so I just updated it.

crypto/secp256k1/secp256k1.go
```diff
@@ -29,10 +29,15 @@ func NewPublicKey(x, y [32]byte) *PublicKey {
...
-	return NewPublicKey(xb, yb)
+	const ptSize = 32
+	var xArray, yArray [ptSize]byte
+	xSlice := x.Bytes()
+	ySlice := y.Bytes()
+	// Copying from a big-endian slice into a big-endian array, so we want padding bytes
+	// on the left if the slice is shorter than the array.
+	copy(xArray[ptSize-len(xSlice):], xSlice)
+	copy(yArray[ptSize-len(ySlice):], ySlice)
+	return NewPublicKey(xArray, yArray)
```
dimalinux, line 40: This was the code change that fixes the problem.

dleq/cgo_dleq_test.go
```diff
@@ -25,3 +30,39 @@ func TestCGODLEq(t *testing.T) {
...
+	for i := 0; i < iterations; i++ {
+		proof, err := (&CGODLEq{}).Prove()
+		require.NoError(t, err)
+		res, err := (&CGODLEq{}).Verify(proof)
+		require.NoError(t, err)
+
+		// The ETH library needs the secret in big-endian format, while the monero library wants it
+		// in little endian format.
+		secretLE := proof.secret[:]
+		secretBE := common.Reverse(secretLE)
```
dimalinux, line 49: This is where I got burned by the original behavior of Reverse.  :) The backing array is in the proof struct, so this was changing the order for all 3 variables.

scripts/install-lint.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
+VERSION="v1.47.2"
 GOBIN="$(go env GOPATH)/bin"
 
 if [[ ! -x "${GOBIN}/golangci-lint" ]]; then
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOBIN}" v1.46.2
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOBIN}" "${VERSION}"
```
dimalinux, line 7: After I had the PR up, I saw that CI was using a newer version of `golangci-lint` than I was using locally. I matched the version and fixed the new complaints.

crypto/secp256k1/secp256k1.go
noot: awesome, makes sense!

noot (approved):
great work, looks good to me. thanks for looking into this issue!

PR #147, opened 2022-08-07
fix various concurrency issues

dimalinux:
This PR fixes:
* Lock issues on the XMR Maker's offer list
* Lock issues on the swapState maps
* Goroutine sync issues in TestError_ShouldOnlyTakeOfferOnce that were sometimes breaking TestSuccess_ConcurrentSwaps
* Ensures that the id field of `types.Offer` is initialized when it is created

For the XMR Maker's offer list, a new package named `protocol/xmrmaker/offers` was created that shifts code out of `protocol/xmrmaker` ensuring that no current or future code can bypass the async lock protections for the offers map.


common/types/offer.go
```diff
@@ -32,6 +38,7 @@ func HexToHash(s string) (Hash, error) {
 // Offer represents a swap offer
 type Offer struct {
 	ID            Hash
+	idMu          sync.Mutex `json:"-"` // prevents race conditions on the first call to GetID
```
dimalinux, line 41: In a previous edition of this PR, @noot wrote:

> I think I'd rather just have a NewOffer function that must be used that sets the ID. however to enforce that NewOffer needs to be used we're have to un-export the struct values, which would cause json.Marshal not to work, but we could just concatenate the values inside the struct. or we could leave the values exported and make sure we use NewOffer.

I'll make that change before we merge. As part of that change, I was wondering if we should do one of:

1. Not use the serialized data from the offer when generating the offer ID. Since the hash includes 8 bytes of random data that we don't store, there is no way for us to do any checks based on the serialized data used for the hash.
2. Keep the serialized data, but also store the random bytes as a nonce. If the nonce is stored, we could do checks on the offer values relative to the ID if there is a use case for doing this.

protocol/xmrmaker/backend_offers.go
```diff
@@ -0,0 +1,35 @@
...
+
+// GetOffers returns all current offers.
+func (b *Instance) GetOffers() []*types.Offer {
+	return b.offerManager.GetOffers()
+}
+
+// ClearOffers clears all offers.
+func (b *Instance) ClearOffers() {
+	b.offerManager.ClearOffers()
+}
```
dimalinux, line 35: These 3 functions were previously in a file named `offers.go` of the same package.  I renamed it to reduce confusion with the other offers related files.

scripts/run-integration-tests.sh
```diff
@@ -51,7 +51,7 @@ sleep 3
 
 # run tests
 echo "running integration tests..."
-TESTS=integration go test ./tests -v
+TESTS=integration go test ./tests -v -count=1
```
dimalinux, line 54: Definitely in the integration tests, where there can be race conditions, we want to rerun all tests and not use cached results.  I made the same change in the unit tests script, but I'm happy to revert that change if we want faster runs of those tests.

tests/integration_test.go
```diff
@@ -205,6 +207,7 @@ func TestSuccess(t *testing.T) {
 				continue
 			}
 			if status != types.CompletedSuccess {
+				cancel()
```
dimalinux, line 210: I'm attempting to speed up the failure when we already have an error by adding these `cancel()` calls.  The taker was getting Refunded here, while the Maker was tied up for 40 seconds locking his Monero, so we'd have to wait for the Maker to hit his timeout clause before the test could synchronize and print out the error.

tests/integration_test.go
```diff
@@ -872,21 +873,28 @@ func TestSuccess_ConcurrentSwaps(t *testing.T) {
...
+	for _, mt := range makerTests {
+		idMap[mt.offerID] = true
+	}
 	offersAfter, err := bc.GetOffers()
 	require.NoError(t, err)
-	require.Equal(t, numConcurrentSwaps, len(offersBefore)-len(offersAfter))
+	for _, o := range offersAfter {
+		id := o.GetID().String()
+		require.Falsef(t, idMap[id], "offer id %s was not removed", id)
+	}
```
dimalinux, line 899: This is the best idea I could come up with to deal with an offer showing up (presumably from a previously exited test), while this test is running. Other suggestions welcome.

protocol/backend/backend.go
```diff
@@ -137,7 +137,7 @@ func NewBackend(cfg *Config) (Backend, error) {
 	}
 
 	if cfg.Environment == common.Development {
-		defaultTimeoutDuration = time.Minute
+		defaultTimeoutDuration = 75 * time.Second
```
dimalinux, line 140: Increased this because, on slow cloud VMs, the XMR locking phase is taking 40 seconds (combined with longer eth transaction times after  abandoning instamine).

protocol/xmrmaker/net.go
```diff
@@ -14,33 +14,39 @@ func (b *Instance) Provides() types.ProvidesCoin {
...
 }
 
-func (b *Instance) initiate(offer *types.Offer, offerExtra *types.OfferExtra, providesAmount common.MoneroAmount,
-	desiredAmount common.EtherAmount) error {
+func (b *Instance) initiate(
+	offer *types.Offer,
+	offerExtra *types.OfferExtra,
+	providesAmount common.MoneroAmount,
+	desiredAmount common.EtherAmount,
+) (*swapState, error) {
```
dimalinux, line 22: This function now returns the swapState so the caller doesn't have to grab the lock again to know what we changed it to.

protocol/xmrmaker/net.go
```diff
@@ -83,31 +92,23 @@ func (b *Instance) HandleInitiateMessage(msg *net.SendKeysMessage) (net.SwapStat
...
+	providedWei := common.EtherToWei(msg.ProvidedAmount)
+	state, err := b.initiate(offer, offerExtra, providedPicoXMR, providedWei)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	s, has := b.swapStates[offerID]
-	if !has {
-		panic("did not store swap state in Instance map")
-	}
```
dimalinux, line 98: The removed code above should have been grabbing the lock before reading from `swapStates`.  The call to `b.initiate` above now returns the state, so we don't need to read it again.

protocol/xmrmaker/swap_state.go
```diff
@@ -77,8 +78,15 @@ type swapState struct {
 	moneroReclaimAddress mcrypto.Address
 }
 
-func newSwapState(b backend.Backend, offer *types.Offer, om *offerManager, statusCh chan types.Status, infoFile string,
-	providesAmount common.MoneroAmount, desiredAmount common.EtherAmount) (*swapState, error) {
+func newSwapState(
+	b backend.Backend,
+	offer *types.Offer,
+	om *offers.Manager,
```
dimalinux, line 84: The type here changed, the rest was just formatting.

protocol/xmrmaker/swap_state_test.go
```diff
@@ -100,7 +100,7 @@ func newTestXMRMaker(t *testing.T) *Instance {
 func newTestInstance(t *testing.T) (*Instance, *swapState) {
 	xmrmaker := newTestXMRMaker(t)
 	infoFile := path.Join(t.TempDir(), "test.keys")
-	swapState, err := newSwapState(xmrmaker.backend, &types.Offer{}, xmrmaker.offerManager, nil, infoFile,
+	swapState, err := newSwapState(xmrmaker.backend, types.NewOffer("", 0, 0, 0), xmrmaker.offerManager, nil, infoFile,
```
dimalinux, line 103: I was just keeping the test the way it was, but it does seems a little strange that `newSwapState` accepts that offer as valid input.

protocol/xmrtaker/net.go
```diff
@@ -17,47 +17,49 @@ func (a *Instance) Provides() types.ProvidesCoin {
...
 func (a *Instance) InitiateProtocol(providesAmount float64, offer *types.Offer) (common.SwapState, error) {
 	receivedAmount := offer.ExchangeRate.ToXMR(providesAmount)
-	err := a.initiate(common.EtherToWei(providesAmount), common.MoneroToPiconero(receivedAmount),
+	state, err := a.initiate(common.EtherToWei(providesAmount), common.MoneroToPiconero(receivedAmount),
 		offer.ExchangeRate, offer.GetID())
 	if err != nil {
 		return nil, err
 	}
 
-	return a.swapStates[offer.GetID()], nil
```
dimalinux, line 26: This access to `a.swapStates` should have been grabbing the lock.  Issue was fixed by having `a.initiate` return the swap state (analogous change was made in `xmrmaker` earlier).

rpc/ws_test.go
```diff
@@ -46,11 +47,12 @@ func (*mockNet) Discover(provides types.ProvidesCoin, searchTime time.Duration)
...
-		Offers: []*types.Offer{
-			{ID: testSwapID},
-		},
-	}, nil
+	var offer types.Offer
+	offerJSON := fmt.Sprintf(`{"ID":%q}`, testSwapID.String())
+	if err := json.Unmarshal([]byte(offerJSON), &offer); err != nil {
+		panic(err)
+	}
+	return &net.QueryResponse{Offers: []*types.Offer{&offer}}, nil
```
dimalinux, line 55: The non-test code doesn't have the need/API to set the offer ID to a specific value, but there was an API to create an offer with a specific ID from JSON, so I just used it here.

tests/integration_test.go
```diff
@@ -721,9 +721,13 @@ func TestError_ShouldOnlyTakeOfferOnce(t *testing.T) {
 	require.Equal(t, 1, len(providers))
 	require.GreaterOrEqual(t, len(providers[0]), 2)
 
-	errCh := make(chan error)
+	errCh := make(chan error, 2)
+
+	var wg sync.WaitGroup
+	wg.Add(2)
```
dimalinux, line 727: It was this test that was occasionally causing `TestSuccess_ConcurrentSwaps` after it to fail. The second attempt to take the offer was sometimes still alive after the test completed and recreated the offer ID here:
https://github.com/noot/atomic-swap/blob/master/protocol/xmrmaker/swap_state.go#L178-L189

common/types/offer.go
```diff
@@ -31,19 +36,41 @@ func HexToHash(s string) (Hash, error) {
...
+}
+
+// setID returns the ID of the offer
+func (o *Offer) setID() {
+	if !o.id.IsZero() {
+		panic("attempt to set offer ID more than once")
 	}
 
+	// TODO: If we are not saving the random nonce below, so we can't recreate the ID, why do we want to
+	//       include the serialised data when generating the ID?
```
noot, line 73: what do you mean by "why do we want to include the serialised data when generating the ID?" 

i guess my original intention was to have the offer ID be based on the contents, but multiple users could make the same offer so i added the random nonce. but yeah, generating a random ID number would be just as effective

protocol/xmrmaker/swap_state_test.go
noot: yeah, it wouldn't be bad to validate that the offer min/max amounts and exchange rate are nonzero

tests/integration_test.go
```diff
@@ -760,21 +766,23 @@ func TestError_ShouldOnlyTakeOfferOnce(t *testing.T) {
...
 			}
 
 			if status != types.CompletedSuccess {
-				errCh <- fmt.Errorf("swap did not exit successfully: got %s", status)
+				errCh <- fmt.Errorf("2nd swap did not exit successfully: got %s", status)
 				return
 			}
 
-			t.Log("> XMRTaker exited successfully")
+			t.Log("> XMRTaker[1] exited successfully")
```
noot, line 779:
```suggestion
			t.Log("> XMRTaker[2] exited successfully")
```

also meganit but i'd index them as 0 and 1 lol, or just by name

noot (approved):
really good stuff, thanks for looking into these issues!

common/types/offer.go
dimalinux: Yes, I was wondering why we serialized any data. If the step is not adding any value, it would make sense to strip it out.

PR #149, opened 2022-08-11
Update go ethereum

dimalinux:
There are other package deps that should get updated, but this PR's primary purpose is updating `go-ethereum`. As part of the update we are now:
* Using the version of `abigen` that automatically matches the version of our `go-ethereum` dependency
* `solc` was updated to 0.8.16 (in the docs, we don't install it, but anyone wanting to generate identical output needs it).
* Minimum go version is now 1.18
* `golangci-lint` updated to v1.48.0 (was needed to play nicely with go 1.18 and the new generated code)
* `dleq/cgo-dleq` submodule updated to latest version


Makefile
```diff
@@ -10,7 +10,7 @@ init:
 	cd dleq/cgo-dleq && make build
 
 .PHONY: lint
-lint: init
```
dimalinux, line 13: We'll just assume people run `make init` before running `make lint`.  I run the command a lot and it was too noisy and time consuming to have those other scripts run.

docs/developing.md
```diff
@@ -53,26 +53,26 @@ This will install Rust (if it isn't already installed) and build the binaries. T
 
 ## Compiling contract bindings
 
-If you update the `Swap.sol` contract for some reason, you will need to re-generate the Go bindings for the contract. **Note:** you do *not* need to do this to try out the swap; only if you want to edit the contract for development purposes.
+If you update the `Swap.sol` contract for some reason, you will need to re-generate the Go bindings
+for the contract. **Note:** you do *not* need to do this to try out the swap; only if you want to
+edit the contract for development purposes.
```
dimalinux, line 58: This block didn't change.  It was all on one line, so I added some newlines.

ethereum/block/ut_contract_test.go
```diff
@@ -31,20 +31,13 @@ var (
 // UTContractMetaData contains all meta data concerning the UTContract contract.
```
dimalinux, line 31: This is a generated file

ethereum/contracts/SwapFactory.sol
```diff
@@ -91,12 +91,6 @@ contract SwapFactory is Secp256k1 {
         emit Ready(swapID);
     }
 
-    // is_ready returns whether a swap has been set to "ready" or not.
-    // note: it will return false, not revert, if the swap does not exist.
-    function is_ready(bytes32 _swapID) public view returns (bool) {
-        return swaps[_swapID] == Stage.READY;
-    }
```
dimalinux, line 98: I didn't want to remove this from the contact until we had a reason to regenerate the bindings, which this PR does.

tests/integration_test.go
```diff
@@ -366,9 +366,9 @@ func TestRefund_XMRMakerCancels_untilAfterT1(t *testing.T) {
 	time.Sleep(time.Second * 5)
 }
 
-// TestRefund_XMRMakerCancels_afterIsReady tests the case where XMRTaker and XMRMaker both lock their funds,
-// but XMRMaker goes offline until past isReady==true and t0, but comes online before t1.
-//  When XMRMaker comes back online, he should claim the ETH, causing XMRTaker to also claim the XMR.
+// TestRefund_XMRMakerCancels_afterIsReady tests the case where XMRTaker and XMRMaker both lock their
+// funds, but XMRMaker goes offline until past isReady==true and t0, but comes online before t1. When
+// XMRMaker comes back online, he should claim the ETH, causing XMRTaker to also claim the XMR.
 func TestRefund_XMRMakerCancels_afterIsReady(t *testing.T) {
```
dimalinux, line 372: The newer version of `golangci-lint` caught the extra space before `When` above.

Makefile
```diff
@@ -1,14 +1,17 @@
...
 
-.PHONY: init
-init:
+$(DLEQ_LIB):
 	./scripts/install-rust.sh
 	git submodule update --init --recursive
 	cd dleq/cgo-dleq && make build
 
+.PHONY: init
+init: $(DLEQ_LIB)
```
dimalinux, line 13: I made the `init` target depend on the existence of `libdleq.so`, then the target commands previously tied to `init` will only be executed if the shared library does not exist.  That prevents those commands from being run every single time you execute commands like `make lint`. To force the rebuild/submodule-update, you can run `make clean init`.

Makefile
```diff
@@ -24,7 +27,7 @@ test-integration:
 
 .PHONY: install
 install: init
-	cd cmd/ && go install && cd ..
```
dimalinux, line 27: Every individual command in a target is run in a separate subshell, so you don't need to restore the directory at the end.

docs/developing.md
```diff
@@ -53,26 +53,26 @@ This will install Rust (if it isn't already installed) and build the binaries. T
...
 ```
 
-Install `abigen`
-```
-git clone https://github.com/ethereum/go-ethereum.git && cd go-ethereum/cmd/abigen
-go install
-```
+We install the `abigen` into the `bin` directory of your GOPATH (`$HOME/go/bin` for most users).
+The version installed is matched to the go-ethereum version that the project currently links with.
+See `scripts/install-abigen.sh` for details.
```
dimalinux, line 70: We already install `golangci-lint` into `${HOME}/go/bin`, so I thought this behavior wouldn't be too unexpected. The other option would be to install it somewhere inside the project like we do with the monero binaries.

Makefile
noot: much nicer, thanks!

noot (approved):
works for me!

PR #151, opened 2022-08-14
Update libp2p

dimalinux:
Updates our libp2p dependencies to the latest stable versions.


net/discovery.go
```diff
@@ -10,9 +10,9 @@ import (
 	libp2phost "github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/peerstore"
-	libp2pdiscovery "github.com/libp2p/go-libp2p-discovery"
```
dimalinux, line 13: This is what prevented the upgrade from being trivial. The repo is deprecated and the code was moved into packages of `github.com/libp2p/go-libp2p`:  https://github.com/libp2p/go-libp2p-discovery/commit/111203389c4886581007b0837ba9d48821f4fb6f

net/host.go
```diff
@@ -140,7 +140,7 @@ func NewHost(cfg *Config) (*host, error) {
 	}
 
 	// create libp2p host instance
-	h, err := libp2p.New(context.Background(), opts...)
```
dimalinux, line 143: I guess the context wasn't getting used: https://github.com/libp2p/go-libp2p/commit/b7bee3855cb86e50440e23b463605ea874c38787

net/host.go
```diff
@@ -177,7 +177,6 @@ func (h *host) Start() error {
 	h.h.SetStreamHandler(protocol.ID(h.protocolID+queryID), h.handleQueryStream)
 	h.h.SetStreamHandler(protocol.ID(h.protocolID+swapID), h.handleProtocolStream)
 
-	h.h.Network().SetConnHandler(h.handleConn)
```
dimalinux, line 180: `SetConnHandler` was removed: https://github.com/libp2p/go-libp2p-core/pull/217
We were only using it to log a message, so I just removed it from our code too.

noot (approved):
looks good! would be nice to eventually add that incoming connection log back in, but I think we will need to implement a libp2p ConnManager for that.

PR #152, opened 2022-08-14
update stagenet script

noot:
update script so that it's for running a single node on stagenet, fix typo


docs/stagenet.md
```diff
@@ -14,13 +14,13 @@ The atomic swap daemon requires a connection to a monero-wallet-rpc process conn
 
 For Linux 64-bit, you can do:
 ```bash
-curl -L https://downloads.getmonero.org/cli/linux64 > monero.tar.bz2
-tar xjvf monero.tar.bz2
+bash scripts/install-monero-linux.sh
```
dimalinux, line 17: The script has execute permissions (which git respects/maintains) and it starts with `#!/bin/bash`. Any reason to not execute it directly?  i.e. `./scripts/install-monero-linux.sh`?

scripts/setup-stagenet.sh
```diff
@@ -7,17 +7,14 @@ echo "starting monerod..."
 sleep 5
 
 echo "starting monero-wallet-rpc on port 18083..."
-nohup ./monero-bin/monero-wallet-rpc --rpc-bind-port 18083 --disable-rpc-login --wallet-dir ./bob-test-keys --stagenet --trusted-daemon
+nohup ./monero-bin/monero-wallet-rpc --rpc-bind-port 18083 --disable-rpc-login --wallet-dir ./node-keys --stagenet --trusted-daemon &> monero-wallet-cli.log &
```
dimalinux, line 10: Should we try the `--detach` and `--log-file` flags to `monero-wallet-rpc`?

dimalinux (approved):

docs/stagenet.md
noot: good point, updated!

scripts/setup-stagenet.sh
noot: that would be nice, didn't realize it had a `--detach` flag as well. I can update all of the scripts in a follow-up

PR #155, opened 2022-08-19
validate monero addresses based on network

### LINKED ISSUE #136: validate externally submitted xmr addresses based on network
eg. when swapping from the ui, if the wrong type of address is used, the swap will not error with the right error as address validation based on network is not happening immediately.

noot:
closes #136


crypto/monero/address.go
```diff
@@ -15,15 +15,31 @@ const (
...
+
+	switch env {
+	case common.Mainnet, common.Development:
+		if b[0] != addressPrefixMainnet {
+			return errInvalidPrefixGotStagenet
+		}
+	case common.Stagenet:
+		if b[0] != addressPrefixStagenet {
+			return errInvalidPrefixGotMainnet
+		}
```
dimalinux, line 42: After the PR enabling the full wallet API, this will be another option:
https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#validate_address

PR #158, opened 2022-08-19
add direct ERC-20 swap support

### LINKED ISSUE #61: implement ability to directly swap for ERC20 tokens
investigate extending the ETH contract to allow for direct swaps for ERC20 tokens. this can probably be achieved by firsltly approving the contact to spend some amount of tokens, then transferring them on the `Claim` step.

doonte:
Fixes #61

This is a work in progress pr

Plan:

1. add ERC-20 support to contract
2. get regular eth swaps working with new contract
3. add to command line/ui

Considerations:
- Storing tokens in the contract guarantees an exact swap
- Fee-on-transfer and rebasing tokens are problematic, since the amount of the token stored in the smart contract may not be equal to the amount originally sent. We can either try to prevent them from being used or store "shares" held by each sender the same way wrapped tokens work
- More centralized tokens stored in the contract (Tether, USDC) could potentially be blacklisted/censored, so suggest more decentralized tokens


cmd/client/main.go
```diff
@@ -86,6 +87,10 @@ var (
 						Name:  "subscribe",
 						Usage: "subscribe to push notifications about the swap's status",
 					},
+					&cli.StringFlag{
+						Name:  "eth-asset",
+						Usage: "Ethereum ERC-20 token address to receive, or the zero address for regular ETH",
+					},
```
noot, line 93: I think we probably want to do this per-swap (ie. in an offer advertisement) as opposed to configuring the entire daemon for it

common/types/ethasset.go
```diff
@@ -0,0 +1,18 @@
...
+type EthAsset ethcommon.Address
+
+func (asset EthAsset) String() string {
+	if ethcommon.Address(asset).Hex() == "0x0000000000000000000000000000000000000000" {
+		return "ETH"
+	}
+	return ethcommon.Address(asset).Hex()
+}
+
+// TODO: is there a cleaner way to create a constant like this?
```
noot, line 17: I think this is fine

common/types/offer_test.go
```diff
@@ -5,17 +5,18 @@ import (
...
 
 func TestOffer_MarshalJSON(t *testing.T) {
-	offer := NewOffer(ProvidesXMR, 100.0, 200.0, 1.5)
+	offer := NewOffer(ProvidesXMR, 100.0, 200.0, 1.5, EthAssetETH)
 	id := offer.GetID()
 	require.False(t, id.IsZero())
 
 	expected := fmt.Sprintf(
-		`{"ID":"%s","Provides":"XMR","MinimumAmount":100,"MaximumAmount":200,"ExchangeRate":1.5}`, id)
+		`{"ID":"%s","Provides":"XMR","MinimumAmount":100,"MaximumAmount":200,"ExchangeRate":1.5,"EthAsset":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}`, id)
```
noot, line 19: you can implement `MarshalJSON` on the `EthAsset` type so that it marshals to a hex string, see https://pkg.go.dev/encoding/json#example-package-CustomMarshalJSON

common/rpctypes/types.go
```diff
@@ -53,6 +53,7 @@ type MakeOfferRequest struct {
 	MinimumAmount float64            `json:"minimumAmount"`
 	MaximumAmount float64            `json:"maximumAmount"`
 	ExchangeRate  types.ExchangeRate `json:"exchangeRate"`
+	EthAsset      types.EthAsset     `json:"ethAsset"`
```
noot, line 56:
```suggestion
	EthAsset      types.EthAsset     `json:"ethAsset,omitempty"`
```

this makes it optional, so that we can default to ETH if it's not set

docs/rpc.md
```diff
@@ -64,6 +64,7 @@ Parameters:
 - `minimumAmount`: minimum amount to swap, in XMR.
 - `maximumAmount`: maximum amount to swap, in XMR.
 - `exchangeRate`: exchange rate of ETH-XMR for the swap, expressed in a fraction of XMR/ETH. For example, if you wish to trade 10 XMR for 1 ETH, the exchange rate would be 0.1.
+- `ethAsset`: Ethereum asset to trade, either an ERC-20 token address or the zero address for regular ETH
```
noot, line 67: mention that it's optional (once it's made optional)

ethereum/contracts/SwapFactory.sol
```diff
@@ -59,25 +63,37 @@ contract SwapFactory is Secp256k1 {
...
         swap.pubKeyClaim = _pubKeyClaim;
         swap.pubKeyRefund = _pubKeyRefund;
+        swap.claimer = _claimer;
         swap.timeout_0 = block.timestamp + _timeoutDuration;
         swap.timeout_1 = block.timestamp + (_timeoutDuration * 2);
-        swap.value = msg.value;
+        swap.asset = _asset;
+        swap.value = _value;
+        if (swap.asset == address(0)) {
+            require(swap.value == msg.value, "value not same as ETH amount sent");
```
noot, line 81: alternatively, we can allow `_value` to be 0 for `swap.asset == address(0)` so we can remove this check and ignore `_value` for eth

ethereum/contracts/SwapFactory.sol
```diff
@@ -106,7 +122,17 @@ contract SwapFactory is Secp256k1 {
...
-        _swap.claimer.transfer(_swap.value);
+        if (_swap.asset == address(0)) {
+            _swap.claimer.transfer(_swap.value);
+        } else {
+            // TODO: this will FAIL for fee-on-transfer or rebasing tokens if the token
+            // transfer reverts (i.e. if this contract does not contain _swap.value tokens),
+            // exposing Bob's secret while giving him nothing
+            
+            // potential solution: wrap tokens into shares instead of absolute values
+            // swap.value would then contain the share of the token
```
noot, line 133: can you expand on this?

protocol/xmrmaker/swap_state.go
```diff
@@ -448,14 +448,25 @@ func (s *swapState) checkContract(txHash ethcommon.Hash) error {
...
 	}
 
-	event, err := s.Contract().ParseNew(*receipt.Logs[0])
+	var event *swapfactory.SwapFactoryNew
+	for _, log := range receipt.Logs {
+		event, err = s.Contract().ParseNew(*log)
+		if err == nil {
+			break
+		}
+	}
```
noot, line 467: is there a case where there would be multiple logs in one receipt? I don't think there would be?

protocol/xmrtaker/swap_state.go
```diff
@@ -426,24 +430,40 @@ func (s *swapState) lockETH(amount common.EtherAmount) (ethcommon.Hash, error) {
...
-	log.Debugf("instantiated swap on-chain: amount=%s txHash=%s", amount, txHash)
+	log.Debugf("instantiated swap on-chain: amount=%s asset=%v txHash=%s", amount, s.ethAsset, txHash)
 
 	if len(receipt.Logs) == 0 {
 		return ethcommon.Hash{}, errSwapInstantiationNoLogs
 	}
 
-	s.contractSwapID, err = swapfactory.GetIDFromLog(receipt.Logs[0])
+	log.Debugf("About to get the id from the log")
+	log.Debugf("log: %v", log)
```
noot, line 445: remove? :p

protocol/xmrtaker/swap_state.go
```diff
@@ -426,24 +430,40 @@ func (s *swapState) lockETH(amount common.EtherAmount) (ethcommon.Hash, error) {
...
+	for _, rLog := range receipt.Logs {
+		s.contractSwapID, err = swapfactory.GetIDFromLog(rLog)
+		if err == nil {
+			break
+		}
+	}
 	if err != nil {
```
noot, line 452: same here, is there a case where there could be multiple logs?

protocol/xmrtaker/swap_state.go
```diff
@@ -426,24 +430,40 @@ func (s *swapState) lockETH(amount common.EtherAmount) (ethcommon.Hash, error) {
...
 
-	t0, t1, err := swapfactory.GetTimeoutsFromLog(receipt.Logs[0])
+	var t0 *big.Int
+	var t1 *big.Int
+	for _, log := range receipt.Logs {
+		t0, t1, err = swapfactory.GetTimeoutsFromLog(log)
+		if err == nil {
+			break
+		}
+	}
```
noot, line 464: and here

swapfactory/swap_factory_test.go
```diff
@@ -44,7 +44,7 @@ func TestSwapFactory_NewSwap(t *testing.T) {
 
 	nonce := big.NewInt(0)
 	tx, err = contract.NewSwap(auth, [32]byte{}, [32]byte{},
-		ethcommon.Address{}, defaultTimeoutDuration, nonce)
+		ethcommon.Address{}, defaultTimeoutDuration, ethcommon.HexToAddress("0000000000000000000000000000000000000000"), big.NewInt(0), nonce)
```
noot, line 47:
```suggestion
		ethcommon.Address{}, defaultTimeoutDuration, types.EthAssetETH, big.NewInt(0), nonce)
```

same for all others in this file?

protocol/xmrmaker/swap_state.go
doonte: Yes, when a token transfer happens, it emits a Transfer event. And also, if the claim is made as an internal contract call, like from within a different smart contract, it may also have other log events.

doonte:
Something we should take into account is the number of decimal places that the ERC-20 token has, which is not always 18.

noot (approved):
looks good to me, thanks so much for the contribution! will need to follow this up asap with #180 (more of a note for me potentially, lol)

PR #170, opened 2022-08-24
New bash library, and well as bash linting and formatting

dimalinux:
This PR allows the unit and integration tests to share the same bash code via a utility library named `testlib.sh`.  The library is designed so that users can source it in their bash shell without actually launching any daemons, and launch the daemons they need on an as-needed basis.

Linting and formatting of bash scripts was added to the Makefile targets `lint` and `format`.  To use these targets, you'll need to run `sudo apt install -y shfmt shellcheck` to install the bash formatter/linter respectively (mentioned in `developer.md` as well).


.gitignore
```diff
@@ -1,56 +1,32 @@
...
+# Output of the go coverage tool
+/coverage.out
 /coverage.html
 
-# Dependency directories (remove the comment below to include it)
-# vendor/
-
-### Go Patch ###
-/vendor/
-/Godeps/
```
dimalinux, line 34: I didn't like having all the boilerplate. If a `vendor` folder showed up in the project, it is something we'd want to know about.  There are no rust artifacts in the top-level git repo, so I cleared them out.

Makefile
```diff
@@ -16,6 +16,7 @@ init: $(DLEQ_LIB)
 lint: init
 	./scripts/install-lint.sh
 	${GOPATH}/bin/golangci-lint run
+	shellcheck scripts/*.sh
```
dimalinux, line 19: We need linting/formatting for the various source types in the `ui` folder too, but I'm not taking that on in this PR.

rpc/server.go
```diff
@@ -79,9 +79,9 @@ func (s *Server) Start() <-chan error {
 		methodsOk := handlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "OPTIONS"})
 		originsOk := handlers.AllowedOrigins([]string{"*"})
 
-		log.Infof("starting RPC server on http://localhost:%d", s.port)
+		log.Infof("starting RPC server on http://127.0.0.1:%d", s.port)
 
-		if err := http.ListenAndServe(fmt.Sprintf(":%d", s.port), handlers.CORS(headersOk, methodsOk, originsOk)(r)); err != nil { //nolint:lll
```
dimalinux, line 84: The logged message was saying the port is only accessible from  `localhost`, while the actual `ListenAndServe` call was using all interfaces. I changed the code to match the logged message.

scripts/build-macos.sh
```diff
@@ -1,41 +1,43 @@
...
 
 echo "building swapcli..."
-cd ../client 
-if ! go build -o swapcli-amd64-darwin ; then 
+cd ../client || exit 1
+if ! go build -o swapcli-amd64-darwin; then
 	exit 1
 fi
 mv swapcli ../..
 echo "done building swapcli."
```
dimalinux, line 23: I can simplify and consolidate a massive amount build code if we rename the folders in `cmd` to match the name of the executables they build. I didn't want to increase the diff in this PR though.

scripts/build-macos.sh
```diff
@@ -1,41 +1,43 @@
...
 
 if [[ -z "${ALL}" ]]; then
-  exit 0
+	exit 0
 fi
 
 echo "build swaprecover..."
-cd ../recover
-if ! go build -o swaprecover-amd64-darwin ; then 
+cd ../recover || exit 1
```
dimalinux, line 30: It was either add the `exit 1`, or add a comment to tell the linter to ignore it.  None of these build scripts are sourced (verses executed), so the exit should be safe.  Most of these lines will go away if we do the directory renames that I mentioned above.

scripts/generate-bindings.sh
```diff
@@ -12,8 +13,9 @@ fi
...
-            --bin ethereum/bin/SwapFactory.bin \
-            --pkg swapfactory \
-            --type SwapFactory \
-            --out swapfactory/swap_factory.go
+"${ABIGEN}" \
+	--abi ethereum/abi/SwapFactory.abi \
+	--bin ethereum/bin/SwapFactory.bin \
+	--pkg swapfactory \
+	--type SwapFactory \
+	--out swapfactory/swap_factory.go
```
dimalinux, line 21: No real changes here, just getting the code to line up when auto-formatted.

scripts/setup-devnet.sh
```diff
@@ -13,37 +13,37 @@ killall -v swapd
 pkill -e -f swapcli
 
 echo "start ganache"
-"$(npm config get prefix)/bin/ganache" --deterministic --accounts=50 --miner.blockTime=1 &> "${LOG_DIR}/ganache.log" &
+"$(npm config get prefix)/bin/ganache" --deterministic --accounts=50 --miner.blockTime=1 &>"${LOG_DIR}/ganache.log" &
```
dimalinux, line 16: I haven't used this script and don't entirely know what it is for, so I didn't try to simplify it.  All the changes here were from the formatter or to make the linter happy.

Makefile
noot: do we need shellcheck here? I'd rather put it under `lint` so that we don't force users to install it

scripts/build-macos.sh
noot: feel free to do this in a follow up!

scripts/setup-devnet.sh
noot: feel free to delete this script, it's a duplicate of what setup-env.sh was

noot (commented):
basically looks good, really nice cleanup, highly appreciated!

could you add back `setup-env.sh`? I suppose it would just need to have 
```
source "scripts/testlib.sh"
start-monerod-regtest
start-ganache
start-alice-wallet
start-bob-wallet
```

I used that script pretty often for setting up an env for manual testing.

Makefile
dimalinux: It is under the `lint` target right now, which only developers would be invoking. After reading your comment, I made the `lint` and `format` targets more granular.  If you are only changing Go code, you can now invoke `make lint-go` to only check the go code without installing `shellcheck`.  CI will hopefully run the go checks, shell checks and future UI code checks.

scripts/setup-devnet.sh
dimalinux: Deleted.  Thanks!

.gitignore
```diff
@@ -1,56 +1,32 @@
...
+/test-integration.log
 
-*.key
-log
-*.log
+# Apparently, swapd will generate libp2p keys with the names below into the current working
+# directory (happens during unit tests too).
+node.key
+xmrmaker.key
+xmrtaker.key
```
dimalinux, line 32: @noot Is this what should be happening.  Do we need a ticket for this?  Where do other programs using libp2p store these keys?

noot: maybe we can put them in the temp dir instead, if they're just for tests? usually for a long-running node, they'll be in the basepath/datadir (which is `~/.atomicswap` for us)

Makefile
noot: sorry, i think i misread it previously. this looks good to me!

noot (approved):
lgtm! just a reminder about my previous comment (re-adding `setup-env.sh` before merging), otherwise this is fantastic :)

PR #173, opened 2022-08-24
persist peerstore to disk

### LINKED ISSUE #84: persist peerstore to disk
the peerstore should be persisted to disk so when we restart we don't have to discover all our peers again. this is pretty easy to do with the libp2p ds-badger library/

noot:
persist peerstore to disk so that peers are saved between restarts.

also lint, go mod tidy

closes #84


net/host_test.go
```diff
@@ -59,9 +59,13 @@ func (s *mockSwapState) Exit() error {
 }
 
 func newHost(t *testing.T, port uint16) *host {
+	dir, err := os.MkdirTemp(os.TempDir(), "atomicswap*")
+	require.NoError(t, err)
```
dimalinux, line 63: I'd replace these 2 lines with:
```
	dir := t.TempDir()
```
You'll get a unique directory that includes the test's name, and the directory will be cleaned up on success.

net/host_test.go
```diff
@@ -71,6 +75,9 @@ func newHost(t *testing.T, port uint16) *host {
 
 	h, err := NewHost(cfg)
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		_ = h.Stop()
```
dimalinux, line 79: Unless there are known times when we expect this to give an error, I'd change it to:
```
assert.NoError(t, h.Stop())
```
Using `assert` instead of `require` above, because you would want the test to continue.

net/host_test.go
noot: nice, updated!

PR #174, opened 2022-08-26
don't advertise 127.0.0.1 on network unless in development mode

### LINKED ISSUE #167: don't advertise local addresses for non-development environments
currently, a multiaddress with 127.0.0.1 is advertised on the network for all environments, but it only should be for development. update this so it's not advertised for "real" environments.

noot:
closes #167


net/filters.go
```diff
@@ -9,6 +9,7 @@ import (
 func newPrivateIPFilters() (privateIPs *ma.Filters, err error) {
 	privateCIDRs := []string{
 		"10.0.0.0/8",
+		"127.0.0.1/0",
```
dimalinux, line 12: Did that work?  If it did, the library must be treating `/0` as if the user wanted `/32`.  The entire `127.0.0.0/8` space is reserved, so that's what I would put down. Lots of linux hosts have multiple loopbacks for various reasons.

noot: right, yeah i think this only worked because the external address gets manually added after: https://github.com/AthanorLabs/atomic-swap/blob/master/net/host.go#L148 will update, thanks for catching this!

PR #184, opened 2022-09-06
fix: add wrapped TxOpts type

noot:
add wrapped `TxOpts` type to fix multiple txsender concurrency issues, since the last PR (#158) made it so that the txsender was per-swap instead of global.

### DIFF
--- a/protocol/backend/backend.go
+++ b/protocol/backend/backend.go
@@ -108,6 +108,8 @@ type backend struct {
 	gasPrice   *big.Int
 	gasLimit   uint64
 
+	txOpts *txsender.TxOpts
+
 	// swap contract
 	contract     *swapfactory.SwapFactory
 	contractAddr ethcommon.Address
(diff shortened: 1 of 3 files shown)


protocol/backend/backend.go
```diff
@@ -150,9 +152,19 @@ func NewBackend(cfg *Config) (Backend, error) {
...
-	var addr ethcommon.Address
+	var (
+		addr   ethcommon.Address
+		txOpts *txsender.TxOpts
+		err    error
+	)
 	if cfg.EthereumPrivateKey != nil {
 		addr = common.EthereumPrivateKeyToAddress(cfg.EthereumPrivateKey)
+
+		// TODO: set gas limit and price?
```
dimalinux, line 163: I was actually thinking that we should eliminate the `--gas-limit` flag from `swapd`.  The estimation works good, so I was thinking that it is just unnecessary complexity. In order to use a fixed gas limit across a wide array of transaction types requiring different amounts of gas, the user would have to set a gas limit much higher than what most individual transactions require. Are there other newish (non-legacy) programs with a similar gas limit option for multiple transaction types?

noot: the reason I originally added it was because certain networks (L2s primarily) don't work with geth's gas limit estimation, as geth always will estimate it too low, and the transactions always fail. I think we could probably put the gas limit in a per-network config, instead of having it as a CLI flag.

protocol/txsender/tx_opts.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &TxOpts{
+		inner: txOpts,
+	}, nil
+}
+
+// Inner returns the bind.TransactOpts contained
```
dimalinux, line 29: I little more context might be helpful to readers. One possible idea:

Inner returns a copy of bind.TransactOpts. The original is never used in a transaction, so the copies given out do not have the Nonce, and Gas* fields initialized allowing them to be initialized dynamically. Reference:
https://github.com/ethereum/go-ethereum/blob/v1.10.23/accounts/abi/bind/base.go#L49-L63

noot: updated!

protocol/backend/backend.go
dimalinux: Do those networks have a fork of the `go-ethereum` libraries that can do correct estimation?

dimalinux (approved):

PR #185, opened 2022-09-07
CLI cleanups

dimalinux:
- Switches to current v2 version of urfave CLI
- Fixes incorrect port values in the help
- Provides defaults directly to urfave for clear `--help` messaging
- `--base-path` flag renamed to `--data-dir`. This matches `monerod`'s' `--data-dir` and almost matches the `geth` flag `--datadir`
- Bootnodes and contract address are provided for stagenet testing
- Bootnodes can be provided individually by repeating the `--bootnodes`/`--bn` flag, or as a comma separated block (original behavior), or even a combination of both. (Makes adding/removing bootnodes a lot easier.)
- More precise error messaging when XMR maker/taker values are off
- Fixes bug in the peer finder where XMR was hardcoded as the provides coin
- `cleanup-test-processes.sh` displays which processes are being killed


cmd/client/errors.go
```diff
@@ -5,10 +5,9 @@ import (
 )
 
 var (
-	errNoMultiaddr      = errors.New("must provide peer's multiaddress with --multiaddr")
 	errNoMinAmount      = errors.New("must provide non-zero --min-amount")
 	errNoMaxAmount      = errors.New("must provide non-zero --max-amount")
 	errNoExchangeRate   = errors.New("must provide non-zero --exchange-rate")
-	errNoOfferID        = errors.New("must provide --offer-id")
```
dimalinux, line 12: The `--offer-id` is now marked as required by `urfave`, so this error wasn't used anymore.

cmd/daemon/contract.go
```diff
@@ -24,9 +24,15 @@ var (
...
 )
 
-func getOrDeploySwapFactory(ctx context.Context, address ethcommon.Address, env common.Environment,
-	basePath string, chainID *big.Int, privkey *ecdsa.PrivateKey,
-	ec *ethclient.Client) (*swapfactory.SwapFactory, ethcommon.Address, error) {
+func getOrDeploySwapFactory(
+	ctx context.Context,
+	address ethcommon.Address,
+	env common.Environment,
+	dataDir string,
```
dimalinux, line 31: `basePath` -> `dataDir` is the only change here.

cmd/daemon/main.go
```diff
@@ -92,35 +94,41 @@ var (
 		Flags: []cli.Flag{
 			&cli.UintFlag{
 				Name:  flagRPCPort,
-				Usage: "port for the daemon RPC server to run on; default 5001",
+				Usage: "Port for the daemon RPC server to run on",
+				Value: defaultRPCPort,
```
dimalinux, line 98: There were a bunch flags reporting incorrect values and multiple defaults. Using the `Value` field (default value), will make the code easier to maintain.

Old version:
```
--rpc-port value     port for the daemon RPC server to run on; default 5001 (default: 0)
```
New Version:
```
--rpc-port value     Port for the daemon RPC server to run on (default: 5005)
```

cmd/daemon/main.go
```diff
@@ -92,35 +94,41 @@ var (
...
 				Name:  flagWSPort,
-				Usage: "port for the daemon RPC websockets server to run on; default 8080",
+				Usage: "Port for the daemon RPC websockets server to run on",
+				Value: defaultWSPort,
 			},
 			&cli.StringFlag{
-				Name:  flagBasepath,
-				Usage: "path to store swap artefacts",
+				Name:  flagDataDir,
+				Usage: "Path to store swap artifacts", //nolint:misspell
```
dimalinux, line 107: I looked in a Canadian dictionary and it said that Canadians spell artifacts the same as Americans.  :) In a computer context, I suspect even people in the UK use the above spelling.

cmd/daemon/main.go
```diff
@@ -132,55 +140,57 @@ var (
...
-				Usage: "comma-separated string of libp2p bootnodes",
+			&cli.StringSliceFlag{
+				Name:    flagBootnodes,
+				Aliases: []string{"bn"},
+				Usage:   "libp2p bootnode, comma separated if passing multiple to a single flag",
 			},
 			&cli.UintFlag{
 				Name:  flagGasPrice,
-				Usage: "ethereum gas price to use for transactions (in gwei). if not set, the gas price is set via oracle.",
+				Usage: "Ethereum gas price to use for transactions (in gwei). If not set, the gas price is set via oracle.",
```
dimalinux, line 164: Capitalizing or not is a matter of taste.  `monerod` and `geth` both capitalize the first word of flag descriptions (unless it is a proper noun that starts with a lower case letter like `libp2p` above). We had a few places with complete sentences like this one, so I thought capitalizing was a better match.

cmd/daemon/main.go
```diff
@@ -414,24 +425,46 @@ func (d *daemon) make(c *cli.Context) error {
...
-		cfg.Basepath,
-	)
+	log.Infof("started swapd with data-dir %s", cfg.DataDir)
 	return nil
 }
 
-func newBackend(ctx context.Context, c *cli.Context, env common.Environment, cfg common.Config,
-	chainID int64, devXMRMaker bool, sm swap.Manager, net net.Host) (backend.Backend, error) {
+func errFlagsMutuallyExclusive(flag1, flag2 string) error {
+	return fmt.Errorf("flags %q and %q are mutually exclusive", flag1, flag2)
```
dimalinux, line 433: I wasn't sure whether to quote the flag name or use a leading `--`. If urfave generates the error for a required flag, it is using quoting like I did above. In any case, it's in a function so we can change it in one place if people don't like it.

cmd/daemon/main.go
```diff
@@ -414,24 +425,46 @@ func (d *daemon) make(c *cli.Context) error {
...
+	return fmt.Errorf("required flag %q not specified", flag)
+}
+
+func newBackend(
+	ctx context.Context,
+	c *cli.Context,
+	env common.Environment,
+	cfg common.Config,
+	devXMRMaker bool,
+	devXMRTaker bool,
```
dimalinux, line 446: We are passing both `devXMRMaker` and `devXMRTaker` now, as it is possible to be in a dev environment and be neither one of these. Before, code was assuming that you must be `devXMRTaker` if you are not `devXMRMaker` in a dev environment. Longer term, I'm thinking we can eliminate these flags and just use config files with the same values set.

cmd/daemon/main_test.go
```diff
@@ -4,69 +4,57 @@ import (
...
 
 	"github.com/stretchr/testify/require"
-	"github.com/urfave/cli"
+	"github.com/urfave/cli/v2"
 )
 
-func newTestContext(t *testing.T, description string, flags []string, values []interface{}) *cli.Context {
-	require.Equal(t, len(flags), len(values))
-
+func newTestContext(t *testing.T, description string, flags map[string]any) *cli.Context {
```
dimalinux, line 13: I moved us from using 2 separate arrays with synced indices to a map triggering a minor rewrite below.

cmd/recover/main.go
```diff
@@ -21,15 +21,13 @@ import (
...
-
-	logging "github.com/ipfs/go-log"
 )
 
 const (
 	flagEnv                  = "env"
 	flagMoneroWalletEndpoint = "monero-endpoint"
 	flagEthereumEndpoint     = "ethereum-endpoint"
-	flagEthereumPrivateKey   = "ethereum-privkey"
+	flagEthereumPrivKey      = "ethereum-privkey"
```
dimalinux, line 30: Updated this to match the corresponding flag in `swapd`.

cmd/recover/main_test.go
```diff
@@ -157,10 +149,12 @@ func TestRecover_withXMRMakerSecretAndContract(t *testing.T) {
...
-		[]string{flagXMRMaker, flagInfoFile},
-		[]interface{}{
-			true,
-			infoFilePath,
+		map[string]any{
+			flagEnv:                  "dev",
+			flagXMRMaker:             true,
+			flagInfoFile:             infoFilePath,
+			flagEthereumPrivKey:      createEthPrivKeyFile(t, common.DefaultPrivKeyXMRMaker),
+			flagMoneroWalletEndpoint: tests.CreateWalletRPCService(t),
```
dimalinux, line 157: Tests like this one had to change due to flags being marked as required.

cmd/utils/utils.go
```diff
@@ -1,63 +1,52 @@
...
+) {
+	if ethPrivKeyFile != "" {
+		fileData, err := os.ReadFile(filepath.Clean(ethPrivKeyFile))
 		if err != nil {
-			return "", fmt.Errorf("failed to read ethereum-privkey file: %w", err)
-		}
-		ethPrivKeyHex = strings.TrimSpace(string(key))
-	} else {
-		if env != common.Development || useExternal {
-			log.Warnf("%s", errNoEthereumPrivateKey)
```
dimalinux, line 43: This code is shared by `swapd` which wants log style messages, and `swapcli` which does not want log style messages, so I eliminated the logging.  The private key flag is required now unless `--external-signer` is passed.

cmd/utils/utils.go
```diff
@@ -1,63 +1,52 @@
...
 		}
+		ethPrivKeyHex := strings.TrimSpace(string(fileData))
+		return ethcrypto.HexToECDSA(ethPrivKeyHex)
+	}
 
-		log.Warn("no ethereum private key file provided, using ganache deterministic key")
-		if devXMRMaker {
-			ethPrivKeyHex = common.DefaultPrivKeyXMRMaker
-		} else {
-			ethPrivKeyHex = common.DefaultPrivKeyXMRTaker
```
dimalinux, line 51: In our integration tests, we have Charlie, Alice and Bob. We don't want Charlie to be sharing non-sharable resources with Alice, so we don't want to assume `devXMRTaker` just because `devXMRMaker` was not set.

common/config.go
```diff
@@ -3,35 +3,49 @@ package common
...
 	EthereumChainID:      MainnetChainID,
 }
 
 // StagenetConfig is the monero stagenet and ethereum ropsten configuration
 var StagenetConfig = Config{
-	Basepath:             fmt.Sprintf("%s/.atomicswap/stagenet", homeDir),
+	DataDir:              fmt.Sprintf("%s/.atomicswap/stagenet", homeDir),
 	MoneroDaemonEndpoint: "http://127.0.0.1:38081/json_rpc",
-	EthereumChainID:      RopstenChainID,
+	EthereumChainID:      GorliChainID,
```
dimalinux, line 32: Users no longer have to pass the chain ID on the command line when doing stagenet testing with Goerli.

common/utils.go
```diff
@@ -32,13 +32,10 @@ func GetTopic(sig string) ethcommon.Hash {
 	return b
 }
 
-// MakeDir makes a directory
+// MakeDir creates a directory, including leading directories, if they don't already exist.
+// File permissions of created directories are only granted to the current user.
 func MakeDir(dir string) error {
-	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
```
dimalinux, line 37: `os.ModePerm` is a mask, so we want to avoid using it as a value which sets very open permissions. This is a great site if you ever need help with the octal values:
https://chmod-calculator.com/

net/host.go
```diff
@@ -218,7 +218,7 @@ func (h *host) logPeers() {
 	}
 }
 
-// close closes host services and the libp2p host (host services first)
+// Stop closes host services and the libp2p host (host services first)
 func (h *host) Stop() error {
```
dimalinux, line 222: Not sure why our linter didn't prevent this.

protocol/xmrmaker/errors.go
```diff
@@ -28,11 +29,55 @@ var (
...
+
+type errBalanceTooLow struct {
+	unlockedBalance float64
+	providedAmount  float64
+}
+
+func (e errBalanceTooLow) Error() string {
+	return fmt.Sprintf("balance of %s XMR is below provided %s XMR",
+		strconv.FormatFloat(e.unlockedBalance, 'f', -1, 64),
+		strconv.FormatFloat(e.providedAmount, 'f', -1, 64),
```
dimalinux, line 45: The `FormatFloat` call above eliminates trailing zeros after the decimal point which, apparently, there is no good way to do using `%` options to `Sprintf`.  In some places we are using `%v`, but that is not the same and can sometimes turn into exponent notation.  When we create a type for the float64 representation of XMR, maybe we can add a `String()` function so we don't have to repeat this code.

protocol/xmrtaker/net.go
```diff
@@ -45,8 +45,16 @@ func (a *Instance) initiate(providesAmount common.EtherAmount, receivedAmount co
 		return nil, errBalanceTooLow
 	}
 
-	s, err := newSwapState(a.backend, offerID, pcommon.GetSwapInfoFilepath(a.basepath), a.transferBack,
-		providesAmount, receivedAmount, exchangeRate, ethAsset)
+	s, err := newSwapState(
+		a.backend,
+		offerID,
+		pcommon.GetSwapInfoFilepath(a.dataDir),
```
dimalinux, line 51: `a.basepath` -> `a.dataDir` was the only change here.

tests/integration_test.go
```diff
@@ -67,7 +67,10 @@ func generateBlocks(num uint64) {
 	}
 
 	fmt.Println("> Generating blocks for test setup...")
-	_ = d.GenerateBlocks(xmrmakerAddr.Address, num)
+	err = d.GenerateBlocks(xmrmakerAddr.Address, num)
+	if err != nil {
+		panic(err)
+	}
```
dimalinux, line 73: `GenerateBlocks` doesn't fail often like it used to in earlier versions of our code.

cmd/daemon/main.go
```diff
@@ -216,14 +226,10 @@ func setLogLevels(c *cli.Context) error {
...
 	level := c.String(flagLog)
-	if level == "" {
-		level = levelInfo
-	}
-
 	switch level {
 	case levelError, levelWarn, levelInfo, levelDebug:
 	default:
-		return errors.New("invalid log level")
+		return fmt.Errorf("invalid log level %q", level)
```
noot, line 232: haven't seen `%q` before, nice!

cmd/daemon/main.go
noot: seems fine to me!

cmd/recover/main.go
noot: it would be good to move all the flags to their own package or something so we don't need to redefine them, as many of them are re-used over the various binaries. no need to do that here, but something to keep in mind

common/config.go
noot: nice :D

common/config.go
```diff
@@ -3,35 +3,49 @@ package common
...
 
 // MainnetConfig is the mainnet ethereum and monero configuration
 var MainnetConfig = Config{
-	Basepath:             fmt.Sprintf("%s/.atomicswap/mainnet", homeDir),
+	DataDir:              fmt.Sprintf("%s/.atomicswap/mainnet", homeDir),
 	MoneroDaemonEndpoint: "http://127.0.0.1:18081/json_rpc",
 	EthereumChainID:      MainnetChainID,
 }
 
 // StagenetConfig is the monero stagenet and ethereum ropsten configuration
```
noot, line 28:
```suggestion
// StagenetConfig is the monero stagenet and ethereum Gorli configuration
```

tests/integration_test.go
```diff
@@ -96,6 +99,9 @@ func generateBlocksAsync() {
 }
 
 func TestXMRTaker_Discover(t *testing.T) {
+	if os.Getenv(generateBlocksEnv) != falseStr {
+		generateBlocks(64)
+	}
```
noot, line 104: why do we need to generate blocks here?

noot (approved):
really nice improvements, thank you!

tests/integration_test.go
dimalinux: I was testing and I got an insufficient balance failure, so I went in and added some block generation to every test that was required a non-zero balance.  Thinking about your comment though, I'll take another pass at those changes in my next PR. Maybe we should eliminate the balance pre-generation before any tests are run?

PR #187, opened 2022-09-08
RPC and WS endpoints share the same port

dimalinux:
This PR combines our HTTP RPC and Websocket services into a single HTTP server, eliminating the `--ws-port` flag to `swapd`.  This lowers end-user complexity, especially when running multiple `swapd` servers on one host.  `swapcli` subscribe interactions are fixed with this PR and the `--daemon-addr http://127.0.0.1:PORT` flag to `swapcli` was removed. Instead of passing separate http and websocket URLs to `swapcli`, `swapcli` commands now use `--swapd-port PORT_NUM` allowing `swapcli` to interact with both services, while only requiring one flag from the end user.

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -1,22 +1,24 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"strings"
 
-	"github.com/ethereum/go-ethereum/common"
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/urfave/cli/v2"
 
+	"github.com/athanorlabs/atomic-swap/common"
 	"github.com/athanorlabs/atomic-swap/common/types"
 	"github.com/athanorlabs/atomic-swap/rpcclient"
 	"github.com/athanorlabs/atomic-swap/rpcclient/wsclient"
 )
 
 const (
-	defaultSwapdAddress           = "http://127.0.0.1:5001"
+	defaultSwapdPort              = 5001
 	defaultDiscoverSearchTimeSecs = 12
+
+	flagSwapdPort = "swapd-port"
 )
 
 var (
@@ -30,7 +32,7 @@ var (
 				Usage:   "List our daemon's libp2p listening addresses",
 				Action:  runAddresses,
 				Flags: []cli.Flag{
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -50,7 +52,7 @@ var (
 						Usage: "Duration of time to search for, in seconds",
 						Value: defaultDiscoverSearchTimeSecs,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -64,7 +66,7 @@ var (
 						Usage:    "Peer's multiaddress, as provided by discover",
 						Required: true,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -84,7 +86,7 @@ var (
 						Usage: "Duration of time to search for, in seconds",
 						Value: defaultDiscoverSearchTimeSecs,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -116,7 +118,7 @@ var (
 						Name:  "eth-asset",
 						Usage: "Ethereum ERC-20 token address to receive, or the zero address for regular ETH",
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -144,14 +146,14 @@ var (
 						Name:  "subscribe",
 						Usage: "Subscribe to push notifications about the swap's status",
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
 				Name:   "get-past-swap-ids",
 				Usage:  "Get past swap IDs",
 				Action: runGetPastSwapIDs,
-				Flags:  []cli.Flag{daemonAddrFlag},
+				Flags:  []cli.Flag{swapdPortFlag},
 			},
 			{
 				Name:   "get-ongoing-swap",
@@ -163,7 +165,7 @@ var (
 						Usage:    "ID of swap to retrieve info for",
 						Required: true,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -176,7 +178,7 @@ var (
 						Usage:    "ID of swap to retrieve info for",
 						Required: true,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -189,7 +191,7 @@ var (
 						Usage:    "ID of swap to retrieve info for",
 						Required: true,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -201,7 +203,7 @@ var (
 						Name:  "offer-id",
 						Usage: "ID of swap to retrieve info for",
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -213,7 +215,7 @@ var (
 						Name:  "offer-ids",
 						Usage: "A comma-separated list of offer IDs to delete",
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
@@ -226,7 +228,7 @@ var (
 						Usage:    "ID of swap to retrieve info for",
 						Required: true,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 			{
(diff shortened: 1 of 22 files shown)


cmd/daemon/main.go
```diff
@@ -413,20 +392,8 @@ func (d *daemon) make(c *cli.Context) error {
...
-	go func() {
-		select {
-		case <-d.ctx.Done():
-			return
-		case err := <-errCh:
-			log.Errorf("failed to start RPC server: %s", err)
-			d.cancel()
-			os.Exit(1)
-		}
-	}()
```
dimalinux, line 426: `Start()` now blocks and it internally uses the context that was passed to `rpcCfg`.

common/coins_test.go
```diff
@@ -29,7 +29,7 @@ func TestEtherAmount(t *testing.T) {
 
 func TestToDecimals(t *testing.T) {
 	val := NewEtherAmount(123456)
-	require.Equal(t, fmt.Sprint(val.ToDecimals(5)), "1.23456")
+	require.Equal(t, "1.23456", FmtFloat(val.ToDecimals(5)))
 	val = NewEtherAmount(1234567890)
-	require.Equal(t, fmt.Sprint(val.ToDecimals(6)), "1234.56789")
```
dimalinux, line 34: The expected/actual values here were reversed and it seemed like as good a place as any to use the `FmtFloat` function added by this PR.  I've never used `fmt.Sprint` like above, but I imagine it behaves like `"%v"` in a formatted string, which in some cases prints undesirable exponent notation.

rpc/mocks_test.go
```diff
@@ -0,0 +1,208 @@
...
+	mcrypto "github.com/athanorlabs/atomic-swap/crypto/monero"
+	"github.com/athanorlabs/atomic-swap/net"
+	"github.com/athanorlabs/atomic-swap/net/message"
+	"github.com/athanorlabs/atomic-swap/protocol/swap"
+	"github.com/athanorlabs/atomic-swap/protocol/txsender"
+)
+
+//
+// This file only contains mock definitions used by other test files
+//
```
dimalinux, line 23: These mocks were previously located in `ws_test.go`.  I didn't realize that they were also being shared with `net_test.go`, so I thought pulling them out into a separate file might make make it clear that they are shared. I put `panic`s in most of them, to make it more obvious to the reader which functions are actually getting invoked.  If we decide to instrument the mocks with testify's mocking at a later time (https://github.com/stretchr/testify#mock-package), we wouldn't need to instrument all functions.

rpc/mocks_test.go
```diff
@@ -0,0 +1,208 @@
...
+func (m *mockXMRMaker) GetOngoingSwapState(hash types.Hash) common.SwapState {
+	panic("not implemented")
+}
+
+func (*mockXMRMaker) MakeOffer(offer *types.Offer) (*types.OfferExtra, error) {
+	offerExtra := &types.OfferExtra{
+		StatusCh: make(chan types.Status, 1),
+		InfoFile: "/dev/null",
+	}
+	offerExtra.StatusCh <- types.CompletedSuccess
```
dimalinux, line 134: I uncommented the `TestSubscribeMakeOffer` test. I'm open to better ideas on how to transition the state for it.

rpc/server.go
```diff
@@ -43,67 +42,71 @@ type Config struct {
 
 // NewServer ...
 func NewServer(cfg *Config) (*Server, error) {
```
dimalinux, line 44: I pulled all the things that can fail in the lead-up to launching the server out of the `Start()` method and into this `NewServer` method. One of the things is allocating the listener, as that allows us to pass `0` for the port and have the OS assign us a free ephemeral port. This is useful when testing, when the actual port used is unimportant, but it must be a free port. We want to be able to query the allocated port back from the `Server` object before calling the blocking method `Start()`.

rpc/server.go
```diff
@@ -43,67 +42,71 @@ type Config struct {
...
+	r := mux.NewRouter()
+	r.Handle("/", rpcServer)
+	r.Handle("/ws", wsServer)
+
+	headersOk := handlers.AllowedHeaders([]string{"content-type", "username", "password"})
+	methodsOk := handlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "OPTIONS"})
+	originsOk := handlers.AllowedOrigins([]string{"*"})
+	server := &http.Server{
+		Addr:              ln.Addr().String(),
+		ReadHeaderTimeout: time.Second,
```
dimalinux, line 78: Interestingly, the linter flagged this as vulnerable to the Slowloris attack (https://www.cloudflare.com/learning/ddos/ddos-attack-tools/slowloris/) without the `ReadHeaderTimeout` value, but as best I can tell the original `ListenAndServe` call we were using would also be vulnerable, but doesn't get flagged by the linter. I guess because there is not API to set it.

net/discovery.go
```diff
@@ -37,6 +37,19 @@ func newDiscovery(ctx context.Context, h libp2phost.Host, bnsFunc func() []peer.
...
+	// so we added the extra guard below:
+	// Panic:  https://github.com/jbenet/goprocess/blob/v0.1.4/impl-mutex.go#L99
+	// Caller: https://github.com/libp2p/go-libp2p-kad-dht/blob/v0.17.0/dht.go#L222
+	//
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+		// not cancelled, continue on
+	}
```
dimalinux, line 51: This was mildly painful. We were introducing a context cancel in the `TestDaemon_DevXMR[MT]aker` tests in `cmd/daemon` to stop the http server and it was triggering a panic in this libp2p code. Apparently, this libp2p code has really slow startup times, as I even tried sleeping for a 2 whole seconds before issuing the context cancel and I was still getting panics.

ui/src/components/TakeDealDialog.svelte
```diff
@@ -14,7 +14,7 @@
   import HelperText from '@smui/textfield/helper-text'
   import { currentAccount, sign } from '../stores/metamask'
 
-  const WS_ADDRESS = 'ws://127.0.0.1:8081'
+  const WS_ADDRESS = 'ws://127.0.0.1:5001/ws'
```
dimalinux, line 17: I suspect this is the only change needed, but since the UI doesn't work for me, I can't test it.

cmd/client/main.go
```diff
@@ -239,17 +243,17 @@ var (
 						Usage:    "Duration of timeout, in seconds",
 						Required: true,
 					},
-					daemonAddrFlag,
+					swapdPortFlag,
 				},
 			},
 		},
-		Flags: []cli.Flag{daemonAddrFlag},
```
dimalinux, line 246: I wanted users to be able to specify `--swapd-port` before or after the subcommand.  The `spf13/cobra` library makes this easy to do, but I guess it is not supported by urfave: https://github.com/urfave/cli/issues/1391

When we define the same flag globally and after the command, the behavior is bad. I wasn't getting an error for the global flag (since it was listed as a global flag), but when reading it, we would get the defaulted value from the flag to the subcommand instead of the globally set value.
Example:
```
./swapcli --swapd-port 5432 address
```
The actual port used would be 5001, because the defaulted value for `--swapd-port` after the `address` subcommand was taking precedence over the 5432 value.

cmd/client/main.go
```diff
@@ -360,27 +364,41 @@ func runMake(ctx *cli.Context) error {
...
+		ethAsset = types.EthAsset(ethcommon.HexToAddress(ethAssetStr))
+	}
+
+	c := newRRPClient(ctx)
+	ourAddresses, err := c.Addresses()
+	if err != nil {
+		return err
+	}
+
+	printOffferSummary := func(offerID string) {
```
noot, line 382:
```suggestion
	printOfferSummary := func(offerID string) {
```

net/discovery.go
noot: good to know, also yeah that's painful

rpc/mocks_test.go
noot: it's a lot more complex to update the tests to not use mocking, as we basically need to start up everything in the daemon. it might be nice to add them as integration tests, instead of unit tests.

noot (commented):
looks good! could you also update the docs to reflect the changes (local.md, rpc.md, and stagenet.md)?

rpcclient/wsclient/wsclient.go
```diff
@@ -20,9 +20,8 @@ type WsClient interface {
 	Close()
 	Discover(provides types.ProvidesCoin, searchTime uint64) ([][]string, error)
 	Query(maddr string) (*rpctypes.QueryPeerResponse, error)
-	SubscribeSwapStatus(id uint64) (<-chan types.Status, error)
-	TakeOfferAndSubscribe(multiaddr, offerID string,
-		providesAmount float64) (id uint64, ch <-chan types.Status, err error)
+	SubscribeSwapStatus(id types.Hash) (<-chan types.Status, error)
+	TakeOfferAndSubscribe(multiaddr, offerID string, providesAmount float64) (ch <-chan types.Status, err error)
```
dimalinux, line 24: I didn't change these functions.  The interface signatures were just out of date with the existing implementation.

noot (approved):
looks good to me!

PR #189, opened 2022-09-11
match folders to exec names

dimalinux:
Changes:
* Matches executable names to folder names
* Moves the non-executable `cmd/utils` package to `cliutil`
* Simplifies build script
* Removes scripts for cross compilations (simpler versions can be added back when we support rust cross compilation)
* Adds initial, not yet perfect, version support for all our executables

Q: What is less than perfect about the current version string?

A: The way I handled auto-insertion of the release tag will not work until/unless we support `go install github.com/athanorlabs/atomic-swap/cmd/swapd@TAG` where TAG is either a release tag or a commit hash. Without building that way, the tag section will always contain the phrase "(devel)".  Example:
```
$ ./swapd --version
swapd version (devel) 9b11733-dirty-go1.18.5
```
Note: I had modified files in my repo, so "-dirty" was appended to the commit hash above.


scripts/build.sh
```diff
@@ -3,39 +3,8 @@
...
 cd "${PROJECT_ROOT}" || exit 1
 
-echo "building swapd..."
-cd cmd/daemon || exit 1
-if ! go build -o swapd; then
-	exit 1
+if [[ -n "${ALL}" ]]; then
+	GOBIN="${PROJECT_ROOT}" go install ./cmd/...
+else
+	GOBIN="${PROJECT_ROOT}" go install ./cmd/swapd ./cmd/swapcli
```
noot, line 9: nice!

noot (approved):
looks good!

PR #190, opened 2022-09-14
fixes for base58 encoding/decoding

dimalinux:
Changes:
* Completely removes the `paxosglobal/moneroutil` base58 codec, as it was buggy and might not have been license compatible.
* `github.com/btcsuite/btcd/btcutil/base58` now provides the base base58 codec and we layer monero specifics on top of it.
* Renames `EncodeMoneroBase58`/`DecodeMoneroBase58` to `MoneroAddrBytesToBase58`/`MoneroAddrBase58ToBytes` to emphasize that the codec is only for monero addresses. 
* Removes `github.com/ebfe/keccak"` dependency in favor of go-ethereum's Keccak256 (removing a dependency without adding a new one).


crypto/monero/address.go
```diff
@@ -11,26 +12,24 @@ import (
...
 )
 
 var (
-	errChecksumMismatch         = fmt.Errorf("invalid address checksum")
-	errInvalidAddressLength     = fmt.Errorf("invalid monero address length")
-	errInvalidPrefixGotMainnet  = fmt.Errorf("invalid monero address: expected stagenet, got mainnet")
-	errInvalidPrefixGotStagenet = fmt.Errorf("invalid monero address: expected mainnet, got stagenet")
+	errChecksumMismatch         = errors.New("invalid address checksum")
+	errInvalidAddressLength     = errors.New("invalid monero address length")
+	errInvalidAddressEncoding   = errors.New("invalid monero address encoding")
```
dimalinux, line 20: I was adding `errInvalidAddressEncoding` and didn't want to use `fmt.Errorf` as there was no string formatting, so I just updated them all to match.

crypto/monero/address.go
```diff
@@ -11,26 +12,24 @@ import (
...
 type Address string
 
 // ValidateAddress checks if the given address is valid
 func ValidateAddress(addr string, env common.Environment) error {
-	b := DecodeMoneroBase58(addr)
-	if len(b) != AddressLength {
-		return fmt.Errorf("%w: got %d, expected %d", errInvalidAddressLength, len(b), AddressLength)
+	b, err := MoneroAddrBase58ToBytes(addr)
+	if err != nil {
+		return err
```
dimalinux, line 32: The new function checks both the length and the encoding and returns an error.

crypto/monero/address_test.go
```diff
@@ -27,3 +28,15 @@ func TestValidateAddress(t *testing.T) {
 	err = ValidateAddress("fake", common.Mainnet)
 	require.True(t, errors.Is(err, errInvalidAddressLength))
 }
+
```
noot, line 31: it would be good to add some test vectors to this file that contain real monero addresses to ensure we can decode them properly

noot (approved):
looks good! subaddress decoding :D

Issue #120, opened 2022-05-24
add checks that monerod/monero-wallet-rpc are running on correct network


noot:
there is no check that monerod/monero-wallet-rpc are connected and running on the right network; this should be added otherwise swaps may fail without the issue being obvious.

PR #192, opened 2022-09-28
monero-wallet-rpc now embedded

dimalinux:
This PR embeds `monero-wallet-rpc` instead of having the user start it separately. As part of this work, the location of all the files we use is [now documented](https://github.com/AthanorLabs/atomic-swap/blob/dimalinux/embedded-monero-wallet-rpc/docs/default-file-locations.md) and code was updated to match the documented behavior.

Behavior changes: Both Ethereum and Monero wallets will be created if they do not exist.

There is a new `personal_balances` RPC API and has `swapcli` support via a new `balances` sub-command.


.gitignore
```diff
@@ -10,10 +10,10 @@
...
 
 # Generated binaries
-swapcli
-swapd
-swaprecover
-swaptester
+/swapcli
+/swapd
+/swaprecover
+/swaptester
```
dimalinux, line 16: We only want to ignore the executables in the root folder and not the `cmd/*` folders that have the same names.

cmd/swapd/main.go
```diff
@@ -209,32 +220,8 @@ type daemon struct {
...
-	_ = logging.SetLogLevel("common", level)
-	_ = logging.SetLogLevel("cmd", level)
-	_ = logging.SetLogLevel("net", level)
-	_ = logging.SetLogLevel("rpc", level)
-	return nil
-}
-
 func runDaemon(c *cli.Context) error {
-	if err := setLogLevels(c); err != nil {
+	if err := logging.SetLogLevel("*", c.String(flagLogLevel)); err != nil {
```
dimalinux, line 224: This may need a partial revert.  I started looking into it, because we were not getting logs from the `monero` package. The downside of this change is that were are getting more logs that we probably want from the libp2p libraries.

cmd/swaprecover/main.go
```diff
@@ -77,8 +105,9 @@ var (
 				Usage: "Ethereum chain ID; eg. mainnet=1, goerli=5, ganache=1337",
 			},
 			&cli.UintFlag{
-				Name:  flagGasPrice,
-				Usage: "Ethereum gas price to use for transactions (in gwei). If not set, the gas price is set via oracle.",
+				Name:   flagGasPrice,
+				Usage:  "Ethereum gas price to use for transactions (in gwei). If not set, the gas price is set via oracle.",
+				Hidden: true, // Not well tested, hiding from end users for now
```
dimalinux, line 110: @noot Just wanted to make sure that you saw that I hid the gas price flag. I didn't want to set end-user expectations that the flag is well tested and works.

tests/integration_test.go
```diff
@@ -27,160 +27,113 @@ const (
...
-	defaultXMRMakerDaemonEndpoint   = "http://localhost:5002"
-	defaultXMRMakerDaemonWSEndpoint = "ws://localhost:5002/ws"
-	defaultCharlieDaemonWSEndpoint  = "ws://localhost:5003/ws"
+	defaultXMRTakerSwapdEndpoint   = "http://localhost:5001"
+	defaultXMRTakerSwapdWSEndpoint = "ws://localhost:5001/ws"
+	defaultXMRMakerSwapdEndpoint   = "http://localhost:5002"
+	defaultXMRMakerSwapdWSEndpoint = "ws://localhost:5002/ws"
+	defaultCharlieSwapdWSEndpoint  = "ws://localhost:5003/ws"
+
+	defaultXMRMakerMoneroEndpoint = "http://127.0.0.1:18083/json_rpc"
```
dimalinux, line 36: The only change in the first block was replacing `Daemon` with `Swapd` to clarify which daemon was being referenced. 

The `defaultXMRMakerMoneroEndpoint` constant was moved from the `common` package. Normally, the end user would not specify a port that `monero-wallet-rpc` will provide RPC services on, but we provide the ability to set the port number. It was needed here in the integration tests, so that we could get Bob's (XMR Maker's) wallet address and mine some monero for him.

scripts/run-integration-tests.sh
```diff
@@ -3,78 +3,114 @@
...
+	done
+	if ! CONTRACT_ADDR="$(cat "${SWAP_TEST_DATA_DIR}/alice/contract-address.json")"; then
+		echo "Failed to get Alice's deployed contract address"
+		stop-daemons
+		exit 1
+	fi
+
+	start-swapd bob \
+		--dev-xmrmaker \
+		"--wallet-port=${BOB_WALLET_PORT}" \
```
dimalinux, line 75: Only Bob uses this `--wallet-port` flag. As noted in `integration_tests.go` below, we needed direct access to the wallet from the integration tests to generate funds for the XMR maker. The `--wallet-port` flag is marked as hidden to discourage people from using it unnecessarily.

scripts/build.sh
```diff
@@ -4,7 +4,7 @@ PROJECT_ROOT="$(dirname "$(dirname "$(readlink -f "$0")")")"
 cd "${PROJECT_ROOT}" || exit 1
 
 if [[ -n "${ALL}" ]]; then
-	GOBIN="${PROJECT_ROOT}" go install ./cmd/...
+	GOBIN="${PROJECT_ROOT}" go install -tags prod ./cmd/...
 else
-	GOBIN="${PROJECT_ROOT}" go install ./cmd/swapd ./cmd/swapcli
+	GOBIN="${PROJECT_ROOT}" go install -tags prod ./cmd/swapd ./cmd/swapcli
```
dimalinux, line 9: Test code from the `monero` package won't get compiled into the binary if you pass the `prod` tag. If you don't care, you can build the binaries without using the `prod` build tag.

cmd/swapd/main.go
```diff
@@ -284,7 +296,7 @@ func (d *daemon) make(c *cli.Context) error {
 
 	// By default, the chain ID is derived from the `flagEnv` value, but it can be overridden if
 	// `flagEthereumChainID` is passed:
-	if c.Uint(flagEthereumChainID) != 0 {
+	if c.IsSet(flagEthereumChainID) {
```
dimalinux, line 299: I'll remove `flagEthereumChainID` in a follow-up PR in favor of dynamic determination of the chain ID.

cmd/swapd/main.go
```diff
@@ -470,8 +457,9 @@ func newBackend(
 
 	contractAddrStr := c.String(flagContractAddress)
 	if contractAddrStr != "" {
-		// We check the contract code at the address later, so we don't need
-		// to tightly validate the address here.
```
dimalinux, line 474: I added that comment in a previous PR, as it was bothering me that were not error checking. The comment was incorrect.  On error we would get an address of all zeros and would not validate the contract code.

cmd/swapd/main.go
```diff
@@ -31,11 +33,6 @@ const (
 	defaultXMRTakerLibp2pPort = 9933
 	defaultXMRMakerLibp2pPort = 9934
 
-	// default libp2p key files
-	defaultLibp2pKey         = "node.key"
-	defaultXMRTakerLibp2pKey = "xmrtaker.key"
-	defaultXMRMakerLibp2pKey = "xmrmaker.key"
```
dimalinux, line 37: We now place the libp2p key in `{DATA_DIR}/net.key`.  The XMR taker and maker use separate data dirs eliminating the need for these.

cmd/swapd/main.go
```diff
@@ -230,6 +238,7 @@ func setLogLevels(c *cli.Context) error {
 	_ = logging.SetLogLevel("cmd", level)
 	_ = logging.SetLogLevel("net", level)
 	_ = logging.SetLogLevel("rpc", level)
+	_ = logging.SetLogLevel("monero", level)
```
dimalinux, line 241: I don't know what the solution is, but it would be nice if we came up with a solution that doesn't require updating every time we add or rename a package.  I tried (but reverted) a change that set the log level for `"*"`, but that had to downside of including lots of verbose logs from the libp2p libraries which we may not want.

cmd/swaptester/main.go
```diff
@@ -189,17 +196,23 @@ func getRandomExchangeRate() types.ExchangeRate {
 }
 
 func generateBlocks() {
-	cXMRMaker := monero.NewWalletClient(common.DefaultXMRTakerMoneroEndpoint)
-
+	cXMRMaker := monero.NewThinWalletClient(defaultXMRMakerMoneroEndpoint)
```
dimalinux, line 199: I changed line above to the XMR Maker, because I didn't understand why the XMR taker would be mining monero.

common/utils.go
```diff
@@ -38,10 +39,14 @@ func MakeDir(dir string) error {
...
-func Exists(path string) (bool, error) {
-	_, err := os.Stat(path)
+// FileExists returns whether the given file exists. If a directory exists
+// with the name of the passed file, an error is returned.
+func FileExists(path string) (bool, error) {
+	st, err := os.Stat(path)
 	if err == nil {
+		if st.IsDir() {
+			return false, fmt.Errorf("%q is occupied by a directory", path)
+		}
```
dimalinux, line 49: All existing callers were checking files and I wanted the directory error check, so I renamed the function to `FileExists`.

docs/default-file-locations.md
```diff
@@ -0,0 +1,59 @@
...
+It is safe to delete this directory if `swapd` is not running.
+
+### {DATA_DIR}/info-{DATE}.json
+
+Stores information on a swap when it reaches the stage where ethereum is locked.
+
+### {DATA_DIR}/contract-address.json
+
+Only written when `--deploy` is passed to swapd. This file stores the address
+that the contract was deployed to along with other data.
```
dimalinux, line 59: When I look at the JSON file, it looks like most of the fields in it should be removed. Should we create a ticket to clean that up?

docs/local.md
```diff
@@ -15,88 +15,72 @@ if you want a more sophisticated setup.
...
+BOOT_NODE=/ip4/127.0.0.1/tcp/9933/p2p/12D3KooWHRi24PVZ6TBnQJHdVyewDRcKFZtYV3qmB4KQo8iMyqik
 ```
-
-Then, mine some blocks on the monero test chain by running the following RPC command, replacing the address with the one from Bob's wallet:
+Now get the ethereum contract address that Alice deployed to. This can be pulled from the Alice's logs,
+the file ..., or if you have `jq` installed (available via `sudo apt install jq`), you can set a
+variable like this:
 ```bash
-curl -X POST http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"generateblocks","params":{"wallet_address":"45GcPCBQgCG3tYcYqLdj4iQixpDZYw1MGew4PH1rthp9X2YrB2c2dty1r7SwhbCXw1RJMvfy8cW1UXyeESTAuLkV5bTrZRe","amount_of_blocks":100}' -H 'Content-Type: application/json'
+CONTRACT_ADDR=$(jq -r .ContractAddress /tmp/xmrtaker/contract-address.json)
```
dimalinux, line 54: This could be documented with `grep` and `sed`, but it is a little uglier. I suspect most developers have `jq` installed.  I'm using `jq` in `scripts/run-integration-tests.sh` for this same purpose.

monero/utils_test.go
```diff
@@ -1,45 +1,36 @@
...
 	heightBefore, err := c.GetHeight()
 	require.NoError(t, err)
 
-	var wg sync.WaitGroup
-	wg.Add(1)
-	go func() {
-		errGen := daemon.GenerateBlocks(addr.Address, 181)
-		require.NoError(t, errGen)
-		wg.Done()
-	}()
```
dimalinux, line 31: Background mining is automatically provided by `CreateWalletClient(t)`, so we no longer need these code blocks.

rpcclient/discover.go
```diff
@@ -65,7 +65,7 @@ func (c *Client) QueryAll(provides types.ProvidesCoin, searchTime uint64) ([]*rp
 		return nil, resp.Error
 	}
 
-	var res *rpctypes.QueryAllResponse
+	var res rpctypes.QueryAllResponse
 	if err = json.Unmarshal(resp.Result, &res); err != nil {
```
dimalinux, line 69: I didn't understand how this code was working.  I thought the object passed to `Unmarshal` had to already be allocated, but maybe it didn't need to be fixed. If you pass a pointer to a pointer, will `json.Unmarshal` allocate the object?

scripts/cleanup-test-processes.sh
```diff
@@ -7,5 +7,16 @@ pkill --echo --uid "${UID}" --full '/ganache.* --deterministic '
...
 # that you don't want to kill, don't use this script!
+if pkill --echo --uid "${UID}" --full '/swapd '; then
+	# If swapd instances were killed, give the monero-wallet-rpc instances
+	# some time to shutdown.
+	sleep 4 # Give time for wallet to shutdown
+fi
+
+# Take note of monero-wallet-rpc instances being killed here if the instances
+# were started by swapd
 pkill --echo --uid "${UID}" --full '/monero-wallet-rpc '
```
dimalinux, line 18: At least on Linux, this has been working pretty good. After I added the short sleep after killing `swapd` instances, I haven't seen any `monero-wallet-rpc` instances that were both created by `swapd` but not shut down when `swapd` exited. Vetting the `monerod` node before starting `monero-wallet-rpc` was also critical to getting this working.  `monero-wallet-rpc` had incredibly slow shutodown times when it could not communicate with the monero node.

rpc/net.go
```diff
@@ -17,7 +17,7 @@ import (
 
 const defaultSearchTime = time.Second * 12
 
-// Net contains the functions required by the rpc service into the network.
+// Net contains the functions required by the rpc service into the network. TODO: "into the network"???
```
dimalinux, line 20: I didn't understand the comment.

net/host.go
```diff
@@ -81,8 +80,8 @@ type Config struct {
 
 // NewHost returns a new host
 func NewHost(cfg *Config) (*host, error) {
-	if cfg.KeyFile == "" {
-		cfg.KeyFile = defaultKeyFile
+	if cfg.DataDir == "" || cfg.KeyFile == "" {
+		panic("required parameters not set")
```
dimalinux, line 84: I made sure you can't force a panic by doing something silly like `--data-dir ""` in the options processing.

cliutil/utils.go
```diff
@@ -9,40 +9,62 @@ import (
...
+	var err error
+
+	switch {
+	case env == common.Development && devXMRMaker:
+		key, err = ethcrypto.HexToECDSA(common.DefaultPrivKeyXMRMaker)
+	case env == common.Development && devXMRTaker:
+		key, err = ethcrypto.HexToECDSA(common.DefaultPrivKeyXMRTaker)
+	default:
+		key, err = ethcrypto.GenerateKey()
+	}
```
noot, line 33: currently, both sides need to have a funded ETH key for the swap to work. once relayers are implemented, the xmr-maker won't need a funded ETH key, but the xmr-taker will still. we should probably add a log for this saying "fund this key"?

cmd/swapcli/main.go
```diff
@@ -289,6 +298,24 @@ func runAddresses(ctx *cli.Context) error {
 	return nil
 }
 
+func runBalances(ctx *cli.Context) error {
```
noot, line 301: nice addition!!

cmd/swapd/main.go
noot: hm, does `github.com/athanorlabs/atomic-swap/*` work as the package name? yeah, definitely don't want those libp2p logs

cmd/swaprecover/main.go
noot: good to know. for deploying on L2s we'll need to document this (or add hard coded values per chain) but this is fine for now

cmd/swapd/main.go
```diff
@@ -78,7 +76,7 @@ const (
 	flagDeploy       = "deploy"
 	flagTransferBack = "transfer-back"
 
-	flagLog = "log"
+	flagLogLevel = "log-level"
```
noot, line 79: nit but could you change this back to `log` unless you feel strongly about this?

cmd/swaptester/main.go
```diff
@@ -113,7 +118,9 @@ func runTester(c *cli.Context) error {
 
 	isDev = c.Bool(flagDev)
 	if !isDev {
-		defaultMoneroClient = monero.NewWalletClient(common.DefaultXMRMakerMoneroEndpoint)
+		// TODO: Why do this when dev flag is not given? Can the code work if it is given?
+		// For this to work, you'll need to pass --wallet-port 18083 to the XMR Maker swapd
+		defaultMoneroClient = monero.NewThinWalletClient(defaultXMRMakerMoneroEndpoint)
```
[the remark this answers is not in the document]
noot: the `defaultMoneroClient` is only used to wait for blocks on the network, whereas on a dev network we'd be forcing new blocks by mining. however I think we can get rid of this and replace it with a daemon call to get the block height: https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_last_block_header

feel free to do it in a follow up or open an issue!

cliutil/utils.go
dimalinux: I'll add a log

docs/default-file-locations.md
noot: what else is in the file? I think looking at the file I have, the other fields seem to now be in the infofile instead, so we can probably remove them (or just move contract address to the infofile)

docs/local.md
```diff
@@ -15,88 +15,72 @@ if you want a more sophisticated setup.
...
-make build
-```
+Note: when using the `--dev-xmrtaker` and `--dev-xmrmaker` flags, Alice's RPC server runs
+on http://localhost:5001 and Bob's runs on http://localhost:5002 by default.
 
-This creates `swapd` and `swapcli` binaries in the root directory.
+Now, in terminal 3, we will interact with the swap daemon using `swapcli`.
 
-To run as Alice, execute in terminal 1:
+First we need mine some monero for Bob. Alice already has Ethereum, because she is using
```
noot, line 67:
```suggestion
First we need mine some XMR for Bob. Alice already has ETH, because she is using
```
nit :p

docs/local.md
```diff
@@ -15,88 +15,72 @@ if you want a more sophisticated setup.
...
 
-Alice will print out a libp2p node address, for example `/ip4/127.0.0.1/tcp/9933/p2p/12D3KooWFUEQpGHQ3PtypLvgnWc5XjrqM2zyvdrZXin4vTpQ6QE5`. This will be used for Bob to connect.
-
-To run as Bob and connect to Alice, replace the bootnode in the following line with what Alice logged, and execute in terminal 2:
-
+To mine some monero blocks for Bob, you can use our bash shell function shown below:
 ```bash
-./swapd --dev-xmrmaker --wallet-file Bob --bootnodes /ip4/127.0.0.1/tcp/9933/p2p/12D3KooWFUEQpGHQ3PtypLvgnWc5XjrqM2zyvdrZXin4vTpQ6QE5
+source scripts/testlib.sh
+mine-monero-for-swapd 5002
```
noot, line 79: this is really nice!

docs/stagenet.md
```diff
@@ -23,15 +24,19 @@ For Linux 64-bit, you can do:
...
+   data. If you skip this step, a new wallet will be created that you can later fund for
+   swaps.
 
 4. Create a Goerli wallet. You can do this using Metamask by selecting "Goerli Test Network" from the networks, then creating a new account with "Create account". I'd recommend naming this new account something explicit like `goerli-swap-account`.
 
-5. Export the private key for this account by navigating to: three dots in upper right of Metamask -> account details -> export private key. Paste this private key into a file named `goerli.key`.
+5. Optional: Export the private key for this account by navigating to: three dots in upper right of
+   Metamask -> account details -> export private key. Paste this private key into a file
+   named `{DATA_DIR}/eth.key`. If you skip this step, a new goerli wallet will be created for you
+   that you can transfer stagenet ether to or fund directly in the next step.
```
noot, line 39:
```suggestion
   that you can transfer Goerli ether to or fund directly in the next step.
```

cmd/swapd/main.go
dimalinux: I think `log-level` will be less confusing for our end users. I, personally, was confused not knowing if the flag was for a log file, a level, a format type like JSON, or something else.

I've worked with a lot of server programs, and most of them that let you set the log level on the command line used the flag `--log-level` or a close variation of it.

Here are some examples using `--log-level LEVEL` verbatim on the command line:
* monerod: https://monerodocs.org/interacting/monerod-reference/
* vault: https://www.vaultproject.io/docs/commands/server
* dockerd: https://docs.docker.com/engine/reference/commandline/cli/
* docker-compose: https://devopstuto-docker.readthedocs.io/en/latest/compose/commands/commands.htm
* mysql-shell: https://dev.mysql.com/doc/mysql-shell/8.0/en/mysql-shell-application-log-introduction.html
* liquidbase: https://docs.liquibase.com/parameters/log-level.html
* gclusterd: https://access.redhat.com/documentation/en-us/red_hat_gluster_storage/3/html/administration_guide/configuring_the_log_level
* saltstack: https://docs.saltproject.io/en/latest/ref/configuration/logging/index.html

cmd/swapd/main.go
dimalinux: I'm not sure.  I'll create a ticket to look into it.

cmd/swaptester/main.go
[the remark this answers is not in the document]
dimalinux: I created a ticket/issue: https://github.com/AthanorLabs/atomic-swap/issues/199

We should be able to update this code the same way unit and regression tests were updated in this PR.  Enable background mining when in the dev environment so the rest of the code can be monero network independent.

docs/default-file-locations.md
dimalinux: https://pastebin.com/QyTt5med

docs/stagenet.md
```diff
@@ -121,24 +139,19 @@ If all goes well, you should see the node execute the swap protocol. If the swap
...
-
-1. Create a stagenet wallet:
-```bash
-curl http://localhost:18083/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"create_wallet","params":{"filename":"stagenet-wallet","password":"","language":"English"}}' -H 'Content-Type: application/json'
-```
  
-2. Find your stagenet address:
+1. Find your stagenet address:
 ```bash
 curl http://localhost:18083/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"get_accounts","params":{}}' -H 'Content-Type: application/json'
```
dimalinux, line 145: That line is accessing `monero-wallet-rpc` directly.  I'll update it to use `swapcli balances` before merging.

monero/wallet_client.go
```diff
@@ -172,3 +248,233 @@ func (c *walletClient) GetHeight() (uint64, error) {
...
+
+	pRead, pWrite, err := os.Pipe()
+	if err != nil {
+		return nil, err
+	}
+	cmd.Stdout = pWrite
+	cmd.Stderr = pWrite
+
+	// Last entry wins if an environment variable is in the list multiple times.
+	// We parse some output, so we want to force English. NO_COLO[U]R=1 failed to
```
dimalinux, line 376: This comment line was killing me.  The commonly used environment variable is only `NO_COLOR`, not `NO_COLOUR`: https://no-color.org/

Unfortunately, there is no way to add a linter exception, at least not that I could figure out, in a line that is already a comment, so I had to pretend that the variable might have multiple spellings.  :)

cliutil/utils.go
dimalinux: Added a log, although I missed the word "fund". If you think we need it, feel free to suggest how to work it in.
```
New ETH wallet key generated in /home/dmitry/.atomicswap/alice-data-stagenet/eth.key
 with public address 0xD594c55bbBfdaC0b62c3C4C0f65533faaaeAA411
```

monero/wallet_client.go
```diff
@@ -172,3 +248,233 @@ func (c *walletClient) GetHeight() (uint64, error) {
...
+			return fmt.Errorf("monerod endpoint %s is not a mainnet node", endpoint)
+		}
+	case common.Development:
+		if info.NetType != "fakechain" {
+			return fmt.Errorf("monerod endpoint %s should have a network type of \"fakechain\" in dev mode",
+				endpoint)
+		}
+	default:
+		panic("unhandled environment type")
+	}
```
noot, line 294: nice, this should close #120 also

monero/wallet_client.go
```diff
@@ -172,3 +248,233 @@ func (c *walletClient) GetHeight() (uint64, error) {
...
+func launchMoneroWalletRPCChild(walletRPCBin string, walletRPCBinArgs ...string) (*os.Process, error) {
+	cmd := exec.Command(walletRPCBin, walletRPCBinArgs...)
+
+	pRead, pWrite, err := os.Pipe()
+	if err != nil {
+		return nil, err
+	}
+	cmd.Stdout = pWrite
+	cmd.Stderr = pWrite
+
```
noot, line 374:
```suggestion
//nolint:misspell
```

monero/wallet_client.go
noot: you can add a `//nolint:misspell` to the top of the whole comment!

protocol/xmrtaker/swap_state.go
```diff
@@ -646,29 +646,25 @@ func (s *swapState) claimMonero(skB *mcrypto.PrivateSpendKey) (mcrypto.Address,
...
-			_ = s.GenerateBlocks(string(depositAddr), 64)
-			_ = s.Refresh()
-		}
-
 		balance, err := s.GetBalance(0)
 		if err != nil {
 			return fmt.Errorf("failed to get balance: %w", err)
 		}
 
+		// TODO: Should this function be waiting for some minimum unlocked balance instead?
```
noot, line 661: probably not, the code here is only used when transferring from the newly-generated swap account (which is a fresh account) back into the user's desired account to receive the swap funds. since it's waiting on the total swap funds then it should wait for the whole balance to unlock.

the only issue that could arise is if the user decides for whatever weird reason to send funds into the swap account, which would end up making this function take longer

rpc/net.go
noot:
```suggestion
// Net contains the network-related functions required by the rpc service.
```

maybe something like this?

rpcclient/discover.go
noot: it should work either way, it worked before so I think both are ok

cliutil/utils.go
noot: I'd add either a bold log after this saying `this address must be funded to take an offer` or something like that, or I'd add a log before the balance check in `xmrtaker/net.go` (where `errBalanceTooLow` is returned) printing out the address again (or returning it in the error) saying "this account needs funds" or something along those lines

docs/default-file-locations.md
noot: this actually appears to be a complete duplicate of the info file if i'm not mistaken? `protocol.write.go` `InfoFileContents` contains what's in the info file which has the contract address, plus all those other fields :/

noot: ah, so in `cmd/main/contract.go` it writes `contract-address.json` to disk if it deploys it. i think we can probably just remove that?

noot (approved):
looks good! only minor suggestions/comments. nice work on this large (and very useful) update :)

monero/wallet_client.go
dimalinux: Thanks!  That worked.

monero/wallet_client.go
dimalinux: Linked issue to PR

protocol/xmrmaker/swap_state.go
```diff
@@ -553,23 +553,12 @@ func (s *swapState) lockFunds(amount common.MoneroAmount) (mcrypto.Address, erro
 
 	log.Infof("locked XMR, txHash=%s fee=%d", txResp.TxHash, txResp.Fee)
 
-	xmrmakerAddr, err := s.GetAddress(0)
+	// wait for a new block
+	height, err := monero.WaitForBlocks(s, 1)
```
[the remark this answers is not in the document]
dimalinux: Yes.  The first call to `CreateWalletClient(t)` in a unit test starts a background miner.

cliutil/utils.go
dimalinux: Additional log added in both places.

PR #196, opened 2022-10-03
remove hardhat dependency from ethereum/ dir

### LINKED ISSUE #191: remove JS dependency from ethereum/ dir
remove JS (hardhat) from the ethereum/ dir and have it only contain contracts and relevant go code/bindings. potentially move the `swapfactory/*` go bindings to `ethereum/`, as there are now more contracts than just `SwapFactory.sol`.

noot:
also move contract bindings to `ethereum/`.

closes #191


cmd/swaprecover/main.go
```diff
@@ -15,12 +15,12 @@ import (
 	"github.com/athanorlabs/atomic-swap/cliutil"
 	"github.com/athanorlabs/atomic-swap/common"
 	mcrypto "github.com/athanorlabs/atomic-swap/crypto/monero"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
```
dimalinux, line 18: It would decrease code complexity if we didn't rename our own packages when importing. Is there a place where the generated code from contracts could be placed in a folder named `contracts`?

dimalinux (approved):
Looks good, but if we can figure out a way to avoid renaming our own import everywhere, it will be a win for code simplicity.

Issue #83, opened 2022-01-23
persist past swap info to disk


noot:
currently past swap information is only stored in memory, it would be nice to persist this to disk. 
- will require a basic database implementation, a key-value store should be fine for this

PR #200, opened 2022-10-08
implement database; persist offers to disk and reload on startup

### LINKED ISSUE #118: save and reload offers to/from disk
currently, offers aren't persisted to disk, so when a node restarts it has no offers. offers should be persisted to disk so when the node restarts, it reloads all its offers again.

### LINKED ISSUE #172: Offer taken off offers list on invalid incoming take
A taker who sends an invalid take of an offer (for example, provided amount too low) will still cause the offer to be removed from the xmr maker's offer list, despite the take offer not being valid. Other peers will not be able to try to take the same offer since it is gone.

To reproduce:
- Create maker and taker1 and taker2
- maker creates an offer
- taker1 takes the offer but with provided amount too low
- taker2 tries to take the same offer, but it no longer exists

noot:
- implement a database which stores offers
- reload those offers on startup
- advertise in the network if there are offers found on startup
closes #118

also fixes #172 - in `protocol/xmrmaker/net.go` the offer was being "taken" (ie. deleted) and not re-added back if there was an error found before `newSwapState` was called.


cmd/swapd/main.go
```diff
@@ -352,14 +358,24 @@ func (d *daemon) make(c *cli.Context) error {
 		return err
 	}
 
+	dbCfg := &chaindb.Config{
+		DataDir: path.Join(cfg.DataDir, "db"),
```
dimalinux, line 362: Before we merge this PR, can you add this to the `default-file-locations.md` document? Include information about the database, like whether it is for internal use by our software only, or if the user might pull information from it, and when the end-use can safely delete it.

cmd/swapd/main.go
```diff
@@ -213,8 +215,9 @@ type xmrmakerHandler interface {
 }
 
 type daemon struct {
-	ctx    context.Context
-	cancel context.CancelFunc
+	ctx      context.Context
+	cancel   context.CancelFunc
+	database *db.Database
```
dimalinux, line 220: We are not in a package named `swap` or `offer`. I think it would help the reader understand the code if we are more specific in our naming like calling the type `swapDaemon` or `swapd` and the database `offerDatabase` or `offerDB`.

cmd/swapd/main.go
noot: yes, for sure!

cmd/swapd/main.go
noot: I plan to put other things in the database other than just offers (see #83) which is why I just named it `database`. I can rename `daemon` though

cmd/swapd/main_test.go
```diff
@@ -140,3 +145,79 @@ func Test_expandBootnodes(t *testing.T) {
 	}
 	require.EqualValues(t, expected, expandBootnodes(cliNodes))
 }
+
+func TestDaemon_PersistOffers(t *testing.T) {
+	const defaultXMRMakerSwapdEndpoint = "http://localhost:5002"
```
dimalinux, line 150: format string?

db/database.go
```diff
@@ -0,0 +1,113 @@
...
+	offerTable := chaindb.NewTable(db, offerPrefix)
+
+	return &Database{
+		offerTable: offerTable,
+	}, nil
+}
+
+// Close flushes and closes the database.
+func (db *Database) Close() error {
+	err := db.offerTable.Flush()
```
dimalinux, line 36: Do we need need this if we are calling close?

protocol/xmrmaker/mocks_generate_test.go
```diff
@@ -2,3 +2,4 @@ package xmrmaker
 
 //nolint:lll
 //go:generate mockgen -destination=mocks_test.go -package $GOPACKAGE github.com/athanorlabs/atomic-swap/protocol/backend Backend
```
dimalinux, line 4: rename?

dimalinux (approved):

PR #201, opened 2022-10-09
removed --ethereum-chain-id flag

dimalinux:
The `--ethereum-chain-id` flag was somewhat redundant, because the chain ID can be obtained from the value passed by the user to `--ethereum-endpoint`. This PR updates the code to dynamically determines the chain ID using the endpoint.


cmd/swapd/main.go
```diff
@@ -338,11 +327,20 @@ func (d *daemon) make(c *cli.Context) error {
...
 
+	ethEndpoint := common.DefaultEthEndpoint
+	if c.String(flagEthereumEndpoint) != "" {
+		ethEndpoint = c.String(flagEthereumEndpoint)
+	}
+	ec, err := ethclient.Dial(ethEndpoint)
+	if err != nil {
+		return err
+	}
+
```
dimalinux, line 338: Moved the code to initialize the ethclient from `newBackend` to here.

cmd/swapd/main.go
```diff
@@ -338,11 +327,24 @@ func (d *daemon) make(c *cli.Context) error {
...
+		ethEndpoint = c.String(flagEthereumEndpoint)
+	}
+	ec, err := ethclient.Dial(ethEndpoint)
+	if err != nil {
+		return err
+	}
+	chainID, err := ec.ChainID(d.ctx)
+	if err != nil {
+		return err
+	}
```
dimalinux, line 341: Moved the above code above from `newBackend`  so we would have access to the Chain ID sooner.

cmd/swapd/main.go
```diff
@@ -353,11 +355,12 @@ func (d *daemon) make(c *cli.Context) error {
 	}
 
 	sm := swap.NewManager()
-	backend, err := newBackend(d.ctx, c, env, cfg, devXMRMaker, devXMRTaker, sm, host)
+	backend, err := newBackend(d.ctx, c, env, cfg, devXMRMaker, devXMRTaker, sm, host, ec)
 	if err != nil {
 		return err
 	}
 	defer backend.Close()
+	log.Infof("created backend with monero endpoint %s and ethereum endpoint %s", backend.Endpoint(), ethEndpoint)
```
dimalinux, line 363: It might make the code easier to understand if we change the backend to use a `has-a` instead of an `is-a` relationship with the monero wallet at some point.

cmd/swapd/main.go
```diff
@@ -538,11 +528,6 @@ func newBackend(
 		return nil, fmt.Errorf("failed to make backend: %w", err)
 	}
 
-	log.Infof("created backend with monero endpoint %s and ethereum endpoint %s",
-		mc.Endpoint(),
-		ethEndpoint,
-	)
```
dimalinux, line 544: Moved to caller, we still have the log.

common/consts.go
```diff
@@ -1,11 +1,7 @@
 package common
 
 const (
-	MainnetChainID = 1 //nolint
-	GorliChainID   = 5
-	GanacheChainID = 1337
-
-	DefaultMoneroDaemonMainnetPort  = 18081
```
dimalinux, line 8: Apparently, the `nolint` was getting applied to all the elements in the `const` block, not just the elements before the blank line.

noot (approved):
looks good!

PR #204, opened 2022-10-11
add relayer support for `claim` contract function

### LINKED ISSUE #33: implement relayer/metatx support for swap contract `claim` function
currently, Bob needs to have some ETH to pay for gas when calling `Claim()` to get his ETH. this is a pretty old ETH problem (eg. I have DAI but no ETH, so the DAI is trapped) - I'm sure there is likely some existing solution to this, just need to explore what's out there.

### LINKED ISSUE #157: consistent function naming
Function names should use mixedCase according to the Solidity style guide: https://docs.soliditylang.org/en/v0.8.16/style-guide.html#function-names

Currently, we have new_swap, set_ready, claim, refund, and verifySecret. So I propose new_swap => `newSwap` (or maybe make it an actionable verb phrase like `createSwap` or `initSwap`), set_ready => `setReady`.

The same goes for local state variables. So perhaps `timeout_0` => `timeout0` and `timeout_1` => `timeout1`. But these are less important.

### LINKED ISSUE #208: panic on database shutdown
By accident I typed `swapd` instead of `swapcli` and discovered this panic. It's unrelated to what I'm working on now, so just ticketing it. Based on the output, it looks related to the database close.
```
$ ./swapd balance --swapd-port 5001
2022-10-22T17:17:55.443-0500	ERROR	cmd	swapd/main.go:270	failed to make daemon: dial tcp 127.0.0.1:8545: connect: connection refused
protocol complete, shutting down...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xfa6081]

goroutine 1 [running]:
github.com/athanorlabs/atomic-swap/db.(*Database).Close(...)
	/home/dima/dev/atomic-swap/db/database.go:36
main.(*daemon).stop(0xc000a27e80)
	/home/dima/dev/atomic-swap/cmd/swapd/main.go:286 +0x21
main.runDaemon(0xc000a27e00)
	/home/dima/dev/atomic-swap/cmd/swapd/main.go:277 +0x15e
github.com/urfave/cli/v2.(*App).RunContext(0x1f29d00, {0x1733c80?, 0xc0000400d0}, {0xc00003c080, 0x4, 0x4})
	/home/dima/go/pkg/mod/github.com/urfave/cli/v2@v2.10.2/app.go:350 +0x9d3
github.com/urfave/cli/v2.(*App).Run(...)
	/home/dima/go/pkg/mod/github.com/urfave/cli/v2@v2.10.2/app.go:247
main.main()
	/home/dima/dev/atomic-swap/cmd/swapd/main.go:201 +0x47
```

noot:
Initial relayer support integration. Allows a relayer endpoint and commission (in percentage) to be specified when making an offer (as the maker is always the one to claim ETH). When `claim()` is called, the xmrmaker submits a signed transaction to a relayer, which submits the tx (paying gas), which calls `claimRelayer()`, which transfers the funds (minus the commission amount) to the xmrmaker. This allows the xmrmaker to not have a funded Ethereum account to swap :D

Additionally fixes the "local" 2-node swap, since `generate_blocks` was removed, the node now starts background mining if `--dev-xmrmaker`. This is a bit janky and would probably be better as a `--background-mine` flag compatible with `--env=dev`, let me know if you want me to separate that into another PR.

closes #33
also closes #157
also closes #208


cmd/swapcli/main.go
```diff
@@ -422,7 +445,14 @@ func runMake(ctx *cli.Context) error {
...
 		defer wsc.Close()
 
-		id, statusCh, err := wsc.MakeOfferAndSubscribe(min, max, types.ExchangeRate(exchangeRate), ethAsset)
+		id, statusCh, err := wsc.MakeOfferAndSubscribe(
+			min,
+			max,
+			types.ExchangeRate(exchangeRate),
+			ethAsset,
+			relayerEndpoint,
+			relayerCommission,
```
dimalinux, line 454: With 6 fields, we may want to create an input structure and make it the sole parameter to `MakeOfferAndSubscribe`.  Then you get labels on the values and, if valid, some values can be omitted. Just a thought, not required.

cmd/swapd/contract.go
```diff
@@ -8,19 +8,20 @@ import (
 	"fmt"
 	"path"
 
+	gsnforwarder "github.com/AthanorLabs/go-relayer/examples/gsn_forwarder"
```
dimalinux, line 11: Is this our code? It seems weird to be importing from a package that includes a folder named `examples` and then a package name with an underscore.  Underscores can be used, but it's considered non-idiomatic: https://go.dev/doc/effective_go#package-names

cmd/swapd/contract.go
```diff
@@ -98,6 +86,82 @@ func getSwapFactory(client *ethclient.Client, addr ethcommon.Address) (*contract
...
+	if err != nil {
+		return ethcommon.Address{}, fmt.Errorf("failed to register domain separator: %w", err)
+	}
+
+	_, err = block.WaitForReceipt(ctx, ec, tx.Hash())
+	if err != nil {
+		return ethcommon.Address{}, err
+	}
+
+	return address, nil
```
dimalinux, line 166: Would it make sense to move some of the above 2 functions to the ethereum package?  Maybe not the part that is writing the contract address to disk, but the rest of it seems like it might be a good fit for that package.

cmd/swapd/main.go
```diff
@@ -180,6 +181,10 @@ var (
 				Name:  flagDeploy,
 				Usage: "Deploy an instance of the swap contract",
 			},
+			&cli.StringFlag{
+				Name:  flagForwarderAddress,
+				Usage: "If --deploy is set, provide the address of the trusted forwarder contract for the deployed swap contract",
+			},
```
dimalinux, line 187: Just throwing out an idea. It's probably not even a good idea. If you wanted to simplify the inter-flag dependencies, you could call this flag `--deploy-with-forwarder ADDRESS`.  Although, I guess that would still leave the question if it is an error if `--deploy` and `--deploy-with-forwarder` are passed on the same command line.

If you leave it as-is, maybe change the description to something like:
> Specifies the Ethereum address of the trusted forwarder contract when deploying the swap contract. Ignored if --deploy is not passed.

cmd/swapd/main.go
```diff
@@ -411,6 +420,7 @@ func (d *daemon) make(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	// this closes the monero.WalletClient
```
dimalinux, line 423: Yeah, converting the backend to use a has-a relationship for the monero and ethereum clients will make a lot of code more obvious. I thought about doing it in my current work, but I think it needs to be done in a PR dedicated to that change alone.

scripts/run-integration-tests.sh
```diff
@@ -56,7 +56,7 @@ start-daemons() {
 	# Wait up to 10 seconds for Alice's swapd instance to start and deploy the swap contract
 	#
 	CONTRACT_ADDR_FILE="${SWAP_TEST_DATA_DIR}/alice/contract-address.json"
-	for _ in {1..10}; do
+	for _ in {1..60}; do
```
dimalinux, line 59: Hopefully, it's not taking a full minute. You can remove the phrase "up to 10 seconds" from the comment.

tests/relayer_integration_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func setupRelayer(t *testing.T) {
+	relayerSk := GetTestKeyByIndex(t, 1)
+	ec, chainID := NewEthClient(t)
+
+	swapContractAddrStr := os.Getenv(contractAddrEnv)
+	if swapContractAddrStr == "" {
+		panic("CONTRACT_ADDR env var not set")
+	}
```
dimalinux, line 50: You could replace these 3 lines with:
```
require.NotEmptyf(t, swapContractAddrStr, "CONTRACT_ADDR environment variable not set")
```

cmd/swapd/contract.go
noot: good point, updated!

tests/relayer_integration_test.go
noot: updated!

scripts/run-integration-tests.sh
noot: updated, I don't think it is but I'm not sure exactly how long it takes to start up so I just put this to be safe

cmd/swapd/main.go
noot: description updated! left the flags as-is for now

cmd/swapcli/errors.go
```diff
@@ -5,9 +5,13 @@ import (
...
 var (
-	errNoMinAmount      = errors.New("must provide non-zero --min-amount")
-	errNoMaxAmount      = errors.New("must provide non-zero --max-amount")
-	errNoExchangeRate   = errors.New("must provide non-zero --exchange-rate")
-	errNoProvidesAmount = errors.New("must provide non-zero --provides-amount")
-	errNoDuration       = errors.New("must provide non-zero --duration")
+	errNoMinAmount                        = errors.New("must provide non-zero --min-amount")
+	errNoMaxAmount                        = errors.New("must provide non-zero --max-amount")
+	errNoExchangeRate                     = errors.New("must provide non-zero --exchange-rate")
+	errNoProvidesAmount                   = errors.New("must provide non-zero --provides-amount")
```
dimalinux, line 11: You could use the flag name constants when creating these strings.  Then it will be easier to update the flag names in the future. There probably isn't a way to fix this, but if someone came up with a good, yet concise, way to make the coin type and numerator/denominator types obvious from these flag names, it would be a definite use experience win.

protocol/check.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"bytes"
+	"context"
+	"errors"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
```
dimalinux, line 11: Can you document:
* Why we need this constant. Why the similar value in the generated swap factory code can not be used.
* How to update this constant the next time the contract code changes.
* How the offsets in `CheckContractCode` were derived and how they work.

cmd/swapd/main_test.go
```diff
@@ -148,7 +148,8 @@ func Test_expandBootnodes(t *testing.T) {
 
 func TestDaemon_PersistOffers(t *testing.T) {
 	defaultXMRMakerSwapdEndpoint := fmt.Sprintf("http://localhost:%d", defaultXMRMakerRPCPort)
-	startupTimeout := time.Second * 9
+	// TODO: figure out a way to tell if the node is fully started, like startedCh
+	startupTimeout := time.Second * 45
```
dimalinux, line 152: Maybe it should be a defect (and test failure) if our server fails to start in 20 seconds even on a slow VM? Have we looked into or need a ticket to look into what is causing the startup delays?

monero/mine_regtest.go
```diff
@@ -0,0 +1,64 @@
...
+				// blocks, not an error that matters unless it is happening frequently.
+				continue
+			} else if err != nil {
+				log.Warnf("failed to mine block: %s", err)
+			}
+
+			log.Debugf("background mined 1 monero block")
+		}
+	}()
+}
```
dimalinux, line 64: I assume mining on the monero node works even when using `--regtest --offline`.  For the use case that this code is for, would passing these values to `monerod` be an option?
* `--start-mining="${REWARDS_ADDRESS}"`
* `--mining-threads=1`

protocol/check.go
dimalinux: I did some research and I guess the problem is [immutable variables](https://docs.sourcify.dev/docs/immutables/).

protocol/check.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return err
+	}
+
+	expectedCode := ethcommon.FromHex(expectedBytecode)
+
+	// sadly this is hard-coded for now since the forwarder address becomes part of the
+	// compiled bytecode.
+	if !bytes.Equal(expectedCode[:1473], code[:1473]) || !bytes.Equal(expectedCode[1543:], code[1543:]) {
+		return errInvalidSwapContract
```
dimalinux, line 29: 1543 - 1473 = 70 hex bytes. 35 binary bytes for the forwarder address?

Maybe compare `len(code)` with `len(expectedCode)` first so we know that those indices on `code` will never trigger a panic?

protocol/check_test.go
```diff
@@ -1,42 +1,31 @@
...
-
 	ec, chainID := tests.NewEthClient(t)
 	ctx := context.Background()
 	pk := tests.GetMakerTestKey(t)
 
 	txOpts, err := bind.NewKeyedTransactorWithChainID(pk, chainID)
 	require.NoError(t, err)
 
-	_, tx, _, err := contracts.DeploySwapFactory(txOpts, ec)
+	_, tx, _, err := contracts.DeploySwapFactory(txOpts, ec, ethcommon.Address{})
```
dimalinux, line 23: Can we deploy it with a couple different addresses and verify that the gap that we don't check matches the expected addresses?

protocol/check.go
noot: updated, let me know if the comments make sense

cmd/swapd/main_test.go
noot: I updated this test to wait on a `startedCh` which is closed when the daemon starts up, seems to work better now.

protocol/check.go
noot: the comment I added should explain what's in those bytes

protocol/check_test.go
noot: yes, added

monero/mine_regtest.go
noot: I'd need to modify the `start-monero-regtest` function in `testlib.sh` for this, which might mess up other things.. can do in a follow up?

cmd/swapcli/errors.go
noot: updated!

dimalinux (approved):

protocol/check.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"bytes"
+	"context"
+	"errors"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// expectedBytecode is generated by deploying an instance of SwapFactory.sol and calling `CodeAt`
```
dimalinux, line 12: I was having trouble replicating the `expectedByteCode`, because you didn't mention what address it was deployed with.  Parsing it out, it looks like it was deployed with `0x79d5a6f7d950e2b616f638f670bdb124bb9b6b67`.

Issue #203, opened 2022-10-11
store info file contents in database


noot:
the info file contains information needed to recover swap funds in case of a swapd crash (after funds are locked but before they are claimed). currently, it's just stored in a json file that's written to disk. however, this info can be stored in the database instead now that we have one, and cleared out upon swap completion.

this will also allow for `swapcli` functionality to be built that allows users to easily query for completed vs. ongoing swaps.

PR #207, opened 2022-10-20
monero transaction checks

dimalinux:
* Bob now waits for his XMR transfer to the A+B wallet to have 10 confirmations before notifying Alice that the XMR is locked.
* Alice now rejects the locked Monero if there is more than one block left before she can spend the funds (1 block freedom is in case of network latency on Alice's side). This eliminates Bob's ability to double spend and allows Alice to sweep funds to her primary wallet from the A+B wallet in a reasonable time frame.
* Sweeping received XMR to the primary wallet is now the default behavior.
* A+B swap wallets are created with a restore height set for quick wallet creation in production where the blockchain is big.
* Fixes SleepWithContext moving it into the common package and having it return an error when the sleep is interrupted by the context being canceled.


common/utils.go
```diff
@@ -56,3 +58,18 @@ func FileExists(path string) (bool, error) {
...
+	select {
+	case <-ctx.Done():
+		if !timer.Stop() {
+			<-timer.C
+		}
+		return ctx.Err()
+	case <-timer.C:
+		return nil
+	}
+}
```
dimalinux, line 75: This was moved from `ethereum/block` to make it available to other packages. This version now returns an error type so the caller knows if the sleep was interrupted.

ethereum/block/wait_for_timestamp.go
```diff
@@ -6,27 +6,19 @@ import (
...
 // WaitForEthBlockAfterTimestamp returns the header of the first block whose timestamp is >= ts.
 func WaitForEthBlockAfterTimestamp(ctx context.Context, ec *ethclient.Client, ts int64) (*ethtypes.Header, error) {
 	timeDelta := time.Duration(ts-time.Now().Unix()) * time.Second
 
 	// The sleep is safe even if timeDelta is negative. We only optimise for timestamps in the future, but if
 	// the timestamp had already passed for some reason, nothing bad happens.
-	sleepWithContext(ctx, timeDelta)
+	if err := common.SleepWithContext(ctx, timeDelta); err != nil {
+		return nil, err
+	}
```
dimalinux, line 21: We don't want to keep on executing the function if the sleep was interrupted by the context being cancelled.

monero/test_support.go
```diff
@@ -85,14 +85,15 @@ func BackgroundMineBlocks(t *testing.T) {
 		cancelFunc()
 		wg.Wait()
 	})
+	// Lower the sleep duration used by WaitForBlock
+	blockSleepDuration = backgroundMineInterval / 3
```
dimalinux, line 89: I realized that our 10 second sleep when we are checking for a new monero block was too long when running unit tests where we are now mining a little less than 2 blocks/sec. I changed `blockSleepDuration` from a constant to a variable, but the only place that overrides it is here in this code that doesn't get compiled into our production binaries.

common/utils.go
```diff
@@ -56,3 +58,18 @@ func FileExists(path string) (bool, error) {
 
 	return false, err
 }
+
+// SleepWithContext is the same as time.Sleep(...) but with preemption if the context is
+// complete. Returns nil if the sleep completed, otherwise the context's error.
+func SleepWithContext(ctx context.Context, d time.Duration) error {
```
dimalinux, line 64: This was moved from ethereum/block to make it available to other packages. This version now returns an error type so the caller knows if the sleep was interrupted.

scripts/testlib.sh
```diff
@@ -121,6 +121,10 @@ start-monerod-regtest() {
 		--rpc-bind-ip=127.0.0.1 \
 		--rpc-bind-port=18081
 	sleep 5
+	# Make sure the blockchain has some initial decoy outputs. Arbitrarily sending the
+	# rewards to the Mastering Monero address.
+	local rewardsAddr=4BKjy1uVRTPiz4pHyaXXawb82XpzLiowSDd8rEQJGqvN6AD6kWosLQ6VJXW9sghopxXgQSh1RTd54JdvvCRsXiF41xvfeW5
+	mine-monero "${rewardsAddr}" > /dev/null
```
dimalinux, line 127: When the first test after running `setup-env.sh` had a monero transfer in it, I was getting errors from the transfer about insufficient outputs.  This seemed to fix it.

common/utils_test.go
```diff
@@ -35,3 +39,48 @@ func TestMakeDir(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "drwx------", fileStats.Mode().String()) // only user has access
 }
+
+func TestFileExists(t *testing.T) {
```
dimalinux, line 43: Just increasing the coverage in `utils.go`.  I had modified this function in a recent PR, but it had no coverage.

scripts/testlib.sh
```diff
@@ -121,6 +121,10 @@ start-monerod-regtest() {
 		--rpc-bind-ip=127.0.0.1 \
 		--rpc-bind-port=18081
 	sleep 5
+	# Make sure the blockchain has some initial decoy outputs. Arbitrarily sending the
+	# rewards to the Mastering Monero address.
+	local rewardsAddr=4BKjy1uVRTPiz4pHyaXXawb82XpzLiowSDd8rEQJGqvN6AD6kWosLQ6VJXW9sghopxXgQSh1RTd54JdvvCRsXiF41xvfeW5
+	mine-monero "${rewardsAddr}" >/dev/null
```
dimalinux, line 127: When the first test after running `setup-env.sh` had a monero transfer in it, I was getting errors from the transfer about insufficient outputs. Mining some initial unspent outputs seemed to fix it.

tests/monero_wallet_rpc.go
```diff
@@ -1,93 +0,0 @@
-package tests
```
dimalinux, line 1: This was obsolete code I should have deleted in a previous PR.

recover/recovery.go
```diff
@@ -85,7 +85,7 @@ func (r *recoverer) WalletFromSharedSecret(pk *mcrypto.PrivateKeyInfo) (mcrypto.
 	}
 
 	kp := mcrypto.NewPrivateKeyPair(sk, vk)
-	return monero.CreateWallet("recovered-wallet", r.env, r.xmrClient, kp)
+	return monero.CreateWallet("recovered-wallet", r.env, r.xmrClient, kp, 0)
```
dimalinux, line 88: We should revisit this when going over the recoverer. If we record the block height before the swap started in the recovery information, we could speed recovery wallet synchronization.

protocol/xmrtaker/swap_state_test.go
```diff
@@ -345,9 +345,19 @@ func TestSwapState_NotifyClaimed(t *testing.T) {
...
+		DestAddr:         xmrAddr,
+		NumConfirmations: monero.MinSpendConfirmations,
+	})
+	require.NoError(t, err)
+	t.Logf("Transfer mined at block=%d with %d confirmations", transfer.Height, transfer.Confirmations)
+
 	// send notification that monero was locked
 	lmsg := &message.NotifyXMRLock{
 		Address: string(xmrAddr),
+		TxID:    transfer.TxID,
```
dimalinux, line 360: Bob isn't currently using the TxID at the current time, but having the value might help with troubleshooting if nothing else.

protocol/xmrtaker/recovery.go
```diff
@@ -47,18 +47,19 @@ func NewRecoveryState(b backend.Backend, dataDir string, secret *mcrypto.Private
...
-		claimedCh:      make(chan struct{}),
-		info:           pswap.NewEmptyInfo(),
+		ctx:               ctx,
+		cancel:            cancel,
+		Backend:           b,
+		sender:            sender,
+		privkeys:          kp,
+		pubkeys:           pubkp,
+		dleqProof:         dleq.NewProofWithSecret(sc),
+		moneroBlockHeight: 0, // TODO: Can we optimise this?
```
dimalinux, line 57: This was the only line that changed.  If we get the block height before the swap into the recovery information, we could speed recovery wallet synchronization.

protocol/xmrtaker/message_handler.go
```diff
@@ -239,68 +236,39 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) (net.Message
...
-		log.Infof("monero block height: %d", height)
-	}
-
-	log.Debug("refreshing client...")
-
 	if err := s.Refresh(); err != nil {
 		return nil, fmt.Errorf("failed to refresh client: %w", err)
 	}
 
-	accounts, err := s.GetAccounts()
```
dimalinux, line 266: We created the A+B swap wallet and know the funds are in the first and only account index, so I simplified this code. I tried to find a way to use `generate_from_keys` to create a non-zero index in an existing wallet, but it appears to not be possible.

protocol/xmrtaker/instance_test.go
```diff
@@ -3,17 +3,22 @@ package xmrtaker
...
 	"github.com/stretchr/testify/require"
 
-	"github.com/athanorlabs/atomic-swap/monero"
+	"github.com/athanorlabs/atomic-swap/common/types"
 )
 
-func TestGetAddress(t *testing.T) {
-	c := monero.CreateWalletClient(t)
-	addr, err := getAddress(c, "", "")
+func TestNewInstance(t *testing.T) {
```
dimalinux, line 12: I removed the `getAddress` function and `TestGetAddress` was the only unit test for `instance.go`. To keep the file and get a little coverage, I added `TestNewInstance`.

protocol/xmrtaker/instance.go
```diff
@@ -41,70 +32,28 @@ type Instance struct {
...
-	)
-
 	// if this is set, it transfers all xmr received during swaps back to the given wallet.
 	if cfg.TransferBack {
-		address, err = getAddress(cfg.Backend, cfg.MoneroWalletFile, cfg.MoneroWalletPassword)
-		if err != nil {
-			return nil, err
-		}
-		cfg.Backend.SetBaseXMRDepositAddress(address)
+		cfg.Backend.SetBaseXMRDepositAddress(cfg.Backend.PrimaryWalletAddress())
```
dimalinux, line 47: The `getAddress` function was removed in favor of asking the wallet what its primary address is. We cache the address when starting the wallet.

monero/wallet_client.go
```diff
@@ -351,9 +443,9 @@ func createWalletRPCService(conf *WalletClientConf) (string, *os.Process, error)
 func getMoneroWalletRPCBin() (string, error) {
 	execName := "monero-wallet-rpc"
 	priorityPath := path.Join("monero-bin", execName)
-	path, err := exec.LookPath(priorityPath)
+	execPath, err := exec.LookPath(priorityPath)
```
dimalinux, line 446: Variable name conflicted with import name. My IDE warns about it, but I don't think our linter does. If we find the knob to enabled that check, I think it would be good.

cmd/swapd/main.go
```diff
@@ -675,11 +676,9 @@ func getProtocolInstances(c *cli.Context, cfg common.Config,
 	walletPassword := c.String(flagMoneroWalletPassword)
 
 	xmrtakerCfg := &xmrtaker.Config{
-		Backend:              b,
-		DataDir:              cfg.DataDir,
-		MoneroWalletFile:     walletFilePath,
-		MoneroWalletPassword: walletPassword,
```
noot, line 681: why were these removed?

monero/wallet_client.go
```diff
@@ -30,11 +35,25 @@ type WalletClient interface {
 	UnlockClient()
 	GetAccounts() (*wallet.GetAccountsResponse, error)
 	GetAddress(idx uint64) (*wallet.GetAddressResponse, error)
+	PrimaryWalletAddress() mcrypto.Address
 	GetBalance(idx uint64) (*wallet.GetBalanceResponse, error)
 	Transfer(to mcrypto.Address, accountIdx, amount uint64) (*wallet.TransferResponse, error)
 	SweepAll(to mcrypto.Address, accountIdx uint64) (*wallet.SweepAllResponse, error)
-	GenerateFromKeys(kp *mcrypto.PrivateKeyPair, filename, password string, env common.Environment) error
-	GenerateViewOnlyWalletFromKeys(vk *mcrypto.PrivateViewKey, address mcrypto.Address, filename, password string) error
+	WaitForTransReceipt(req *WaitForReceiptRequest) (*wallet.Transfer, error)
```
noot, line 42:
```suggestion
	WaitForTxReceipt(req *WaitForReceiptRequest) (*wallet.Transfer, error)
```

or just
```suggestion
	WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Transfer, error)
```

sorry for the nit, just want to name things consistenty since we use `tx` a lot

noot: I'm also unsure about use of the word receipt since that's ethereum specific, maybe it could just be `WaitForTransfer` or something like that?

monero/wallet_client.go
```diff
@@ -135,14 +172,53 @@ func (c *walletClient) GetBalance(idx uint64) (*wallet.GetBalanceResponse, error
...
+	if err != nil {
+		return nil, err
+	}
+
+	var transfer *wallet.Transfer
+
+	for {
+		transferResp, err := c.rpc.GetTransferByTxid(&wallet.GetTransferByTxidRequest{
+			TxID:         req.TxID,
+			AccountIndex: 0,
```
noot, line 191: will the account index always be zero?

monero/wallet_client.go
```diff
@@ -135,14 +172,53 @@ func (c *walletClient) GetBalance(idx uint64) (*wallet.GetBalanceResponse, error
...
+		})
+		if err != nil {
+			return nil, err
+		}
+		transfer = &transferResp.Transfer
+		// wait for transaction be mined (height set) even if 0 confirmations requested
+		if transfer.Height > 0 && transfer.Confirmations >= req.NumConfirmations {
+			break
+		}
+		log.Infof("Received %d of %d confirmations of XMR TXID=%s (height=%d)",
```
noot, line 201: move the log above the if statement (so it logs the exit case?)

net/message/message.go
```diff
@@ -227,7 +227,8 @@ func (m *NotifyETHLocked) Type() Type {
 
 // NotifyXMRLock is sent by XMRMaker to XMRTaker after locking his XMR.
 type NotifyXMRLock struct {
-	Address string
+	Address string // address the monero was sent to
```
noot, line 230: note to self: I think this can *probably* be removed since the address can be derived from the two parties's public keys.

cmd/swapd/main.go
dimalinux: Yes, they were no longer needed here, as we now cache the primary address and information to re-open the primary wallet in the `walletClient` struct.

monero/wallet_client.go
dimalinux: It's not really a receipt in either blockchain. I originally implemented your second suggestion as I liked the orthogonality, but it causes a name conflict with the ethereum method of the same name. I'll implement the change in the next week when I convert the wallet interfaces to use a has-a relationship in the backend. Then we can have orthogonality without naming conflicts.

monero/wallet_client.go
dimalinux: I'll fix it. Currently, it will always be zero, but the `Transfer` method that is called before it is passing the index as a parameter, so I'll update this function to do the same.

monero/wallet_client.go
dimalinux: I'll do that.

protocol/xmrtaker/recovery.go
noot: can add this in #203

monero/wallet_client.go
noot: sounds good!

noot (approved):
looks good, nice work!

noot (commented):
looks like there's some CI failures, will re-check after those are fixed

monero/utils.go
```diff
@@ -10,44 +11,47 @@ import (
...
+			return 0, err
+		}
 
-			height, err := client.GetHeight()
-			if err != nil {
-				continue
-			}
+		height, err := client.GetHeight()
+		if err != nil {
+			return 0, err
```
dimalinux, line 39: It didn't make sense for `Refresh` to terminate the function on error and then have `GetHeight` behave differently, so I just simplified and converted 2 loops into 1.

tests/integration_test.go
```diff
@@ -83,6 +82,15 @@ func mineMinXMRMakerBalance(t *testing.T, minBalance common.MoneroAmount) {
...
 }
 
+func (s *IntegrationTestSuite) newSwapdWSClient(ctx context.Context, endpoint string) wsclient.WsClient {
+	wsc, err := wsclient.NewWsClient(ctx, endpoint)
+	require.NoError(s.T(), err)
+	s.T().Cleanup(func() {
+		wsc.Close()
+	})
+	return wsc
+}
```
dimalinux, line 92: We were not closing the websocket connections when the tests ended, so I added this and converted all the tests to use it.

PR #213, opened 2022-10-27
persist swap info to disk

### LINKED ISSUE #83: persist past swap info to disk

noot:
- implement `swapTable` for database which puts/gets swaps from the db
- update `protocol/swap.manager` to accept the db, load swaps from disk on start-up and cache any ongoing ones
- when the swap managers gets or updates swaps, it also writes them to disk
- it also fetches past swaps from disk

closes #83


db/database.go
```diff
@@ -25,9 +28,11 @@ func NewDatabase(cfg *chaindb.Config) (*Database, error) {
 	}
 
 	offerTable := chaindb.NewTable(db, offerPrefix)
+	swapTable := chaindb.NewTable(db, swapPrefix)
```
dimalinux, line 31: The `Close()` function further down didn't change, so github won't let me comment there, but I don't see `swapTable` getting closed in it. You might move these 2 lines into the `Database` declaration, since the intermediate variables don't look like they are improving code readability.

db/database.go
```diff
@@ -4,17 +4,20 @@ import (
...
 	"github.com/ChainSafe/chaindb"
 )
 
-var (
+const (
 	offerPrefix = "offer"
+	swapPrefix  = "swap"
 )
 
 // Database is the persistent datastore used by swapd.
```
dimalinux, line 17: Can we document what the key/value pairs are for the 2 databases here?

db/database.go
```diff
@@ -87,10 +100,6 @@ func (db *Database) ClearAllOffers() error {
 	for iter.Valid() {
 		// key is the offer ID
 		key := iter.Key()
```
dimalinux, line 102: You could rename the variable to `offerID` and then the comment wouldn't be needed.

db/database.go
```diff
@@ -100,3 +109,65 @@ func (db *Database) ClearAllOffers() error {
...
 }
+
+// PutSwap puts the given swap in the database.
+func (db *Database) PutSwap(s *swap.Info) error {
+	val, err := json.Marshal(s)
+	if err != nil {
+		return err
+	}
+
+	key := s.ID
```
dimalinux, line 120: I always assumed these ID values indexing swaps were different from offer IDs, but it looks like they are the same. We could avoid a lot of confusion if we called this `ID` field `OfferID` as well.

So we are mapping the ID value in binary (verses hex) and it maps to a JSON (text based) object. We may need some kind of versioning in the `swap.Info` struct at some point.

db/database.go
```diff
@@ -100,3 +109,65 @@ func (db *Database) ClearAllOffers() error {
...
+func (db *Database) GetAllSwaps() ([]*swap.Info, error) {
+	iter := db.swapTable.NewIterator()
+	defer iter.Release()
+
+	swaps := []*swap.Info{}
+	for iter.Valid() {
+		key := iter.Key()
+
+		// if the key becomes longer than 32, we're not iterating over swaps
+		if len(key) > 32 {
```
dimalinux, line 155: You could put a constant like this at the top to avoid hard-coding it in multiple places:
```
const offerIDLength = ethcommon.HashLength
```

db/database.go
```diff
@@ -100,3 +109,65 @@ func (db *Database) ClearAllOffers() error {
...
+
+	return &s, nil
+}
+
+// GetAllSwaps returns all swaps in the database.
+func (db *Database) GetAllSwaps() ([]*swap.Info, error) {
+	iter := db.swapTable.NewIterator()
+	defer iter.Release()
+
+	swaps := []*swap.Info{}
```
dimalinux, line 150: The linter in my IDE complains when we do declarations like this.  It recommends:
```
var offers []*types.Offer
```
`nil` plays nice with `append` and wouldn't break anything in this function (or most good go code), but if you are worried that someone calling this function might not be treating an empty slice and `nil` the same way, the linter also likes this:
```
offers := make([]*types.Offer, 0)
```

dimalinux, line 150: The linter in my IDE complains when we do declarations like this.  It recommends:
```
var swaps []*swap.Info
```
`nil` plays nice with `append` and wouldn't break anything in this function (or most good go code), but if you are worried that someone calling this function might not be treating an empty slice and `nil` the same way, the linter also likes this:
```
swaps := make([]*swap.Info, 0)
```

db/database.go
```diff
@@ -100,3 +109,65 @@ func (db *Database) ClearAllOffers() error {
 
 	return nil
 }
+
+// PutSwap puts the given swap in the database.
```
dimalinux, line 113: Can we document and maybe even unit test what happens if `s.ID` is already in the database? Does it generate an error or does it overwrite the existing entry?

db/database.go
dimalinux: Reading down further in the PR, we'd also want to mention when `swapTable` entries are updated.

db/database_test.go
```diff
@@ -0,0 +1,99 @@
...
+		types.EthAssetETH,
+	)
+	err = db.PutOffer(offerA)
+	require.NoError(t, err)
+
+	infoA := &swap.Info{
+		ID: types.Hash{0x1},
+	}
+	err = db.PutSwap(infoA)
+	require.NoError(t, err)
```
dimalinux, line 84: Can we add an update example?

db/database.go
```diff
@@ -64,8 +69,16 @@ func (db *Database) GetAllOffers() ([]*types.Offer, error) {
 
 	offers := []*types.Offer{}
 	for iter.Valid() {
+		key := iter.Key()
+
+		// if the key becomes longer than 32, we're not iterating over offers
+		if len(key) > 32 {
+			break
+		}
```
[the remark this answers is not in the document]
noot: the underlying iterator is a bit messed up, it returns everything in the db starting at keys starting with `offer` and only slices off the prefix for those. I'll work on fixing that/making a workaround cause I agree this is cursed

and yeah, I think it could be reduced to 16 bytes, that leaves 2^128 different offer IDs which is more than enough

db/database.go
noot: > I always assumed these ID values indexing swaps were different from offer IDs, but it looks like they are the same. We could avoid a lot of confusion if we called this ID field OfferID as well.
yeah the offer ID just becomes the swap ID. sure, we can rename one to the other? or just call them all `ID`?

> So we are mapping the ID value in binary (verses hex)
what do you mean? the ID types for both types.Offer and swap.Info are types.Hash

db/database_test.go
noot: done

db/database.go
noot: done

db/database.go
noot: updated

db/database.go
noot: done!

db/database.go
noot: comment added, let me know what you think

db/database.go
dimalinux: I wrote that when I was first reviewing the PR and didn't know how the code worked yet. One might expect that since the values are encoded in human readable text (JSON), that the ID's might also be text (hex). At the time, I was just clarifying that it was a binary->text mapping. (Not saying that we need to change it.)

db/database.go
```diff
@@ -85,18 +112,76 @@ func (db *Database) ClearAllOffers() error {
...
+		value := iter.Value()
+
+		var s swap.Info
+		err := json.Unmarshal(value, &s)
 		if err != nil {
-			return err
+			return nil, err
 		}
+
+		swaps = append(swaps, &s)
```
dimalinux, line 182: Are there any other sanity checks that we should do besides the unmarshal? Besides data corruption, there is also the potential issue of the data we storing changing with new releases.

protocol/swap/manager.go
```diff
@@ -1,168 +1,156 @@
...
 	sync.RWMutex
 	ongoing map[types.Hash]*Info
 	past    map[types.Hash]*Info
 }
 
-// NewManager ...
-func NewManager() Manager {
+// NewManager returns a new Manager that uses the given database.
+// It loads all ongoing swaps into memory on construction.
+// Completed swaps are not loaded into memory.
```
dimalinux, line 32: So `ongoing` is fully populated and `past` is just a cache of recently queried values for speed?  We way want to comment that on the struct.

protocol/swap/manager.go
```diff
@@ -1,168 +1,156 @@
...
 	defer m.RUnlock()
-	ids := make([]types.Hash, len(m.past))
-	i := 0
+	ids := make(map[types.Hash]struct{})
 	for id := range m.past {
-		ids[i] = id
+		ids[id] = struct{}{}
+	}
+
+	// TODO: do we want to cache all past swaps since we're already fetching them?
```
dimalinux, line 80: We might need a max results parameter and a way to sort the past swaps from most recent to least recent. Even if they all fit in memory, the user may want to limit the number retrieved.

protocol/swap/manager.go
```diff
@@ -1,168 +1,156 @@
...
+func (m *manager) GetPastSwap(id types.Hash) (*Info, error) {
 	m.RLock()
 	defer m.RUnlock()
-	return m.past[id]
+	s, has := m.past[id]
+	if has {
+		return s, nil
+	}
+
+	// TODO: do we want to cache this swap?
```
dimalinux, line 113: I would say yes.  If you are going to cache any, a most recently used algorithm would be the way to go. If the database is fast, we may not need a cache at all. We are only serving a single user, so simplicity beats scalability.

protocol/swap/manager.go
```diff
@@ -1,168 +1,156 @@
...
 
 // CompleteOngoingSwap marks the current ongoing swap as completed.
-func (m *manager) CompleteOngoingSwap(id types.Hash) {
+func (m *manager) CompleteOngoingSwap(id types.Hash) error {
 	m.Lock()
 	defer m.Unlock()
 	s, has := m.ongoing[id]
 	if !has {
-		return
+		return nil
```
dimalinux, line 135: Why is this a non-error, non-logging event. Do we sometimes call this function twice on the same ID?  Maybe we can add a comment for the answer?

protocol/swap/manager.go
```diff
@@ -1,168 +1,156 @@
...
 	}
 
 	m.past[id] = s
 	delete(m.ongoing, id)
+
+	// re-write to db, as status has changed
```
dimalinux, line 141: I'm trying to think of how to make this more clear, but no great ideas are coming to my mind. It's a little confusing that this function is "marking" the "ongoing swap" as "completed" and updating the DB, but the status must be getting changed somewhere else before we are called.

db/database.go
```diff
@@ -85,18 +112,76 @@ func (db *Database) ClearAllOffers() error {
...
+	key := s.ID
+	return db.swapTable.Put(key[:], val)
+}
+
+// HasSwap returns whether the db contains a swap with the given ID.
+func (db *Database) HasSwap(id types.Hash) (bool, error) {
+	return db.swapTable.Has(id[:])
+}
+
+// GetSwap returns a swap with the given ID, if it exists.
```
dimalinux, line 143: Can we document if it returns `nil` or an error if it doesn't exist? Maybe add a unit test for the expected behavior too. (I looked at the tests to find the answer, but it wasn't in there.)

protocol/swap/manager.go
```diff
@@ -1,168 +1,156 @@
...
+	return m.db.PutSwap(s)
+}
+
+func (m *manager) getSwapFromDB(id types.Hash) (*Info, error) {
+	has, err := m.db.HasSwap(id)
+	if err != nil {
+		return nil, err
+	}
```
dimalinux, line 149: Can't you just call `GetSwap` first and skip the `HasSwap` check?

protocol/swap/types.go
```diff
@@ -0,0 +1,60 @@
...
+type Info struct {
+	ID             types.Hash // swap offer ID
+	Provides       types.ProvidesCoin
+	ProvidedAmount float64
+	ReceivedAmount float64
+	ExchangeRate   types.ExchangeRate
+	EthAsset       types.EthAsset
+	Status         Status
+	statusCh       <-chan types.Status `json:"-"`
+}
```
dimalinux, line 21: We need some custom marshaling on this data structure like we did for the Offer type so that, for example, the ID and EthAsset show up in hex instead of [0,1,...31] syntax. I can help out with that if you need it.

protocol/swap/types.go
```diff
@@ -0,0 +1,60 @@
...
+// NewInfo creates a new *Info from the given parameters.
+// Note that the swap ID is the same as the offer ID.
+func NewInfo(
+	id types.Hash,
+	provides types.ProvidesCoin,
+	providedAmount, receivedAmount float64,
+	exchangeRate types.ExchangeRate,
+	ethAsset types.EthAsset,
+	status Status,
+	statusCh <-chan types.Status,
```
dimalinux, line 32: Do we want this function? I'm not saying we don't, but I'll play devil's advocate here.  If you initialize the `Info` struct directly you get field labeling on all 7 fields. You could change the function `NewInfo` to take `*Info` as a parameter and get labeling, but that would kind of defeat the purpose.

protocol/swap/types.go
```diff
@@ -0,0 +1,60 @@
...
+		Status:         status,
+		statusCh:       statusCh,
+	}
+	return info
+}
+
+// NewEmptyInfo returns an empty *Info
+func NewEmptyInfo() *Info {
+	return &Info{}
+}
```
dimalinux, line 50: Any chance the caller of this function could use `nil` instead? `nil` makes things more clear when troubleshooting.

db/database.go
noot: hm, like if the `swap.Info` type changes? I guess in that case, the unmarshal should fail (probably). do you think adding a version byte would be good?

protocol/swap/manager.go
noot: I'll make an issue for this! might need to update `swap.Info` to have a `completed time.Time` or something

protocol/swap/manager.go
noot: good question, this should probably return an error since it shouldn't be called twice. will update

protocol/swap/manager.go
noot: right, the main solutions I can think of are separating ongoing and completed swaps within the db into two separate "tables", or writing the status to a key separate than the `swap.Info` is written to

db/database.go
noot: it errors if it doesn't, will add a test

protocol/swap/manager.go
noot: yeah, i just need to wrap the database error for if the key isn't found, will update

db/database.go
dimalinux: I was thinking we might need a semantic version as the first field.  If the major version changed, we'd have to use a new struct, because it was a backwards incompatible change.  We could only support one previous major version to keep things simple.  We would throw an error if we discovered a version newer than the latest version that the current `swapd` supports. Minor versions could be increased if you are just adding a field, but the older version would still be compatible. (That's off the top of my head, it may not be perfect.)

dimalinux: Not saying that all needs to be done this pull request, but we will want it before we get too many users.

protocol/swap/types.go
noot: updated, adding custom marshalling for `Hash` and `EthAsset`. was able to remove the marshalling for `Offer`, since those were the only two types that were custom marshalled

protocol/swap/types.go
noot: I prefer having constructors for types where all fields are required, otherwise it's unclear what fields need to be populated and how imo. also IDEs should show all argument names, so I don't necessarily see the need for labelling

protocol/swap/types.go
noot: it's only used by `NewRecoveryState`. I think this will require a bit of a refactor of that, can address in a follow up

protocol/swap/manager.go
noot: added

dimalinux (approved):

PR #214, opened 2022-10-27
updated how we check the contract

dimalinux:
Changes:
* Moved the swap factory contract checking code to the `ethereum` folder. That package seemed like a better fit given that we already had utility functions in it that were not generated code.
* We track the indexes of the trusted forwarder address in the compiled byte code in exact 20 byte (address length) increments.
* `CheckSwapFactoryContractCode` verifies that the same trusted forwarder address exists at all expected locations and returns the parsed value from the deployed contract.
* Unit tests were added that explicitly help the developer update the expected contract code as well as the slice of trusted forwarder indexes.

Not changed in this PR:

We still need a solution where gets Bob a forwarder that he is willing to use. Having the contract checking code return the parsed forwarder may help us add the feature, depending on what solution we go with.


noot (approved):
looks good!

PR #219, opened 2022-11-03
implement event watcher for contract events; remove unneeded network messages

### LINKED ISSUE #19: ensure event subscriptions work as intended
currently, the event subscriptions are a bit janky, we're using the go-ethereum abigen bindings which allow for watching for events, but it seems they don't always fire at the right time. investigate this, maybe it's something we're doing, or it might be a limitation with go-ethereum as their docs seem to indicate this isn't fully implemented.

the current workaround is to poll for events, and to send network messages when some contract function has been called.

noot:
as title says, `NotifyReady`, `NotifyClaimed` and `NotifyRefund` network message types were removed, as these can all be determined by watching the contract events instead.

- implement `ethereum/watcher` package which is a simple ethereum filter loop that watches for events
- refactor protocol/xmrmaker and protocol/xmrtaker to be event-driven via `Event` types and a handler loop as opposed to having state changes occur when the network package called `HandleProtocolMessage
- swap `Event`s can now be triggered via network message, contract event, or other (eg swap cancel, remote peer disconnecting)

closes #19


ethereum/watcher/watcher.go
```diff
@@ -0,0 +1,100 @@
...
+
+	eth "github.com/ethereum/go-ethereum"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+var checkForBlocksTimeout = time.Second
+
+// EventFilterer filters the chain for specific events (logs).
```
dimalinux, line 16: Is this a typo with the intended name being `EventFilter` (without the extra `er`)?

noot: not really, I guess I named it this since there was a `FilterLogs` function which this calls, making it a `Filterer` in a way, lol. I think either one is fine, can change if this bothers you?

dimalinux: I think it is just such an awkward word to say, that people avoid it. It's like calling someone that answers questions for a living an answerer. Whatever you decide is fine. I just wanted to make sure it wasn't a typo. :)

https://github.com/search?q=eventFilter&type=code (1 million results)
https://github.com/search?q=eventFilterer&type=code (28 results)

ethereum/watcher/watcher.go
```diff
@@ -0,0 +1,100 @@
...
+				continue
+			}
+
+			// let's see if we have logs
+			logs, err := f.ec.FilterLogs(f.ctx, f.filterQuery)
+			if err != nil {
+				continue
+			}
+
+			found := []ethtypes.Log{}
```
dimalinux, line 80: `found` is usually a boolean variable.  You could do this rename and make Goland happy at the same time:
```
var matches []ethtypes.Log
```

ethereum/watcher/watcher.go
```diff
@@ -0,0 +1,100 @@
...
+			for _, l := range logs {
+				if l.Topics[0] != f.topic {
+					continue
+				}
+
+				found = append(found, l)
+			}
+
+			if len(found) != 0 {
+				f.logCh <- found
```
dimalinux, line 90: Why do we push slices on the channel instead of individual logs? Are the entries in the slices duplicates?  If yes, do we need more than the first entry? (Just trying to understand the code better. Not asking for any change.)

protocol/xmrmaker/swap_state.go
```diff
@@ -136,31 +142,68 @@ func newSwapState(
...
+	)
+
+	refundedWatcher := watcher.NewEventFilterer(
+		ctx,
+		b.EthClient(),
+		b.ContractAddr(),
+		ethHeader.Number,
+		refundedTopic,
+		logRefundedCh,
+	)
```
dimalinux, line 172: How do these [removed log events](https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/types/log.go#L53-L55) and the [`Removed` field](https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/events.go#L30-L31) in the actual logs work? Do we need to be checking that field somewhere?

ethereum/watcher/watcher.go
```diff
@@ -0,0 +1,100 @@
...
+	return &EventFilterer{
+		ctx:         ctx,
+		ec:          ec,
+		topic:       topic,
+		filterQuery: filterQuery,
+		logCh:       logCh,
+	}
+}
+
+// Start starts the EventFilterer. It watches the chain for logs.
```
dimalinux, line 49: Can we add some comments somewhere about why we went with the implementation below?

The ethereum client has `SubscribeNewHead(...)` and `SubscribeFilterLogs(...)` which someone reading the code might expect, or might try to rewrite the code to use. I'm going to guess that polling is more reliable?

protocol/xmrmaker/swap_state.go
```diff
@@ -136,31 +142,68 @@ func newSwapState(
...
-	return s, nil
-}
+	// set up ethereum event watchers
+	logReadyCh := make(chan []ethtypes.Log)
+	logRefundedCh := make(chan []ethtypes.Log)
 
-func (s *swapState) lockState() {
-	s.stateMu.Lock()
-}
+	ctx, cancel := context.WithCancel(b.Ctx())
```
dimalinux, line 154: It took me a while to track this down.  I wanted to understand how the event filterer go routines were terminating.  It looks like this context inherits from the global context, so it will be canceled if the global context is canceled, but this subcontext only lasts for the length of the swap, so when the swap ends, this context is canceled and the swap specific filtering go routines are terminated. Let me know if that is correct. If you have any ideas on how to make this more obvious when reading the code, please add whatever you think might help.

protocol/xmrtaker/event.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (s *swapState) handleEvent(event Event) {
+	// events are only used once, so their error channel can be closed after handling.
+	switch e := event.(type) {
+	case *EventKeysReceived:
+		log.Infof("EventKeysReceived")
+		defer close(e.errCh)
+
+		if reflect.TypeOf(s.nextExpectedEvent) != reflect.TypeOf(&EventKeysReceived{}) {
```
dimalinux, line 116: I assume we are using reflection because it is awkward to check if `s.nextExpectedEvent` is nil and `.(type)` will panic on nil?  Maybe in the future we could have some sort of `None` event and not allow nil? Formatted prints, like the one on the next line, would also be improved.

protocol/xmrmaker/event.go
```diff
@@ -0,0 +1,198 @@
...
+			return
+		}
+
+		err := s.handleEventETHLocked(e)
+		if err != nil {
+			e.errCh <- fmt.Errorf("failed to handle EventETHLocked: %w", err)
+			return
+		}
+
+		s.setNextExpectedEvent(&EventContractReady{})
```
dimalinux, line 109: It is confusing that the `errCh` inside `EventContractReady` is not getting initialized. If all we need is the type for next expected event, maybe we should be storing an enumerated value or the type itself instead of an uninitialized object of the type?

protocol/xmrmaker/swap_state.go
noot: hm, yeah seems that should be checked in case of re-org, will update

ethereum/watcher/watcher.go
noot: yeah, I found geth's subscriptions really wonky and didn't really work, I can retry with `SubscribeFilterLogs` and see.

protocol/xmrmaker/swap_state.go
noot: yeah that's correct, a context is done when it cancels or one of its ancestor contexes is cancelled. not really sure other than comments, what do you think?

protocol/xmrtaker/event.go
noot: yeah I was going to change the `Event` interface to have a `Type()` function, would make these checks nicer. will add

ethereum/watcher/watcher.go
dimalinux: Given that we are creating one filter per swap, it would probably make sense to consider the offerID part of the filter and handle it here.

protocol/xmrmaker/event.go
```diff
@@ -0,0 +1,198 @@
...
+		return types.KeysExchanged
+	case *EventContractReady:
+		return types.XMRLocked
+	default:
+		return types.UnknownStatus
+	}
+}
+
+// Event represents a swap state event.
+type Event interface{}
```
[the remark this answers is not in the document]
noot: updated to have a `Type()` method

protocol/xmrmaker/event.go
noot: updated to have a `Type()` method

ethereum/watcher/watcher.go
noot: updated to push individual logs

protocol/xmrmaker/swap_state.go
noot: added comments, also updated the two deferred functions to match more closely.

ethereum/watcher/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+
+				if l.Removed {
+					continue
+				}
+
+				f.logCh <- l
+			}
+
+			// the filter inclusive of the latest block when `ToBlock` is nil, so we add 1
+			f.filterQuery.FromBlock = big.NewInt(0).Add(currHeader.Number, big.NewInt(1))
```
dimalinux, line 93: In the comment, you probably wanted to write `The filter includes the ...`.  Can we rename `header` to `prevHeader`? Don't we want to be adding +1 to the previous header and not the current header to ensure that we never skip a block? Possible scenarios I'm thinking of are:
* Pushing onto the log channel blocks for an extended period
* 2 blocks are created while we are sleeping
* A new block is created after we get the header, but before we execute the filter

ethereum/watcher/watcher.go
```diff
@@ -0,0 +1,99 @@
...
+				continue
+			}
+
+			for _, l := range logs {
+				if l.Topics[0] != f.topic {
+					continue
+				}
+
+				if l.Removed {
+					continue
```
dimalinux, line 86: We probably want to log this.  People are definitely going to be attacking the system by sending out simultaneous transactions that take more than one block to reach consensus on. A removed log is an attack indicator.

dimalinux (approved):
Looks good! Look at my last couple comments/questions before merging. I think there may be a couple minor tweaks that we want.

ethereum/watcher/watcher.go
noot: updated the comment!

I think the loop is fine, since the filter query `ToBlock` is nil so it auto-sets to the current block number, and is also inclusive on that block. so for example if it queries blocks 1-4, `currHeader.Number` is 4, then the next query will have `FromBlock = 5` up until and including whatever the new highest block is

dimalinux: I understand why it works now.  I was proposing that we increment from `prevHeader` (what I'll call `header`), but your code is already doing exactly that. Your initializing `FromBlock` at the end of the loop where `curHeader`, for practical purposes, is `prevHeader`.

Was the goal of doing it this way, instead of initializing `FromHeader` before the filter operation, to have `FromHeader` be `nil` on the 1st loop iteration? Setting `FromHeader` to `nil` searches from the genesis block.  Will that search be fast enough on mainnet?

noot: there isn't a reason to search the logs from genesis - right now it searches from the block which the swap starts at (see swap_state.go newSwapState for both maker/taker). so FromBlock on the first iteration is the start block of the swap

PR #220, opened 2022-11-03
relay updates

dimalinux:
* Updates to the latest go-relayer version which can has a `Stop()` method and can also be terminated by canceling the context it was created with.
* Swapd's RPC server context handling was improved
* Reformats the contracts with `prettier-plugin-solidity@1.0.0-rc.1`
* Adds test helper for checking/updating the goerli contract


cmd/swapd/contract.go
```diff
@@ -74,53 +70,25 @@ func deploySwapFactory(
...
-		return ethcommon.Address{}, nil, err
-	}
-	txOpts, err := bind.NewKeyedTransactorWithChainID(privkey, chainID)
-	if err != nil {
-		return ethcommon.Address{}, nil, fmt.Errorf("failed to make transactor: %w", err)
-	}
-
 	if (forwarderAddress == ethcommon.Address{}) {
 		// deploy forwarder contract as well
-		address, err := deployForwarder(ctx, ec, txOpts) //nolint:govet
```
dimalinux, line 92: The removed code from this file was just moved to the `ethereum` folder (`contracts` package). I needed it for writing tests in that package. The parts that seemed like application specific logic were left here:
* Returning an error if the eth key was not passed (the version in the contracts package assumes nil was not passed)
* Deciding whether to deploy  a forwarder if none was set (`DeploySwapFactory` in the `ethereum` folder will deploy the contract with an all zero forwarder if that's what you ask it to do)
* Writing out the contract address to the file system

cmd/swapd/main.go
```diff
@@ -262,28 +262,21 @@ func setLogLevels(level string) {
 }
 
 func runDaemon(c *cli.Context) error {
+	ctx, cancel := context.WithCancel(c.Context)
+	defer cancel()
+	go signalHandler(ctx, cancel)
```
dimalinux, line 267: The function `d.wait()` wasn't waiting and was only interacting with the context, so I decoupled it from the daemon, renamed it to `signalHandler`, and invoke it earlier.

cmd/swapd/main.go
```diff
@@ -298,30 +291,42 @@ func newEmptyDaemon(ctx context.Context, cancel context.CancelFunc) *daemon {
 }
 
 func (d *daemon) stop() error {
-	if d.database != nil {
-		err := d.database.Close()
```
dimalinux, line 302: I didn't like that the database was getting closed while services that modify the database are still running, so I reversed the shutdown order.  I also didn't want the failure to shut down one service to prevent us from shutting down the other services.  I'm not thrilled with the result, but it was the best I could come up with while restraining the refactor to just this function.

cmd/swapd/main.go
```diff
@@ -423,23 +428,23 @@ func (d *daemon) make(c *cli.Context) error { //nolint:gocyclo
 		DataDir: path.Join(cfg.DataDir, "db"),
 	}
 
-	db, err := db.NewDatabase(dbCfg)
+	sdb, err := db.NewDatabase(dbCfg)
```
dimalinux, line 431: `db` as an imported package name. Just avoiding the naming conflict.

cmd/swapd/main.go
```diff
@@ -423,23 +428,23 @@ func (d *daemon) make(c *cli.Context) error { //nolint:gocyclo
...
 	if err != nil {
 		return err
 	}
-	d.database = db
+	d.database = sdb
 
 	sm := swap.NewManager()
-	backend, err := newBackend(d.ctx, c, env, cfg, devXMRMaker, devXMRTaker, sm, host, ec)
+	swapBackend, err := newBackend(d.ctx, c, env, cfg, devXMRMaker, devXMRTaker, sm, host, ec)
```
dimalinux, line 438: `backend` was an imported package name.

cmd/swapd/main.go
```diff
@@ -681,12 +686,12 @@ func getProtocolInstances(c *cli.Context, cfg common.Config,
 		TransferBack: c.Bool(flagTransferBack),
 	}
 
-	xmrtaker, err := xmrtaker.NewInstance(xmrtakerCfg)
+	xmrTaker, err := xmrtaker.NewInstance(xmrtakerCfg)
```
dimalinux, line 689: `xmrtaker` and `xmrmaker` variable changed to mixed case to not conflict with the imported package name.

cmd/swapd/wait.go
```diff
@@ -1,22 +1,23 @@
...
 	defer signal.Stop(sigc)
 
 	select {
 	case <-sigc:
 		fmt.Println("signal interrupt, shutting down...")
-		d.cancel()
-	case <-d.ctx.Done():
+		cancel()
+	case <-ctx.Done():
 		fmt.Println("protocol complete, shutting down...")
```
dimalinux, line 21: Is it intentional that these are using `fmt.Println` instead of `log.SomeLevel`?

ethereum/check_swap_factory_contract_test.go
```diff
@@ -56,8 +62,9 @@ func TestExpectedSwapFactoryBytecodeHex(t *testing.T) {
 // forwarderAddressIndexes slice of trusted forwarder locations is not updated. Use this
 // test to update the slice.
 func TestForwarderAddressIndexes(t *testing.T) {
-	// arbitrary sentinel address that we search for in the contract byte code
-	trustedForwarder := ethcommon.HexToAddress("0x64e902cD8A29bBAefb9D4e2e3A24d8250C606ee7")
```
dimalinux, line 60: `DeploySwapFactoryWithKey` is registering the forwarder, so the tests are now deploying forwarders instead of using fake addresses.

ethereum/check_swap_factory_contract_test.go
```diff
@@ -93,18 +102,36 @@ func TestCheckSwapFactoryContractCode(t *testing.T) {
...
-		DeployERC20Mock(txOpts, ec, "ERC20Mock", "MOCK", ethcommon.Address{0x1}, big.NewInt(9999))
-	require.NoError(t, err)
-	_, err = block.WaitForReceipt(context.Background(), ec, tx.Hash())
+	// Deploy a forwarder contract and then try to verify it as SwapFactory contract
+	contractAddr := deployForwarder(t, ec, pk)
+	_, err := CheckSwapFactoryContractCode(context.Background(), ec, contractAddr)
+	require.ErrorIs(t, err, errInvalidSwapContract)
+}
+
+func TestGoerliContract(t *testing.T) {
```
dimalinux, line 114: I was trying to make this test file a "one stop shop" to help any of us make all needed updates after the contracts change.  For now you have to manually enable the test unless you want me to enable it by default.

ethereum/contracts/ERC20.sol
```diff
@@ -119,13 +119,10 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
...
-        public
-        view
-        virtual
-        override
-        returns (uint256)
-    {
+    function allowance(
+        address owner,
+        address spender
+    ) public view virtual override returns (uint256) {
```
dimalinux, line 125: Style changed with this release: https://github.com/prettier-solidity/prettier-plugin-solidity/releases/tag/v1.0.0-rc.1

ethereum/deploy_util.go
```diff
@@ -0,0 +1,133 @@
...
+
+	_, err = block.WaitForReceipt(ctx, ec, tx.Hash())
+	if err != nil {
+		return ethcommon.Address{}, err
+	}
+
+	err = registerDomainSeparator(ctx, ec, privkey, address, contract)
+	if err != nil {
+		return ethcommon.Address{}, err
+	}
```
dimalinux, line 87: I didn't look into what this registration does.  Is it free if it is already registered? If not, I'm wondering if we can only do it in one place.  Either here or in `DeploySwapFactoryWithKey`?

ethereum/deploy_util.go
```diff
@@ -0,0 +1,133 @@
...
+
+	log.Infof("deployed SwapFactory.sol: address=%s tx hash=%s", address, tx.Hash())
+
+	return address, sf, nil
+}
+
+// DeployGSNForwarderWithKey deploys and registers the GSN forwarder using the passed
+// private key to pay the gas fees.
+func DeployGSNForwarderWithKey(
```
dimalinux, line 63: Is that a good function name for the specific forwarder that we are deploying?

rpc/server.go
```diff
@@ -107,13 +107,34 @@ func (s *Server) Start() error {
...
+		// immediately without servicing already connected clients.
+		if err := s.httpServer.Shutdown(s.ctx); err != nil {
+			log.Warnf("http server shutdown errored: %s", err)
+		}
+		// We shut down because the context was cancelled, so that's the error to return
+		return s.ctx.Err()
+	case err := <-serverErr:
+		log.Errorf("RPC server failed: %s", err)
+		return err
+	}
```
dimalinux, line 130: I did research on the best way to handle the context with an HTTP server when updating the relayer and put the same solution here.

ethereum/deploy_util.go
```diff
@@ -0,0 +1,133 @@
...
+	address ethcommon.Address,
+	contract *gsnforwarder.Forwarder,
+) error {
+
+	txOpts, err := newTXOpts(ctx, ec, privkey)
+	if err != nil {
+		return err
+	}
+
+	tx, err := contract.RegisterDomainSeparator(txOpts, gsnforwarder.DefaultName, gsnforwarder.DefaultVersion)
```
noot, line 105: now that I'm thinking about this, it should only register the domain separator if it isn't already set. do you want to add that or should I? the domain separators are stored in the forwarder contract map `domains` and is the hash of a few things, done in `registerDomainSeparator`

cmd/swapd/wait.go
noot: no, log should also be fine

ethereum/check_swap_factory_contract_test.go
```diff
@@ -93,18 +102,36 @@ func TestCheckSwapFactoryContractCode(t *testing.T) {
...
+	// comment out the next line to test the default goerli contract
+	t.Skip("requires access to non-vetted external goerli node")
+	const goerliEndpoint = "https://ethereum-goerli-rpc.allthatnode.com"
+	// you can temporarily put a funded goerli key below to deploy the test contract
```
noot, line 118:
```suggestion
	// you can temporarily put a funded goerli private key below to deploy the test contract
```

ethereum/check_swap_factory_contract_test.go
noot: this isn't really a test but just redeploys the contract if its code is out of date, right? maybe it should have a comment that explains that?

ethereum/check_swap_factory_contract_test.go
```diff
@@ -93,18 +102,36 @@ func TestCheckSwapFactoryContractCode(t *testing.T) {
...
+	const goerliKey = ""
+	ec, err := ethclient.Dial(goerliEndpoint)
 	require.NoError(t, err)
+	defer ec.Close()
 
-	_, err = CheckSwapFactoryContractCode(context.Background(), ec, contractAddr)
-	require.ErrorIs(t, err, errInvalidSwapContract)
+	parsedTFAddr, err := CheckSwapFactoryContractCode(context.Background(), ec, common.StagenetConfig.ContractAddress)
+	if errors.Is(err, errInvalidSwapContract) && goerliKey != "" {
```
noot, line 125: should there be an `errors.Is(err, errInvalidSwapContract) && goerliKey == ""` error case?

ethereum/deploy_util.go
noot: yeah, seems good

ethereum/deploy_util.go
noot: we should check if the domain is registered already in `registerDomainSeparator`

rpc/server.go
noot: nice!

noot (approved):
looks good, just one comment about checking the domain. can be done in a follow-up though

ethereum/check_swap_factory_contract_test.go
dimalinux: Unless you add a goerli private key, it only checks the address.  The reason for the skip, is because it is connecting to an outside gonerli endpoint that we don't know much about. It seems reliable so far. Did you have an opinion on removing the `t.Skip` from the test or keeping it?  Another option might be to use an environment variable for an Infura API key that we test for in the golang code.  It looks like there is secrets store for github actions: https://docs.github.com/en/rest/actions/secrets

ethereum/check_swap_factory_contract_test.go
dimalinux: We fail this branch of the `if` on line 132. The alternate branch fails on 134 if the error is not nil.

ethereum/deploy_util.go
dimalinux: I'll create a separate PR for that tomorrow.

ethereum/deploy_util.go
dimalinux: I'll create a separate PR for that tomorrow.

PR #221, opened 2022-11-06
check if the domain separator is registered before registering

dimalinux:
Code now checks if the domain separator is registered before registering.

### DIFF
--- a/ethereum/deploy_util.go
+++ b/ethereum/deploy_util.go
@@ -88,7 +88,7 @@ func isDomainSeparatorRegistered(
 	ec *ethclient.Client,
 	forwarderAddr ethcommon.Address,
 	forwarder *gsnforwarder.Forwarder,
-) (bool, error) {
+) (isRegistered bool, err error) {
 	chainID, err := ec.ChainID(ctx)
 	if err != nil {
 		return false, err
@@ -99,7 +99,8 @@ func isDomainSeparatorRegistered(
 	if err != nil {
 		return false, err
 	}
-	return forwarder.Domains(nil, ds) // isRegistered, error
+	opts := &bind.CallOpts{Context: ctx}
+	return forwarder.Domains(opts, ds)
 }
 
 func registerDomainSeparatorIfNeeded(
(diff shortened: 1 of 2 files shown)


ethereum/deploy_util.go
```diff
@@ -81,28 +75,73 @@ func DeployGSNForwarderWithKey(
...
+	if err != nil {
+		return false, err
+	}
+	name := gsnforwarder.DefaultName
+	version := gsnforwarder.DefaultVersion
+	ds, err := common.GetEIP712DomainSeparator(name, version, chainID, forwarderAddr)
+	if err != nil {
+		return false, err
+	}
+	return forwarder.Domains(nil, ds) // isRegistered, error
```
noot, line 102: same the return values instead of the comment?

also nil `*bind.CallOpts` is okay? did not know that

ethereum/deploy_util.go
```diff
@@ -81,28 +75,73 @@ func DeployGSNForwarderWithKey(
...
+	}
+
+	if !isRegistered {
+		err = registerDomainSeparator(ctx, ec, privKey, forwarderAddr, forwarder)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
noot, line 128:
```suggestion
	if isRegistered {
	    return nil
	}

	return registerDomainSeparator(ctx, ec, privKey, forwarderAddr, forwarder)
```

ethereum/deploy_util.go
dimalinux: It's the same as passing an empty `&bind.CallOpts{}`.  If you are calling a view-only function that, for some reason, accesses `msg.sender`, it won't work.  Looking at the structure, I should have passed it for the context field.  I'll change it:
```
type CallOpts struct {
	Pending     bool            // Whether to operate on the pending state or the last known one
	From        common.Address  // Optional the sender address, otherwise the first account is used
	BlockNumber *big.Int        // Optional the block number on which the call should be performed
	Context     context.Context // Network context to support cancellation and timeouts (nil = no timeout)
}
```

noot (approved):

PR #223, opened 2022-11-07
better json handling of Offer and Info types

dimalinux:
This PR is not for `master`. It's for the `noot/db-swap` branch.

Since we are going to be storing the JSON from the `Offer` and `Info` structs in a non-structured database, this PR adds a version field and cleaner JSON handling (offer IDs and EthAsset addresses are in hex instead of comma separated bytes).

### DIFF
--- a/common/types/ethasset.go
+++ b/common/types/ethasset.go
@@ -1,6 +1,9 @@
 package types
 
 import (
+	"fmt"
+	"strings"
+
 	ethcommon "github.com/ethereum/go-ethereum/common"
 )
 
@@ -14,6 +17,26 @@ func (asset EthAsset) String() string {
 	}
 	// TODO: get name of asset from contract?
 	return ethcommon.Address(asset).Hex()
+}
+
+// MarshalText returns the hex representation of the EthAsset
+func (asset EthAsset) MarshalText() ([]byte, error) {
+	return []byte(asset.String()), nil
+}
+
+// UnmarshalText assigns the EthAsset from the input text
+func (asset *EthAsset) UnmarshalText(input []byte) error {
+	inputStr := string(input)
+	switch {
+	case strings.EqualFold(inputStr, "ETH"):
+		*asset = EthAsset{}
+		return nil
+	case ethcommon.IsHexAddress(inputStr):
+		*asset = EthAsset(ethcommon.HexToAddress(inputStr))
+		return nil
+	default:
+		return fmt.Errorf("invalid asset value %q", inputStr)
+	}
 }
 
 // Address ...
--- a/common/types/hash.go
+++ b/common/types/hash.go
@@ -1,0 +1,35 @@
+package types
+
+import (
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// Hash represents a 32-byte hash
+type Hash = ethcommon.Hash
+
+// EmptyHash is an empty Hash
+var EmptyHash = Hash{}
+
+// IsHashZero returns true if the hash is all zeros, otherwise false
+func IsHashZero(h Hash) bool {
+	return h == EmptyHash
+}
+
+// HexToHash decodes a hex-encoded string into a hash
+func HexToHash(s string) (Hash, error) {
+	h, err := hex.DecodeString(strings.TrimPrefix(s, "0x"))
+	if err != nil {
+		return Hash{}, err
+	}
+	if len(h) != len(Hash{}) {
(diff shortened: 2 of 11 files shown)


common/types/ethasset.go
```diff
@@ -16,6 +19,26 @@ func (asset EthAsset) String() string {
...
+		*asset = EthAsset{}
+		return nil
+	case ethcommon.IsHexAddress(inputStr):
+		*asset = EthAsset(ethcommon.HexToAddress(inputStr))
+		return nil
+	default:
+		return fmt.Errorf("invalid asset value %q", inputStr)
+	}
+}
+
```
dimalinux, line 41: I learned something in this PR:
```
type EthAsset ethcommon.Address     // does not inherit methods (declaration higher up in this file)
type EthAsset = ethcommon.Address   // has all methods available on ethcomon.Address
```
The go-ethereum project has these methods for its Address type, but we are doing special handling of some asset names (only "ETH" for now), so we couldn't have used their methods anyways. The existing declaration type is ideal. With our Hash type, we are doing assignment in the type declaration.

common/types/offer.go
```diff
@@ -2,60 +2,35 @@ package types
...
-	"encoding/json"
 	"fmt"
 
-	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/Masterminds/semver/v3"
 	"golang.org/x/crypto/sha3"
 )
 
-// Hash represents a 32-byte hash
-type Hash [32]byte
```
dimalinux, line 14: I moved everything related to the `Hash` type to `hash.go`

common/types/offer.go
```diff
@@ -2,60 +2,35 @@ package types
...
 
 // Offer represents a swap offer
 type Offer struct {
-	id            Hash
-	Provides      ProvidesCoin
-	MinimumAmount float64
-	MaximumAmount float64
-	ExchangeRate  ExchangeRate
-	EthAsset      EthAsset
+	Version       *semver.Version `json:"version"`
```
dimalinux, line 16: https://github.com/Masterminds/semver

common/types/offer.go
```diff
@@ -2,60 +2,35 @@ package types
...
+	ID            Hash            `json:"offer_id"`
```
dimalinux, line 17: It would have been nice to keep this private, but the amount of extra code needed to keep it private didn't seem to justify the benefits.

common/types/offer.go
```diff
@@ -2,60 +2,35 @@ package types
...
+	Provides      ProvidesCoin    `json:"provides"`
+	MinimumAmount float64         `json:"min_amount"`
+	MaximumAmount float64         `json:"max_amount"`
```
dimalinux, line 20: I'd been wanting to shorten these variables. With the JSON decoupled from the variable names, we can still do it at some point.  :)

common/types/offer.go
```diff
@@ -84,59 +59,6 @@ func (o *Offer) String() string {
...
-	o.MaximumAmount = ou.MaximumAmount
-	o.ExchangeRate = ou.ExchangeRate
-	if ou.EthAsset == "" {
-		// Default to EthAssetETH
-		o.EthAsset = EthAssetETH
-	} else {
-		o.EthAsset = EthAsset(ethcommon.HexToAddress(ou.EthAsset))
-	}
-	return nil
-}
```
dimalinux, line 138: These functions were not moved, they were deleted. The ID field is now public and it has automatic JSON handling from the go-ethereum project.

common/types/hash.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// Hash represents a 32-byte hash
+type Hash = ethcommon.Hash
```
dimalinux, line 12: Using assignment above, we get all the JSON handing, as well as the `String()` and `Hex()` methods of the original type. Only downside is that we can't extend the type with our own methods like `IsZero`.  I changed that method to `IsHashZero` below and it just takes the hash as a parameter.

common/types/hash.go
```diff
@@ -0,0 +1,35 @@
...
+		return Hash{}, err
+	}
+	if len(h) != len(Hash{}) {
+		return Hash{}, fmt.Errorf("invalid len=%d hash", len(h))
+	}
+
+	var hash Hash
+	copy(hash[:], h)
+	return hash, nil
+}
```
dimalinux, line 35: There is a `ethcommon.HexToHash` function that we could use instead, but it doesn't return an error value. The JSON marshaling for `ethcommon.Hash` does throw an error if the string is not hex or if the hex length is not 64 nibbles.

PR #224, opened 2022-11-08
rework backend wallet interfaces

dimalinux:
Right now the [Backend interface](https://github.com/AthanorLabs/atomic-swap/blob/0bdb6f6f2a2995187649ebad58a81663fd56b5ef/protocol/backend/backend.go#L36-L84) is enormous.  Even bigger than it appears in that link due to embedded interfaces. The interface has an "is-a" relationship with both Ethereum and Monero client operations. This can make it difficult to understand the code, as it is not always obvious which wallet or network functions are being used. This PR breaks the wallet operations into 2 separate sub interfaces so the wallet/blockchain being used is always obvious to any code auditor.

Let me know what you think. We can keep some or all of the changes. If nothing else, hopefully this PR will aid us in determining ways to simplify the backend.


cmd/swapd/main.go
```diff
@@ -445,9 +445,9 @@ func (d *daemon) make(c *cli.Context) error { //nolint:gocyclo
 		return err
 	}
 
-	// this closes the monero.WalletClient
-	defer swapBackend.Close()
-	log.Infof("created backend with monero endpoint %s and ethereum endpoint %s", swapBackend.Endpoint(), ethEndpoint)
+	defer swapBackend.MoneroClient().Close()
```
dimalinux, line 448: The name `MoneroClient()` might be too long.  Suggestions welcome!

protocol/backend/backend_test.go
```diff
@@ -18,12 +19,15 @@ func TestWaitForReceipt(t *testing.T) {
 	ec, chainID := tests.NewEthClient(t)
 	privKey := tests.GetTakerTestKey(t)
 
+	gasPrice, err := ec.SuggestGasPrice(ctx)
```
dimalinux, line 22: I learned about this function, so I just put it in here.

rpc/server.go
```diff
@@ -111,8 +114,7 @@ func (s *Server) Start() error {
 	go func() {
 		// Serve never returns nil. It returns http.ErrServerClosed if it was terminated
 		// by the Shutdown.
-		err := s.httpServer.Serve(s.listener)
-		serverErr <- fmt.Errorf("RPC server failed: %w", err)
```
dimalinux, line 115: We were already adding this prefix in the log below.

tests/integration.go
```diff
@@ -1,3 +0,0 @@
-package tests
-
-// need this file or we get `no non-test Go files in ~/go/src/github.com/athanorlabs/atomic-swap/tests`
```
dimalinux, line 3: File was no longer needed since we have `ganache.go`

net/message/message.go
```diff
@@ -124,7 +124,7 @@ type SendKeysMessage struct {
 
 // String ...
 func (m *SendKeysMessage) String() string {
-	return fmt.Sprintf("SendKeysMessage OfferID=%s ProvidedAmount=%v PublicSpendKey=%s PublicViewKey=%s PrivateViewKey=%s DLEqProof=%s Secp256k1PublicKey=%s EthAddress=%s", //nolint:lll
+	return fmt.Sprintf("SendKeysMessage OfferID=%s ProvidedAmount=%v PublicSpendKey=%s PublicViewKey=%s PrivateViewKey=%s DLEqProof=%s Secp256k1PublicKey=%s Address=%s", //nolint:lll
```
noot, line 127: why changed?

protocol/backend/backend.go
```diff
@@ -2,82 +2,51 @@ package backend
...
-	log                    = logging.Logger("protocol/backend")
 	defaultTimeoutDuration = time.Hour * 24
 )
 
 // Backend provides an interface for both the XMRTaker and XMRMaker into the Monero/Ethereum chains.
 // It also interfaces with the network layer.
 type Backend interface {
-	monero.WalletClient
+	XMR() monero.WalletClient
+	ETH() extethclient.EthClient
```
noot, line 29:
```suggestion
	XMRClient() monero.WalletClient
	ETHClient() extethclient.EthClient
```

nit but I'd rather name them something like this? otherwise it's a bit unclear imo

protocol/backend/backend.go
```diff
@@ -89,23 +58,14 @@ type backend struct {
 	swapManager swap.Manager
 
 	// monero endpoints
```
noot, line 60: update comment?

noot (approved):
I think this is an improvement, I like how it moves the eth-specific stuff to its own type.

protocol/backend/backend.go
dimalinux: It was one of the things I wanted to discuss with you. I originally went with what you were suggesting, but they are used in so many places that brevity might be a bigger code-readability win. It's a compromise either way. I can create another branch so you can see the difference side-by-side.

net/message/message.go
dimalinux: I'll fix that. Automated refactoring overreach. :)

PR #225, opened 2022-11-08
add FakeDLEq; add pure go + darwin build commands

noot:
this PR attempts to fix the build for non-linux platforms by removing the Rust dependency completely.

it adds make commands `make build-go` which builds the pure go project and `make build-go-darwin` which builds the pure go project for `GOOS=darwin` and `GOARCH=arm64`.

**WARNING**: the pure go build should not be used in reality, as the DLEq prover has been stubbed out and now proves nothing.


monero/wallet_client.go
```diff
@@ -496,7 +496,7 @@ func launchMoneroWalletRPCChild(walletRPCBin string, walletRPCBinArgs ...string)
 	cmd.Env = append(os.Environ(), "LANG=C", "LC_ALL=C", "TERM=dumb")
 
 	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Pdeathsig: syscall.SIGTERM,
+		//Pdeathsig: syscall.SIGTERM,
```
noot, line 499: @dimalinux I had to comment this out since it doesn't build for `GOOS=darwin`

dimalinux: The goal here is that if `swapd` (or a unit test for that matter) is killed in a way that doesn't shut down the wallet, the Linux OS will send SIGTERM to the `monero-wallet-rpc` process. You can read more about it under `PR_SET_PDEATHSIG` in the `prctl` man page. We could separate out the `SysProcAttr` handling into different files with different OS build tags.

dimalinux:
> WARNING: the pure go build should not be used in reality, as the DLEq prover has been stubbed out and now proves nothing.

What is the goal if the build result is dangerous? How do we prevent people from using it once the build is fixed?

Makefile
```diff
@@ -61,6 +61,15 @@ install: init
...
 	./scripts/build.sh
 
+.PHONY: build-go
+build-go:
+	go build ./cmd/swapd
+	go build ./cmd/swapcli
+
+.PHONY: build-go-darwin
+build-go-darwin:
+	GOOS=darwin GOARCH=arm64 $(MAKE) build-go
```
dimalinux, line 71: I'm a little confused.  `build`, `build-all`, `build-go` and `build-go-darwin` are all go specific, but the `build-*` targets depend on `init` which builds the rust code. Which targets are supposed to be setting `cgodleq`?

noot: yeah, I agree this is confusing, I updated the tags so `cgodleq` is default again, and `build-go` and `build-go-darwin` use `-tags=fakedleq`

dimalinux: I don't see anyone setting `-tags=fakedleq`. Did you push that change? If you want me to set an environment variable here in the Makefile and update the script to use it, Iet me know.

noot: it didn't push for some reason, updated

dimalinux (approved):
If these targets are needed longer term, we should rename them to something like `build-fakedleq` and `build-fakedleq-darwin` to decrease the chance of someone getting confused and thinking that the resulting executables are safe to perform swaps with.

Issue #212, opened 2022-10-27
auto-recover on startup if incomplete swaps are detected


noot:
follow up to #83 and #203 - since we're now storing swaps and their status (basically completed or not) to disk, we can detect if we have incomplete swaps on start-up and auto-recover if so. this would prevent users from having to manually use the `swaprecover` binary, and perhaps it could eventually be deprecated.

PR #227, opened 2022-11-13
store swap recovery information in database, remove "infofile"

### LINKED ISSUE #202: deprecate `contract-addresses.json`
currently, if swapd deploys a swap contract (for testing purposes) it gets written to `contract-address.json`. however, this isn't really needed, as contract addresses used in swaps are stored in the swap info file.

### LINKED ISSUE #203: store info file contents in database

noot:
as title says, also:
- implement `RecoveryDB` for storing the info instead
- note: this PR removes the `swaprecover` binary as I couldn't be bothered to fix it as it's going to be removed in a follow up (#212)

closes #203


rpcclient/wsclient/wsclient.go
```diff
@@ -255,26 +255,35 @@ func (c *wsClient) TakeOfferAndSubscribe(multiaddr, offerID string,
 		return nil, err
 	}
 
-	// read ID from connection
-	message, err := c.read()
-	if err != nil {
-		return nil, fmt.Errorf("failed to read websockets message: %s", err)
-	}
+	readMessageFn := func() (string, error) {
```
dimalinux, line 258: `TakeOfferAndSubscribe` is 172 lines lines. Maybe we could pull this code out into it's own function and then the code auditor doesn't have to carefully read it to understand which variables are shared with the outer scope?

protocol/xmrmaker/net.go
```diff
@@ -73,6 +70,9 @@ func (b *Instance) initiate(
 
 // HandleInitiateMessage is called when we receive a network message from a peer that they wish to initiate a swap.
 func (b *Instance) HandleInitiateMessage(msg *net.SendKeysMessage) (net.SwapState, net.Message, error) {
+	b.swapMu.Lock()
+	defer b.swapMu.Unlock()
```
dimalinux, line 74: So these were the missing lines that were needed to prevent the race condition when 2 parties try to take the same offer simultaneously?

noot: maybe, I don't think it fully fixes it, I think the issue is more how the test is written or another issue - when I looked at the logs, the issue was that one of the takers was hanging.

db/recovery_db.go
```diff
@@ -0,0 +1,108 @@
...
+	return db.db.Put(key[:], val)
+}
+
+// PutContractSwapInfo stores the given contract swap ID (which is not the same as the daemon
+// swap ID, but is instead a hash of the `SwapFactorySwap` structure)
+// and contract swap structure for the given swap ID.
+func (db *RecoveryDB) PutContractSwapInfo(id types.Hash, swapID [32]byte, swap contracts.SwapFactorySwap) error {
+	val, err := json.Marshal(&struct {
+		SwapID [32]byte
+		Swap   contracts.SwapFactorySwap
```
dimalinux, line 69: Can we pull this out, give it the semantic version field, add JSON annotations, use a hash type that will auto-serialize as hex, and add at least one unit test for the JSON marshaling so you can see what the JSON looks like when reading the code? I is a little painful, but the odds of the format never changing are probably low. Having the ID in hex will help us troubleshoot. I can update the PR later today if you want.

rpcclient/wsclient/wsclient.go
noot: updated!

protocol/xmrtaker/recovery.go
```diff
@@ -57,7 +56,6 @@ func NewRecoveryState(b backend.Backend, dataDir string, secret *mcrypto.Private
 		walletScanHeight: 0, // TODO: Can we optimise this?
```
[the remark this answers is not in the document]
noot: yeah, I realized when trying to implement another issue we need the eth start height as well, will add

db/recovery_db.go
noot: updated, but I'm not sure if I see the point of adding the version necessarily since this info is likely not going to be persisted long-term - it's only stored while a swap is ongoing, and will end up being deleted after

dimalinux: Yeah, then it isn't needed. I thought a swap might be created with one version of `swapd` and might possibly be recovered by a newer `swapd` version.  I didn't add it in `dimalinux/db-swap-info-json`.

db/types.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/athanorlabs/atomic-swap/common/types"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// EthereumSwapInfo represents information required on the Ethereum side in case of recovery
+type EthereumSwapInfo struct {
+	StartNumber     *big.Int                  `json:"startNumber"`
```
dimalinux, line 14: Can we document and, if appropriate, rename this field?  Is it the block number of the `newSwap` transaction? Is it the same for the maker and the taker? I wanted to rename it in `dimalinux/db-swap-info-json`, but I didn't know what it was without digging, so I left it.

dimalinux (approved):
Looks good. I provided a PR with human friendly JSON serialization of `SwapFactorySwap` for merging into this PR's branch. I also left one last comment/request about the `StartNumber` field of `EthereumSwapInfo` in this PR.

db/types.go
noot: yeah it's the block number of the `newSwap`  tx, is the same for both, will update

PR #231, opened 2022-11-14
handle decimals of ERC20 tokens properly

### LINKED ISSUE #186: ERC20 support: handle decimal places other than 18
Currently, we parse the taker's `providesAmount` to [18 decimal places](https://github.com/AthanorLabs/atomic-swap/blob/000b734f55061218dad8593306efdb89c9bb5f1a/protocol/xmrtaker/net.go#L20). Not all tokens have 18 decimal places, like USDT/USDC have 6, so if we tried to swap those using the CLI or RPC, we would be swapping more than we wanted to.

We could (1)
- Specify the number of decimal places a token has alongside its address in the trusted token list
- Check this number against the number returned by the RPC

Including the number of decimal places a token has in our list will protect against a malicious RPC server which might return more decimals than are really there.

Or, (2) we choose to support only tokens that have 18 decimal places. DAI or TrueUSD are stablecoins with 18 decimal places that would be alternatives to USDC or USDT. This is the less robust approach but requires no effort or UI changes. It is unknown whether token demand is strong or not anyway.

It might be useful if the taker RPC calls/initiate request handle the amount to send with big ints instead of floats. That way the decimal places only need to be decided in the UI.

noot:
closes #186


common/coins.go
```diff
@@ -61,13 +61,9 @@ func (a EtherAmount) AsEther() float64 {
...
 
-// ToDecimals returns the amount rounded to a fixed number of decimal places
-func (a EtherAmount) ToDecimals(decimals uint8) float64 {
-	decimalsValue := big.NewInt(0).Exp(big.NewInt(10), big.NewInt(int64(decimals)), nil)
-	wei := big.NewFloat(0).SetInt(a.BigInt())
-	ether := big.NewFloat(0).Quo(wei, big.NewFloat(0).SetInt(decimalsValue))
-	res, _ := ether.Float64()
-	return res
+// AsStandard returns the wei amount as ether
+func (a EtherAmount) AsStandard() float64 {
```
dimalinux, line 65: We had discussed renaming `EtherAmount` to `WeiAmount` and `MoneroAmount` to `PiconeroAmount` so the type names match the value being stored. Would this PR be a good time to make that change?

common/coins.go
```diff
@@ -80,3 +76,57 @@ func (a EtherAmount) String() string {
...
+type ERC20TokenAmount struct {
+	amount   *big.Int
+	numUnits float64 // 10^decimals
+}
+
+// NewERC20TokenAmountFromBigInt converts some amount in the smallest token denomination into an ERC20TokenAmount.
+func NewERC20TokenAmountFromBigInt(amount *big.Int, decimals float64) *ERC20TokenAmount {
+	return &ERC20TokenAmount{
+		amount:   amount,
+		numUnits: math.Pow(10, decimals),
```
dimalinux, line 90: `math.Pow10(decimals)`

common/coins.go
```diff
@@ -80,3 +76,57 @@ func (a EtherAmount) String() string {
...
+	}
+}
+
+// NewERC20TokenAmount converts some amount in the smallest token denomination into an ERC20TokenAmount.
+func NewERC20TokenAmount(amount int64, decimals float64) *ERC20TokenAmount {
+	return &ERC20TokenAmount{
+		amount:   big.NewInt(amount),
+		numUnits: math.Pow(10, decimals),
```
dimalinux, line 98: `math.Pow10(decimals)`

common/coins.go
```diff
@@ -80,3 +76,57 @@ func (a EtherAmount) String() string {
...
+	}
+}
+
+// NewERC20TokenAmountFromDecimals converts some amount of standard token in standard format
+// to its smaller denomination.
+func NewERC20TokenAmountFromDecimals(amount float64, decimals float64) *ERC20TokenAmount {
+	numUnits := math.Pow(10, decimals)
```
dimalinux, line 105: `math.Pow10(decimals)`

common/coins.go
```diff
@@ -80,3 +76,57 @@ func (a EtherAmount) String() string {
...
+	amt := big.NewFloat(amount)
+	mult := big.NewFloat(numUnits)
+	res, _ := big.NewFloat(0).Mul(amt, mult).Int(nil)
```
dimalinux, line 108: When writing big.Int/big.Float code, I think the preferred way to create zero values is `new(big.Int)` and `new(big.Float)`. It executes less code and is a few less characters. It is what the standard and other polished libraries use. Feel free to change it or leave it.  I won't write it on every use.

dimalinux: Is this code supposed to be truncating (what it is doing) or rounding? Can we clearly document the expected behavior?

common/coins_test.go
```diff
@@ -27,9 +27,12 @@ func TestEtherAmount(t *testing.T) {
 	require.Equal(t, amountUint, etherAmount.BigInt().Int64())
 }
 
-func TestToDecimals(t *testing.T) {
-	val := NewEtherAmount(123456)
-	require.Equal(t, "1.23456", FmtFloat(val.ToDecimals(5)))
-	val = NewEtherAmount(1234567890)
-	require.Equal(t, "1234.56789", FmtFloat(val.ToDecimals(6)))
+func TestERC20TokenAmount(t *testing.T) {
+	amount := float64(33.999999999)
```
dimalinux, line 31: Can we add some more test values here like `33.000000005` and `33.0000000005`?

common/coins.go
noot: yep, done!

common/coins.go
noot: all updated

common/coins.go
noot: updated, and added a comment explaining it.

common/coins_test.go
noot: yep, added more cases

common/coins.go
```diff
@@ -80,3 +77,76 @@ func (a EtherAmount) String() string {
...
+
+// String ...
+func (a *ERC20TokenAmount) String() string {
+	return a.BigInt().String()
+}
+
+// round rounds the input *big.Float to a *big.Int.
+// eg. if the input is 33.49, it returns 33.
+// if the input is 33.5, it returns 34.
+func round(num *big.Float) *big.Int {
```
dimalinux, line 141: There's another possible algorithm here if you like it better (not saying it is better and I didn't test it): https://stackoverflow.com/a/60810320/2049639

dimalinux (approved):
Looks good!

PR #232, opened 2022-11-16
misc fixes

dimalinux:
Misc updates/fixes:
* Run relayer in integration tests as a dedicated binary (had dependencies on this PR: https://github.com/AthanorLabs/go-relayer/pull/6)
* Remove warnings in our github actions
* Update libp2p libs (work that was done in `dimalinux/libp2p-nat-traversal` that we want to keep)
* Clean-up clean target
* Ports some signal handing fixes made in the relayer to atomic-swap
* Cherry-picks the xmrmaker instance lock fix from `noot/db-swap-info`

Note, since we have a top level folder named `relayer`, it wasn't possible to place the binary in the root folder. All of the executables that we build are now placed in a `bin` folder at the root of the project.


.github/workflows/build-macos.yaml
```diff
@@ -11,13 +11,13 @@ jobs:
     strategy:
       matrix:
         go-version: [1.19.x]
-        platform: [macos-latest]
+        platform: [macOS-12]
```
dimalinux, line 14: 12 is the latest they support, `latest` currently points to 11, but they plan on updating it to 12 and are giving a warning to discourage users from using the `latest` tag.

cmd/swapd/signal_handler.go
```diff
@@ -0,0 +1,29 @@
+package main
```
dimalinux, line 1: This isn't a new file.  I renamed `wait.go` to `signal_handler.go` and made some updates.

scripts/build.sh
```diff
@@ -3,8 +3,23 @@
...
 if [[ -n "${ALL}" ]]; then
-	GOBIN="${PROJECT_ROOT}" go install -tags=prod ./cmd/...
+	go install -tags=prod ./cmd/...
 else
-	GOBIN="${PROJECT_ROOT}" go install -tags=prod ./cmd/swapd ./cmd/swapcli
+	go install -tags=prod ./cmd/swapd ./cmd/swapcli
 fi
+
+# Will change the suffix below to @latest before merging
+go install github.com/athanorlabs/go-relayer/cmd/relayer@95cca1a2b8b2b16a78f9a0d8fa367cf2e9c95a4f
```
dimalinux, line 25: I'll remove the commit hash after we merge this PR:
https://github.com/AthanorLabs/go-relayer/pull/6

Note: Normally you could put a branch name after the `@` instead of a commit hash which requires more frequent updating, but they don't allow you to have a slash in the branch name and I was using `dimalinux/key-handling`.

scripts/run-integration-tests.sh
```diff
@@ -7,23 +7,84 @@ cd "${PROJECT_ROOT}" || exit 1
...
+RELAYER_PORT=7799
+
+# First 5 deterministic ganache keys (only 4 used at the moment)
+GANACHE_KEYS=(
+	"4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d" # Key 0
+	"6cbed15c793ce57650b9877cf6fa156fbef513c4e6134f022a85b1ffdd59b2a1" # Key 1
+	"6370fd033278c143179d81c5526140625662b8daa446c22ee2d73db3707e620c" # Key 2
+	"646f1ce2fdad0e6deeeb5c7e8e5543bdde65e86029e2fd9fc169899c440a7913" # Key 3
+	"add53f9a7e588d003326d1cbf9e4a43c061aadd9bc938c843a79e7b4fd2ad743" # Key 4
+)
```
dimalinux, line 20: The first five keys in `ganache.go` are commented out and these are reserved for the integration tests and/or code that isn't in a `*_test.go` file.

scripts/run-integration-tests.sh
```diff
@@ -7,23 +7,84 @@ cd "${PROJECT_ROOT}" || exit 1
...
 source "scripts/testlib.sh"
 check-set-swap-test-data-dir
-mkdir -p "${SWAP_TEST_DATA_DIR}/"{alice,bob,charlie}
 
-# Charlie uses deterministic ganache key #49
-CHARLIE_ETH_KEY="${SWAP_TEST_DATA_DIR}/charlie/eth.key"
-echo "87c546d6cb8ec705bea47e2ab40f42a768b1e5900686b0cecc68c0e8b74cd789" >"${CHARLIE_ETH_KEY}"
+RELAYER_PORT=7799
+
+# First 5 deterministic ganache keys (only 4 used at the moment)
```
noot, line 13:
```suggestion
# First 5 deterministic ganache keys (only 4 using at the moment)
```

noot: also we can just have 4 keys here for now, if only 4 are used?

dimalinux: I over-reserved 5 keys in `tests/ganache.go` so that there would be a spare key for future use that wouldn't require modifying anything other than this script. Do you want me to change it in both places?

scripts/run-integration-tests.sh
```diff
@@ -7,23 +7,84 @@ cd "${PROJECT_ROOT}" || exit 1
...
 
 # This is the local multiaddr created when using ./tests/alice-libp2p.key on the default libp2p port
 ALICE_MULTIADDR=/ip4/127.0.0.1/tcp/9933/p2p/12D3KooWAAxG7eTEHr2uBVw3BDMxYsxyqfKvj3qqqpRGtTfuzTuH
 ALICE_LIBP2PKEY=./tests/alice-libp2p.key
 LOG_LEVEL=debug
 
+start-relayer() {
+	local log_file="${SWAP_TEST_DATA_DIR}/relayer.log"
+	echo "Starting relayer with logs in ${log_file}"
+	./bin/relayer \
```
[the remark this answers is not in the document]
dimalinux: This is in `atomic-swap/bin/`. This PR is placing all the binaries created by `make build` in that folder.  We are [matching the relayer binary](https://github.com/AthanorLabs/atomic-swap/blob/diamlinux/misc-fixes/scripts/build.sh#L24-L27) to the version declared in `go.mod`, so breaking changes in the relayer won't require immediate/synchronized updates in `atomic-swap`. You may want to have two versions of `atomic-swap` checked out on the same host that may, legitimately, require different versions of the relayer. (That was my logic for the decision at least.)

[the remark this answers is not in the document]
noot: okay, that seems reasonable, thanks for the explanation!

noot (approved):
looks good!

PR #233, opened 2022-11-18
Modify scripts to run on MacOS

Janaka-Steph:
Not intended for merge since dockerizing with a Linux image is a better approach!
I just wanted to document the changes I had to make in order to run the scripts on MacOS

### DIFF
--- a/docs/local.md
+++ b/docs/local.md
@@ -51,7 +51,7 @@ Now get the ethereum contract address that Alice deployed to. This can be pulled
 the file ..., or if you have `jq` installed (available via `sudo apt install jq`), you can set a
 variable like this:
 ```bash
-CONTRACT_ADDR=$(jq -r .ContractAddress /tmp/xmrtaker/contract-address.json)
+CONTRACT_ADDR=$(jq -r .ContractAddress "${TMPDIR}"/xmrtaker-581848395/contract-address.json)
 ```
 
 Now start Bob's swapd instance in terminal 2:
--- a/scripts/cleanup-test-processes.sh
+++ b/scripts/cleanup-test-processes.sh
@@ -2,12 +2,12 @@
 # This script can be handy to kill test processes launched manually
 # for debugging or still hanging around for other reasons.
 
-pkill --echo --uid "${UID}" --full '/monerod .* --regtest '
-pkill --echo --uid "${UID}" --full '/ganache.* --deterministic '
+pkill -U "${UID}" -f '/monerod .* --regtest '
+pkill -U "${UID}" -f '/ganache.* --deterministic '
 
 # If you have monero-wallet-rpc or swapd processes owned by the current user
 # that you don't want to kill, don't use this script!
-if pkill --echo --uid "${UID}" --full '/swapd '; then
+if pkill -U "${UID}" -f '/swapd '; then
 	# If swapd instances were killed, give the monero-wallet-rpc instances
 	# some time to shutdown.
 	sleep 4 # Give time for wallet to shutdown
@@ -15,7 +15,7 @@ fi
 
 # Take note of monero-wallet-rpc instances being killed here if the instances
 # were started by swapd
-pkill --echo --uid "${UID}" --full '/monero-wallet-rpc '
+pkill -U "${UID}" -f '/monero-wallet-rpc '
 
 # Don't use the exit value of the last pkill, since it will exit with
 # non-zero value if there were no processes to kill.
--- a/scripts/install-monero-linux.sh
+++ b/scripts/install-monero-linux.sh
@@ -4,7 +4,7 @@
 # deleting the monero-bin symlink or the version specific folder that it links to. This
 # script changes directories and should be executed, not sourced.
 #
-arch=linux64
+arch=macarm8
 
 PROJECT_ROOT="$(dirname "$(dirname "$(readlink -f "$0")")")"
 cd "${PROJECT_ROOT}"
--- a/scripts/testlib.sh
+++ b/scripts/testlib.sh
@@ -62,7 +62,7 @@ mine-monero-for-swapd() {
 
 check-set-swap-test-data-dir() {
 	if [[ -z "${SWAP_TEST_DATA_DIR}" ]]; then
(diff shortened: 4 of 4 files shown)


docs/local.md
```diff
@@ -30,7 +56,7 @@ Warning: the command below will kill running instances of `ganache`, `monerod`,
 To avoid confusion, delete any data directories from old runs of `swapd` that used
 the flags `--dev-xmrtaker` or `--dev-xmrmaker`:
 ```bash
-rm -rf /tmp/xmrtaker-* /tmp/xmrmaker-*
+rm -rf "${TMPDIR:-/tmp}"/xmr[mt]aker-*
```
dimalinux, line 59: The above will use the value of TMPDIR, which is always defined on macOS, or `/tmp` if TMPDIR is not defined.

ethereum/block/testdata/generate-bindings.sh
```diff
@@ -1,9 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
dimalinux, line 1: Macos 12 has a legacy version of bash installed (3.2 from 2007) and they plan on removing bash entirely (default shell is now zsh). All the scripts were adjusted so that the first `bash` executable in the user's path is used. You can easily install bash 5.2 using brew on macos.

ethereum/block/testdata/generate-bindings.sh
```diff
@@ -1,9 +1,9 @@
...
 # Generate the UTContract.sol Go bindings into a file named ut_contract_test.go of the
 # parent directory for use by unit tests.
 
 # Use the contract's directory (where this script is) as the current working directory:
-cd "$(dirname "$(readlink -f "$0")")"
+cd "$(dirname "$(realpath "$0")")"
```
dimalinux, line 6: These commands do the same thing, but `readlink` on macOS doesn't support the `-f` flag.  `realpath` is installed by default in Linux distributions, and by brew/coreutils on macOS.

scripts/cleanup-test-processes.sh
```diff
@@ -1,22 +1,34 @@
...
 
-pkill --echo --uid "${UID}" --full '/monerod .* --regtest '
-pkill --echo --uid "${UID}" --full '/ganache.* --deterministic '
-pkill --echo --uid "${UID}" --full '/relayer'
+pkill_cmd=(pkill --echo --uid "${UID}" --full)
+if [[ "$(uname)" == 'Darwin' ]]; then
+	pkill_cmd=(pkill -l -U "${UID}" -f)
+fi
+
+echo "Stoping any monerod regest instances"
```
dimalinux, line 10: There is no `--echo` flag equivalent for the macOS `pkill` command, so I added these echos. They are arguably helpful even on Linux.

scripts/install-monero-linux.sh
```diff
@@ -1,12 +1,27 @@
...
+elif [[ "${os_name}" == Linux ]] && [[ "${arch_name}" == aarch64 ]]; then
+	xmr_arch=linuxarm8
+elif [[ "${os_name}" == Darwin ]] && [[ "${arch_name}" == x86_64 ]]; then
+	xmr_arch=mac64
+elif [[ "${os_name}" == Darwin ]] && [[ "${arch_name}" == arm64 ]]; then
+	xmr_arch=macarm8
+else
+	echo "OS=${os_name} on ARCH=${arch_name} is not supported"
+	exit 1
+fi
```
dimalinux, line 22: I tested on my Raspberry PI, and with this PR, we are good to go on Linux/ARM:
```
pi$ uname -s -m 
Linux aarch64
```

scripts/testlib.sh
```diff
@@ -165,7 +165,11 @@ start-ganache() {
 	# shellcheck disable=SC2155
 	local ganache_exec="$(npm config get prefix)/bin/ganache"
 	check-set-swap-test-data-dir
-	NODE_OPTIONS="--max_old_space_size=8192" nohup \
+	local nohup_cmd=nohup
+	if [[ "$(uname)" == 'Darwin' ]]; then
+		nohup_cmd=
+	fi
```
dimalinux, line 171: Not sure what was going on here.  Using nohup was causing the ganache command to fail:
```
nohup: can't detach from console: Inappropriate ioctl for device
```
I didn't try from a graphical console. All the work was done over ssh.

As a side note, you can run ganache from docker, eliminating the need to install ganache:
```
docker run --name ganache -p 8545:8545 --detach --rm trufflesuite/ganache:latest --deterministic --accounts=50 --miner.blockTime=1
```

dimalinux (approved):

docs/local.md
```diff
@@ -65,7 +91,7 @@ Now get the ethereum contract address that Alice deployed to. This can be pulled
 the file ..., or if you have `jq` installed (available via `sudo apt install jq`), you can set a
 variable like this:
 ```bash
-CONTRACT_ADDR=$(jq -r .ContractAddress /tmp/xmrtaker-*/contract-address.json)
+CONTRACT_ADDR=$(jq -r .ContractAddress "${TMPDIR-/tmp}"/xmrtaker-*/contract-address.json)
```
noot, line 94: `"${TMPDIR-/tmp}"` - what is this syntax? is it just the os tempdir?

scripts/cleanup-test-processes.sh
noot, line 10:
```suggestion
echo "Stopping any monerod regtest instances"
```

noot: same for the other `echo`s below

noot (approved):
looks good, only minor typo comments!

docs/local.md
dimalinux: `TMPDIR` is always defined on macOS, and it is something that `mktemp` and the Go libraries will use if it is defined on Linux (but it is not typically defined on Linux systems).  The bash syntax here uses the variable before the `-` if the variable is defined and non-empty, or the value after the `-` if `TMPDIR`is not defined.  Effectively, `TMPDIR` will get used on macOS while `/tmp` will be used on typical Linux systems.

PR #234, opened 2022-11-19
provides custom JSON marshalling for SwapFactorySwap type

dimalinux:


db/types.go
```diff
@@ -11,8 +11,8 @@ import (
 
 // EthereumSwapInfo represents information required on the Ethereum side in case of recovery
 type EthereumSwapInfo struct {
-	StartNumber     *big.Int                  `json:"startNumber"`
-	SwapID          types.Hash                `json:"swapID"`
+	StartNumber     *big.Int                  `json:"start_number"`
+	SwapID          types.Hash                `json:"swap_id"`
 	Swap            contracts.SwapFactorySwap `json:"swap"`
-	ContractAddress ethcommon.Address         `json:"contractAddress"`
+	ContractAddress ethcommon.Address         `json:"contract_address"`
```
dimalinux, line 17: All the other database marshalling is using snake case. That may have been a mistake on my part, as I see our JSON RPC APIs are using camel case. Recently, I'd been spending far more time looking at the Monero APIs which use snake. This makes the naming uniform at least for the database types. If we want to use camel, we should update them for all the database types.

noot: yeah I'd prefer camel, I originally used camel since that's what ethereum uses

noot (approved):
looks good, we can change to camel case in a follow up, not a huge deal

PR #235, opened 2022-11-19
implement auto-recover for xmrmaker

### LINKED ISSUE #163: periodically check if contract gets set to ready when waiting for t0 in xmrmaker.`tryClaim`
see title, if the contract gets set to ready while we're waiting for t0 then just go straight to `claimFunds()`.

noot:
- update `xmrmaker.Instance` to check if there are incompleted swaps on creation; if there are, then load required info from the RecoveryDB and create a swap state for it 
- note: only swaps where we've locked XMR but haven't gotten any funds are recovered; otherwise, they're just aborted and the counterparty to refund. I could update this in a follow-up to also recover if ETH has been locked but no XMR has.
- implement `newSwapStateFromOngoing` which creates a new swap state for an ongoing swap, rename previous `newSwapState` to `newSwapStateFromStart`
- remove `recovery.go` as it's unused, the logic is cleaner now as it re-uses everything from `swapState` and basically just continues the swap where it leaves off

TODO:
- [x] remove swap info from recoveryDB when swap completes
- [x] update offerManager to only delete offer from db when swap actually completes
- [x] write relayer info in `offerExtra` to db as well, reload if doing recovery
- [x] write swap status to disk every time it updates (maybe refactor to have a map of id -> status directly?)
- [x] check if we have shared secret key in db; if so, recover XMR from that directly

related to #212 - will do xmrtaker in another PR since there's already a lot of changes
closes #163 as it changes the recovery logic to just use a normal `swapState`


protocol/xmrmaker/checks.go
```diff
@@ -80,3 +84,90 @@ func checkContractSwapID(msg *message.NotifyETHLocked) error {
 
 	return nil
 }
+
+// checkContract checks the contract's balance and Claim/Refund keys.
+// if the balance doesn't match what we're expecting to receive, or the public keys in the contract
+// aren't what we expect, we error and abort the swap.
+func (s *swapState) checkContract(txHash ethcommon.Hash) error {
```
noot, line 91: this hasn't been modified, I just copy pasted it from swap_state.go since that file was getting long

protocol/xmrmaker/instance_test.go
```diff
@@ -0,0 +1,185 @@
...
+		RecoveryDB:          rdb,
+	}
+
+	b, err := backend.NewBackend(bcfg)
+	require.NoError(t, err)
+
+	return b
+}
+
+func newTestInstanceAndDB(t *testing.T) (*Instance, *offers.MockDatabase) {
```
noot, line 111: this function and all helper funcs above were copy pasted from swap_state_test.go

common/types/offer.go
```diff
@@ -68,9 +68,9 @@ func (o *Offer) IsSet() bool {
 
 // OfferExtra represents extra data that is passed when an offer is made.
 type OfferExtra struct {
-	StatusCh          chan Status
-	RelayerEndpoint   string
-	RelayerCommission float64
+	StatusCh          chan Status `json:"-"`
+	RelayerEndpoint   string      `json:"relayerEndpoint"`
+	RelayerCommission float64     `json:"relayerCommission"`
```
dimalinux, line 73: Is the `RelayerCommission` used anywhere that a slight difference in the bits that are written out during serialization and the bits that are read back in could cause an issue?

db/recovery_db.go
```diff
@@ -16,6 +16,7 @@ const (
 	moneroHeightPrefix         = "xmrheight"
 	swapPrivateKeyPrefix       = "privkey"
 	sharedSwapPrivateKeyPrefix = "sprivkey"
+	relayerInfoPrefix          = "relayer"
 )
 
 // RecoveryDB contains information about ongoing swaps requires for recovery
```
dimalinux, line 22: requires => required

ethereum/block/wait_for_receipt.go
```diff
@@ -29,6 +30,10 @@ var (
 // we return an error describing why.
 func WaitForReceipt(ctx context.Context, ec *ethclient.Client, txHash ethcommon.Hash) (*ethtypes.Receipt, error) {
 	for i := 0; i < maxRetries; i++ {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
```
dimalinux, line 35: Is that an acceptable way of checking the context status?  I've been writing this, which consumes more lines, but only because that's the way I've seen other code doing the check. I don't know what is considered best practice.
```go
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	default:
	}

```

protocol/xmrmaker/instance.go
```diff
@@ -57,15 +60,126 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to get monero start height for ongoing swap, id %s: %s", s.ID, err)
+	}
+
+	sharedKey, err := b.backend.RecoveryDB().GetSharedSwapPrivateKey(s.ID)
+	if err == nil {
+		b.backend.XMRClient().Lock()
+		defer b.backend.XMRClient().Unlock()
+
+		// TODO: do we want to transfer this back to the original account?
```
dimalinux, line 119: That is a definite yes in my opinion.

db/recovery_db.go
```diff
@@ -109,6 +138,23 @@ func (db *RecoveryDB) PutSwapPrivateKey(id types.Hash, keys *mcrypto.PrivateKeyP
 	return db.db.Put(key[:], val)
 }
 
+// GetSwapPrivateKey returns the swap private key share, if it exists.
```
dimalinux, line 141: "Private key share" is vague given all the keys our code deals with. I assume this is the private spend key of the Monero A+B wallet? Do we want to adopt some kind of consistent terminology for the A+B wallet that we use in the code everywhere?

dimalinux: Looking further down, I think the next function is returning what I thought this function might be returning.  Would it be correct to say that this one is returning "Alice or Bob's half of the A+B wallet spend key"?

protocol/xmrmaker/instance.go
noot: can open another issue for this

ethereum/block/wait_for_receipt.go
noot: functionally these are the same, as cancelling a context sets the error and also closes the `done` channel: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/context/context.go;l=397

not sure which is preferred, imo they're the same so either is ok. if it's not in a select statement I don't really see the need to use `ctx.Done`

common/types/offer.go
noot: I want to say probably not but I'd have to double check how much precision is lost. `RelayerCommission` will generally be on the order of `0.01`, do you think that's acceptable?

db/recovery_db.go
noot: key share generally means one part of a private key, which in this case means one party's "half" of the A+B key.  the "shared" key is the private key of the A+B wallet (a different function)

I've been using "shared" key for the A+B wallet, do you have a preference for clearer naming?

ethereum/block/wait_for_receipt.go
```diff
@@ -16,8 +16,9 @@ import (
 
 const (
 	// in total, we will wait up to 1 hour for a transaction to be included
-	maxRetries           = 360
-	receiptSleepDuration = time.Second * 10
+	maxRetries = 360
+	// TODO: this should vary based on environment, can increase to 12s for mainnet.
+	receiptSleepDuration = time.Second * 2
```
dimalinux, line 21: You could probably just leave it, unless the concern is that we'll generate too many queries to the ethereum endpoint. Polling every 2 seconds won't generate any kind of load on the swapd host. If you wanted to do it *without* the environment value, you could take the delta between the last two blocks. The code that is using it doesn't know how many seconds have passed since the last block was generated before calling sleep, so we probably don't want to wait the full inter-block delta anyways.

noot: I changed it because it was causing some unit tests to fail, since I didn't want a unit test waiting 10+ seconds for a receipt when it could have gotten it much faster

db/recovery_db.go
noot: just saw your second comment, yes exactly

common/types/offer.go
dimalinux: As long as we are not taking a hash of the result anywhere, a slight difference in the binary representation probably doesn't matter. Here's the example I was seeing somewhere else in our code with an exchange rate of 0.05 being reprinted out. (Not a binary difference, but a string difference in that example.)
```
func TestFloat(t *testing.T) {
	jsonData := []byte("0.05")
	var f float64
	err := json.Unmarshal(jsonData, &f)
	require.NoError(t, err)
	t.Logf("%.20f", f)
}
```
Output:
```
0.05000000000000000278
```

db/recovery_db.go
dimalinux: So let's put the word "Share" in this function. `GetSwapPrivateKeyShare`? Right now the comment says "the swap private key share" implying that there is only one, but there are 2 and this key is our half if I understand correctly.  The next function is named `GetSharedSwapPrivateKey`. Maybe we can remove the word "Shared" from it so no one can confuse it with the word "Share". I don't have a suggestion, as I still don't know what it is storing without digging through lots of code. Is it storing the private view key on Bob's side and the private spend key on Alice's side?

ethereum/block/wait_for_receipt.go
dimalinux: I understand.  When I wrote "leave it" above, I was referring to the TODO to change the code yet again in the future. The code change you made is a good one.

protocol/swap/manager.go
```diff
@@ -138,6 +145,19 @@ func (m *manager) GetOngoingSwap(id types.Hash) (*Info, error) {
 	return s, nil
 }
 
+// GetOngoingSwaps returns all ongoing swaps.
+func (m *manager) GetOngoingSwaps() ([]*Info, error) {
+	m.RLock()
+	defer m.RUnlock()
```
dimalinux, line 151: Is it safe that we return pointers to the `Info` structs both here and in `GetOngoingSwap`? We don't have any synchronization primitives inside the `Info` struct and its status field gets changed over time. Is there any risk of 2 go routines accessing the same `Info` structure?  If one go routine modifies the status and the other is reading the status, the second one may not see the changes in a timely manner. My question is not related to the locking here on the `ongoing` map, but looking at the locks made me think of it.

protocol/xmrmaker/instance.go
```diff
@@ -57,15 +60,126 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
+		return err
+	}
+
+	for _, s := range swaps {
+		if s.Provides != types.ProvidesXMR {
+			continue
+		}
+
+		if s.Status == types.KeysExchanged || s.Status == types.ExpectingKeys {
+			// TODO: set status to aborted, delete info from recovery db
```
dimalinux, line 93: Is that a lot of code? Any reason to not put it in this PR?

protocol/xmrmaker/instance.go
```diff
@@ -57,15 +60,126 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
+		}
+	}
+
+	return nil
+}
+
+func (b *Instance) createOngoingSwap(s *swap.Info) error {
+	// check if we have shared secret key in db; if so, recover XMR from that
+	// otherwise, create new swap state from recovery info
+	moneroStartHeight, err := b.backend.RecoveryDB().GetMoneroStartHeight(s.ID)
```
dimalinux, line 109: I was assuming we were going to place the Monero start-scanning height in the `swap.Info` struct. Was there an advantage to storing it separately?

dimalinux: I guess the same question applies to the relayer further down.

protocol/xmrmaker/instance.go
```diff
@@ -57,15 +60,126 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
+
+	relayerInfo, err := b.backend.RecoveryDB().GetSwapRelayerInfo(s.ID)
+	if err != nil {
+		// we can ignore the error; if the key doesn't exist,
+		// then no relayer was set for this swap.
+		relayerInfo = &types.OfferExtra{}
+	}
+
+	b.swapMu.Lock()
+	defer b.swapMu.Unlock()
```
dimalinux, line 158: Do we need to hold the lock when calling `newSwapStateFromOngoing`? If we only need the lock for one line below, I would just write:
```
b.swapMu.Lock()
b.swapStates[s.ID] = ss
b.swapMu.Unlock()
```
That would be the most self documenting about what requires synchronization if, in fact, it is the only line that needs synchronization.

protocol/xmrmaker/instance.go
```diff
@@ -57,15 +60,126 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
+	ss, err := newSwapStateFromOngoing(
+		b.backend,
+		offer,
+		relayerInfo, // TODO: store relayer info in db also
```
dimalinux, line 162: Is this TODO still relevant? I thought we just read it from the DB on line 150.

protocol/xmrmaker/instance.go
```diff
@@ -57,15 +60,126 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
+
+	err = inst.checkForOngoingSwaps()
+	if err != nil {
+		return nil, err
+	}
+
+	return inst, nil
+}
+
+func (b *Instance) checkForOngoingSwaps() error {
```
dimalinux, line 81: I get confused by the letter `b` which normally stands for `backend` in our code. Did you consider calling the receiver here `inst` or maybe calling the struct `Maker` with a receiver named `m`? Just throwing out ideas. We don't have to make them in this PR.

ethereum/block/wait_for_receipt.go
noot: oh woops, sorry I misunderstood lol. cool, can remove the TODO then

protocol/xmrmaker/instance.go
noot: yeah, I actually ended up changing this in a follow up PR to this, it was bothering me too lol

common/types/offer.go
noot: good point, this isn't getting hashed anywhere so I think it should be okay. but definitely something to keep in mind

protocol/swap/manager.go
noot: I want to say this is *probably* okay since `GetOngoingSwaps` is only called when the instances start up, and by the RPC server. but I'll change it to be safe

protocol/xmrmaker/instance.go
noot: I forgot lol, will add!

protocol/xmrmaker/instance.go
noot: when I wrote this I thought there was a reason we might have needed it during `newSwapStateFromOngoing` but re-reading it, don't think we actually do. updated it!

protocol/xmrmaker/instance.go
noot: honestly good question, I guess in my head I was thinking that recovery-specifc info would be stored separately, but I guess start heights/relayer info would make sense as part of `*swap.Info` also, since they're constant anyways.

noot: ah, I guess I didn't add the relayer info since it's specific to the xmrmaker. will leave that as-is but add the start heights to `*swap.Info`

db/recovery_db.go
noot: I think `GetSharedSwapPrivateKey` is fine as the key is essentially a Diffie-Hellman shared key. the only other term that might be okay is "common key"? 

I could also update the comment to say "our swap private key share" to make it clearer?

> Is it storing the private view key on Bob's side and the private spend key on Alice's side?
It's storing the private (spend and view) key on both sides. it corresponds to `swapState.privkeys` for both

protocol/xmrmaker/instance.go
noot: #237

crypto/monero/crypto.go
```diff
@@ -57,6 +57,22 @@ func NewPrivateKeyPairFromBytes(skBytes, vkBytes []byte) (*PrivateKeyPair, error
 	}, nil
 }
 
+// NewPrivateKeyPairFromHex returns a PrivateKeyPair from the given hex-encoded byte
+// representation of a private spend and view key.
+func NewPrivateKeyPairFromHex(skHex, vkHex string) (*PrivateKeyPair, error) {
```
dimalinux, line 62: Why are we passing the view key when it is [derived from the spend key](https://github.com/AthanorLabs/atomic-swap/blob/master/crypto/monero/crypto.go#L137-L161)?

If we wanted it as a sanity check of the spend key, it could make sense, but I don't see the code doing this. Can we simplify? We only need to store the private spend key to the database.

noot: updated to only store spend keys in the db!

dimalinux (approved):

PR #236, opened 2022-11-20
first stab at adding package comments everywhere

dimalinux:
Version v1.50.1 of `golangci-lint` requires package comments. While we could probably figure out how to disable this, I think package comments will improve the code. This PR is just for the package comments. It does not upgrade the linter.

Feel free to make changes directly to the branch. The online editor might be handy. If you break something, I'll fix it.


.golangci.yml
```diff
@@ -23,7 +23,6 @@ linters-settings:
     # Correct spellings using locale preferences for US or UK.
     # Default is to use a neutral variety of English.
     # Setting locale to US will correct the British spelling of 'colour' to 'color'.
-    locale: UK
```
dimalinux, line 26: Hope that is okay. Having to rewrite every comment, changing words to end in "ise", was slowly killing me. :)

cliutil/utils.go
```diff
@@ -1,3 +1,6 @@
+// Package cliutil provides utility functions shared by multiple main packages.
+// TODO: createAndWriteEthKeyFile and GetEthereumPrivateKey are only used by swapd,
+// so we should move them out of here.
```
dimalinux, line 3: I think these were here, because the recoverer called `GetEthereumPrivateKey`. Now that it is gone, we could place them in swapd's main package. Or if you think they'll be reused, I'll just delete the TODO.

dleq/dleq.go
```diff
@@ -1,3 +1,6 @@
+// Package dleq provides and interface named Interface layer on top of the cgo-dleq layer
+// that in turn provides a golang interface to the Distributed Log Equality (DLEQ)
+// algorithm code written in Rust.
```
dimalinux, line 3: I definitely need help on this one.

ethereum/contracts/SwapFactory.sol
```diff
@@ -19,9 +19,9 @@ contract SwapFactory is ERC2771Context, Secp256k1 {
     }
 
     struct Swap {
-        // contract creator, Alice
+        // individual swap creator, Alice
         address payable owner;
```
dimalinux, line 23: Not related to package comments. I was just looking at the contract to better understand the relayer and the comment above looked wrong.

net/message/message.go
```diff
@@ -1,3 +1,4 @@
+// Package message provides the types for messages that are sent between swapd instances.
```
dimalinux, line 1: That comment probably isn't specific enough.

protocol/common.go
```diff
@@ -1,3 +1,7 @@
+// Package protocol has no obvious theme and doesn't have much to do with a protocol. The
+// EthereumAssetAmount seems like it belongs with the coins types in the common package.
+// KeysAndProof seems like it belongs in some crypto package. ConvertContractSwapToMsg
+// seems like it belongs in the net/message package.
```
dimalinux, line 4: I tried to figure out what the theme was, but my thought was that it might be better to just reorganize the code. Let me know what you think. I could be wrong.

rpc/server.go
```diff
@@ -1,3 +1,6 @@
+// Package rpc provides the HTTP server for for incoming JSON-RPC and websocket requests
+// to swapd from the local host. It is also providing stuff related to the libp2p
+// interactions. ... and types for both ... Need help describing the common theme here.
```
dimalinux, line 3: Need help with this one.

common/rpctypes/types.go
```diff
@@ -1,3 +1,7 @@
+// Package rpctypes provides the serialized types for queries and responses shared by
+// swapd's JSON-RPC server and client-side libraries. It also contains a PostRPC method
+// that should maybe be moved to the rpc package? (And probably should be using
+// `github.com/gorilla/rpc/v2/json2` for much of what it does?)
```
dimalinux, line 4: I'll move it to `rpcclient`.

common/types/status.go
```diff
@@ -1,3 +1,4 @@
+// Package types is for types that are shared by multiple packages
```
dimalinux, line 1: Move types in `common` to here.

crypto/hash.go
```diff
@@ -1,3 +1,6 @@
+// Package crypto is for an ethereum hash (Keccak256) wrapper that returns a 32-byte array
```
dimalinux, line 1: Package crypto is for cryptographic code used by both Monero and Ethereum. Chain specific crypto is in subpackages.

dleq/dleq.go
dimalinux: // Package dleq provides the sub-api that we need from the cgo-dleq package, which in turn is a wrapper to the Discrete Log Equality (DLEQ) algorithm code written in Rust.

ethereum/contracts/SwapFactory.sol
```diff
@@ -19,9 +19,9 @@ contract SwapFactory is ERC2771Context, Secp256k1 {
...
-        // address allowed to claim the ether in this contract
+        // address allowed to claim the swap's ether (Bob's address or Forwarder contract)
```
dimalinux, line 24: Revert to original comment

protocol/common.go
dimalinux: Package protocol has functions that are used by both the maker and taker during execution of the swap.

protocol/txsender/sender.go
```diff
@@ -1,3 +1,7 @@
+// Package txsender provides a common Sender interface for swapd instances using an
+// ethereum key that is directly managed by swapd (`privateKeySender`) as well as an
+// external sender (`ExternalSender`), where private key management and transaction
+// signing is done via an external entity like Metamask.
```
dimalinux, line 4: Each Sender implementation signs and submits the transaction to the network.

rpc/server.go
dimalinux: Package rpc provides the HTTP server for for incoming JSON-RPC and websocket requests to swapd from the local host. The answers to these queries come from 3 subsystems: net, personal and swap.






 It is also providing stuff related to the libp2p
interactions. ... and types for both ... Need help describing the common theme here.

common/types/status.go
dimalinux: A lot of those types have functions associated with them. I'm not going to push the scope of this PR with those changes.

dleq/dleq.go
```diff
@@ -1,3 +1,4 @@
+// Package dleq provides the sub-api that we need from the go-dleq package.
```
dimalinux, line 1: "sub-api" is too vague. It should be describing exactly what the functionality provided to the atomic swap protocol is from a high level. I'll work with you on this unless you have a suggestion off the top of your head.

dleq/dleq.go
```diff
@@ -24,14 +25,6 @@ func NewProofWithoutSecret(p []byte) *Proof {
 	}
 }
 
-// NewProofWithSecret returns a new Proof with the given secret.
-// Note that the returned proof actually lacks the `proof` field.
-func NewProofWithSecret(s [32]byte) *Proof {
-	return &Proof{
-		secret: s,
-	}
-}
```
dimalinux, line 33: No one was calling this.  I can put it back if we were planning on using it.

common/rpctypes/types.go
dimalinux: Done in separate PR: https://github.com/AthanorLabs/atomic-swap/pull/241

noot (approved):
looks good!

PR #238, opened 2022-11-21
implement auto-recover for xmrtaker 

### LINKED ISSUE #212: auto-recover on startup if incomplete swaps are detected

noot:
- similar to previous PR, implement checking for ongoing swaps on startup for the xmrtaker instance and restart them if they exist.

closes #212


protocol/xmrtaker/claim.go
```diff
@@ -0,0 +1,157 @@
...
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+	"github.com/athanorlabs/atomic-swap/monero"
+
+	eth "github.com/ethereum/go-ethereum"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+)
+
+func (s *swapState) tryClaim() error {
+	if !s.info.Status.IsOngoing() {
```
noot, line 17: the functions in this file haven't been changed, i just moved them out of swap_state.go since that file was pretty large

db/recovery_db.go
```diff
@@ -98,7 +98,7 @@ func (db *RecoveryDB) GetContractSwapInfo(id types.Hash) (*EthereumSwapInfo, err
 }
 
 // PutSwapPrivateKey stores the given ephemeral swap private key share for the given swap ID.
-func (db *RecoveryDB) PutSwapPrivateKey(id types.Hash, sk *mcrypto.PrivateSpendKey, env common.Environment) error {
+func (db *RecoveryDB) PutSwapPrivateKey(id types.Hash, sk *mcrypto.PrivateSpendKey) error {
 	val, err := json.Marshal(sk.Hex())
 	if err != nil {
 		return err
```
dimalinux, line 104: I added some code [in this PR](https://github.com/AthanorLabs/atomic-swap/pull/239) that will let you marshal/unmarshal `sk` directly. Calling `json.Marshal(sk.Hex())` is equivalent to `[]byte(fmt.Sprintf("%q", sk.Hex()))` if you don't want to check the error. Just pointing that out, not saying it should be one way or the other.

db/recovery_db.go
```diff
@@ -157,12 +156,61 @@ func (db *RecoveryDB) GetSharedSwapPrivateKey(id types.Hash) (*mcrypto.PrivateSp
...
+		return nil, nil, err
+	}
+
+	return sk, vk, nil
+}
+
 // DeleteSwap deletes all recovery info from the db for the given swap.
 func (db *RecoveryDB) DeleteSwap(id types.Hash) error {
 	keys := [][]byte{
+		getRecoveryDBKey(id, relayerInfoPrefix),
```
dimalinux, line 209: I'm not opposed to leaving the Recovery information separate from the Info information, but it feels like we can end up with dangling recovery entries (recovery entries that point to a Info entry that doesn't exist). And we'll never delete those entries if they are dangling.  Can that happen?

protocol/xmrmaker/event.go
```diff
@@ -237,9 +237,10 @@ func (s *swapState) handleEventContractReady() error {
 		log.Warnf("failed to claim funds from contract, attempting to safely exit: %s", err)
 
 		// TODO: retry claim, depending on error (#162)
-		if err = s.exit(); err != nil {
+		if err := s.exit(); err != nil { //nolint:govet
```
dimalinux, line 240: I know it's uglier, but I would use another name or save the outer error in a variable. Intentionally using shadowing is too subtle. There is a good chance the next person to modify the code, not realizing its delicateness, will reintroduce the bug.

protocol/xmrmaker/swap_state.go
```diff
@@ -133,12 +133,14 @@ func newSwapStateFromStart(
...
-	if err := b.SwapManager().AddSwap(info); err != nil {
+	if err = b.SwapManager().AddSwap(info); err != nil {
 		return nil, err
 	}
 
-	offerExtra.StatusCh <- stage
-	return newSwapState(
+	defer func() {
+		offerExtra.StatusCh <- stage
+	}()
```
dimalinux, line 142: Should this be at the end of the function after we've cleared all the error scenarios, or is it okay here?

protocol/xmrtaker/swap_state.go
```diff
@@ -154,13 +123,100 @@ func newSwapState(
...
+		moneroStartNumber,
 		statusCh,
 	)
 	if err = b.SwapManager().AddSwap(info); err != nil {
 		return nil, err
 	}
 
+	defer func() {
+		statusCh <- stage
+	}()
```
dimalinux, line 135: Do we want to execute this even if `newSwapState` below returns an error?

protocol/xmrtaker/swap_state.go
```diff
@@ -154,13 +123,100 @@ func newSwapState(
...
+		info,
+		ethSwapInfo.StartNumber,
+		info.MoneroStartHeight,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if b.ContractAddr() != ethSwapInfo.ContractAddress {
+		return nil, errContractAddrMismatch(ethSwapInfo.ContractAddress.String())
```
dimalinux, line 173: In the longer term, who is going to be specifying the contract address? Will it be Bob (the maker), so that he can control what trusted forwarder is used? Forgetting about what is appropriate for this pull request, I'm wondering if this is a scenario that we'll need to be able to handle later on.

protocol/xmrtaker/swap_state.go
```diff
@@ -154,13 +123,100 @@ func newSwapState(
...
+	b backend.Backend,
+	transferBack bool,
+	info *pswap.Info,
+	ethStartNumber *big.Int,
+	moneroStartNumber uint64,
+) (*swapState, error) {
+	_, err := b.XMRDepositAddress(nil)
+	if transferBack && err != nil {
+		return nil, errMustProvideWalletAddress
+	}
```
dimalinux, line 196: In theory this shouldn't be possible.  We automatically create a Monero wallet if it doesn't exist.  Transfer back is the default. We talked about removing alternate deposit addresses as complexity that might do more damage than good (since money can be lost if the address is wrong), but there was concern that a UI might let the user set an alternate deposit address that would only work as a taker and not as a maker. I would have nuked the `XMRDepositAddress` function, but I didn't want to force a rushed decision. Now that some time has passed, what do you think: Is the use case of an alternate XMR deposit address desirable enough to justify the extra code and non-orthogonal maker/taker Monero handling?

protocol/xmrtaker/swap_state.go
```diff
@@ -154,13 +123,100 @@ func newSwapState(
...
+
+	if !b.ETHClient().HasPrivateKey() {
+		transferBack = true // front-end must set final deposit address
+	}
```
dimalinux, line 200: This is complicated enough to warrant a comment above the block. I had to think about it for a while:
```
// If the user specified `--external-signer=true` (no private eth key in the client) and
// explicitly set `--transfer-back=false` (overriding the default behavior), we override
// their decision and set it back to `true`. ... [explanation of why here] ...
```

db/recovery_db.go
noot: I don't think can happen as `swap.Info` isn't ever deleted from the database, only recovery info is.

protocol/xmrtaker/swap_state.go
noot: so far it's Alice who controls what contract is used, since she sends the address when notifying Bob of `newSwap` being executed. 

in terms of the trusted forwarder, Bob will need to verify the forwarder bytecode if he decides to use a relayer

protocol/xmrtaker/swap_state.go
noot: I think for the UI, users will need to specify the address for the monero to go to - is there currently an option at the moment for them to do that via CLI? I don't think there is, other than if they make the wallet beforehand and point to that.

protocol/xmrtaker/swap_state.go
noot: updated so it executes only after all err checks

protocol/xmrmaker/event.go
noot: updated

db/recovery_db.go
```diff
@@ -157,12 +156,61 @@ func (db *RecoveryDB) GetSharedSwapPrivateKey(id types.Hash) (*mcrypto.PrivateSp
 	return mcrypto.NewPrivateSpendKeyFromHex(skHex)
 }
 
+type xmrmakerKeys struct {
+	PublicSpendKey string
+	PrivateViewKey string
```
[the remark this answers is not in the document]
noot: updated

dimalinux (approved):

protocol/xmrtaker/swap_state.go
dimalinux: No, is it a flag that needs to be added to `swapcli take ...`?

PR #239, opened 2022-11-22
use standard wallet computation of the view key

dimalinux:
While I don't think that our existing view key calculation is insecure, it is not the standard version that most wallets use. This PR puts us on the common implementation and provides tests to make sure we match up with `monero-wallet-rpc` generated wallets.

Apparently, you can even generate a random view key, but it wouldn't be compatible with other wallet implementations. This is from page 37 of "Zero To Monero 2nd edition":
image

This is from page 25:
image

Other notes on the bits that Monero decided not to clear are here:
https://github.com/FiloSottile/edwards25519/blob/v1.0.0/scalar.go#L136-L142


crypto/monero/crypto_test.go
```diff
@@ -96,3 +98,65 @@ func TestNewPublicKeyPairFromHex(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		skBytes, err := hex.DecodeString(tt.spendKey)
+		require.NoError(t, err)
+		skScalar, err := ed25519.NewScalar().SetCanonicalBytes(skBytes)
+		require.NoError(t, err)
+		psk := &PrivateSpendKey{
+			key: skScalar,
+		}
```
noot, line 157:
```suggestion
		psk, err := NewPrivateKeyFromHex(tt.spendKey)
		require.NoError(t, err)
```

when you merge in master :)

crypto/monero/sc_reduce.go
```diff
@@ -0,0 +1,359 @@
+package mcrypto
+
+// ScReduce32 is the same as scReduce, but it takes a 32 byte input that we widen first
+func ScReduce32(s [32]byte) [32]byte {
```
noot, line 4: don't export this?

noot (commented):
nice catch!!

crypto/monero/crypto.go
```diff
@@ -187,6 +189,34 @@ func (k *PrivateSpendKey) Bytes() []byte {
...
+	keyBytes, err := hex.DecodeString(inputStr)
+	if err != nil {
+		return err
+	}
+	if len(keyBytes) != privateKeySize {
+		return errInvalidInput
+	}
+	k.key, err = ed25519.NewScalar().SetCanonicalBytes(keyBytes)
+	return err
+}
```
dimalinux, line 218: I thought this might making the marshalling of the PrivateSpendKey to keystore cleaner. Let me know.

crypto/monero/crypto.go
```diff
@@ -322,10 +352,7 @@ func GenerateKeys() (*PrivateKeyPair, error) {
 		return nil, fmt.Errorf("failed to set bytes: %w", err)
 	}
 
-	sk := &PrivateSpendKey{
-		seed: seed,
```
dimalinux, line 326: Wasn't used after the unused signing code was removed.

crypto/monero/crypto_test.go
dimalinux: I was thinking this PR might be worth merging before yours.  :)

crypto/monero/sc_reduce.go
```diff
@@ -0,0 +1,359 @@
...
+//
+//	s[0]+256*s[1]+...+256^63*s[63] = s
+//
+// Output:
+//
+//	s[0]+256*s[1]+...+256^31*s[31] = s mod l
+//	where l = 2^252 + 27742317777372353535851937790883648493.
+//
+//nolint:ineffassign
+func scReduce(out *[32]byte, s *[64]byte) {
```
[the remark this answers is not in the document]
dimalinux: Done

noot (approved):

PR #241, opened 2022-11-28
Cleanup RPC posting

dimalinux:
This was too big for the package comments PR, so I separated it out:

* Moves RPC posting out of the `rpctypes` package making it a member function of the RPC client type
* Does proper context handling
* Handles JSON serialization/deserialization in the POST method, simplifying code for callers
* Takes advantage of the Gorilla JSON-RPC 2.0 client library which simplifies error handling
* Updates HTTP transport to use DialContext in place of the deprecated Dial


common/rpctypes/utils.go
```diff
@@ -1,70 +1 @@
...
-		}).Dial,
-	}
-	httpClient = &http.Client{
-		Transport: transport,
-		Timeout:   httpClientTimeout,
-	}
-)
-
-// PostRPC posts a JSON-RPC call to the given endpoint.
-func PostRPC(endpoint, method, params string) (*Response, error) {
```
dimalinux, line 32: The new version is in `rpcclient/client.go`.

common/rpctypes/utils.go
```diff
@@ -1,70 +1 @@
...
-		return nil, err
-	}
-
-	r, err := http.NewRequest("POST", endpoint, buf)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
-	}
-	r.Header.Set("Content-Type", contentTypeJSON)
-
-	ctx, cancel := context.WithTimeout(context.Background(), callTimeout)
```
dimalinux, line 46: The old version had a base context that wasn't using the existing context of the client object that was invoking it.

rpcclient/client.go
```diff
@@ -1,13 +1,85 @@
...
+// is used for the full lifetime of the client.
+func NewClient(ctx context.Context, endpoint string) *Client {
 	return &Client{
+		ctx:      ctx,
 		endpoint: endpoint,
 	}
 }
+
+// Post makes a JSON-RPC call to the client's endpoint, serialising any passed request
+// object and deserializing any passed response object from the POST response body. Nil
```
dimalinux, line 50: That is so weird. The linter forced me to rewrite "serializing" with an "s", but it is allowing "deserializing".  I will harmoniZe the spelling of these two words after the linter update makes it to master.  :)

rpcclient/swap.go
```diff
@@ -41,25 +31,11 @@ func (c *Client) GetOngoingSwap(id string) (*rpc.GetOngoingResponse, error) {
...
-	if err != nil {
-		return nil, err
-	}
-
-	resp, err := rpctypes.PostRPC(c.endpoint, method, string(params))
-	if err != nil {
-		return nil, err
-	}
-
-	if resp.Error != nil {
```
dimalinux, line 54: The JSON-RPC wrapper around the real request/response is all handled by the Gorilla library now.

tests/integration_test.go
```diff
@@ -60,12 +60,13 @@ func (s *IntegrationTestSuite) SetupTest() {
...
 
-func (s *IntegrationTestSuite) setSwapTimeout(timeoutSeconds uint64) {
-	ac := rpcclient.NewClient(defaultXMRTakerSwapdEndpoint)
-	err := ac.SetSwapTimeout(timeoutSeconds)
+	// Reset XMR Maker and Taker between tests, so tests starts in a known state
+	ac := rpcclient.NewClient(context.Background(), defaultXMRTakerSwapdEndpoint)
+	err := ac.SetSwapTimeout(defaultSwapTimeout)
+	require.NoError(s.T(), err)
+	bc := rpcclient.NewClient(context.Background(), defaultXMRMakerSwapdEndpoint)
+	err = bc.ClearOffers(nil)
```
dimalinux, line 69: Adding context handling to rpcclient broke a bunch of integration tests, because they were calling `bc.ClearOffers(nil)` in a defer statement that executed after the context that `bc` was created with was cancelled. I removed the cleanup in the individual tests in favor of putting the swapd instances into a know state before each test starts here.

tests/integration_test.go
```diff
@@ -826,7 +799,9 @@ func (s *IntegrationTestSuite) testSuccessConcurrentSwaps(asset types.EthAsset)
 	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
 	defer cancel()
 
-	s.setSwapTimeout(swapTimeout)
+	ac := rpcclient.NewClient(ctx, defaultXMRTakerSwapdEndpoint)
+	err := ac.SetSwapTimeout(swapTimeout)
+	require.NoError(s.T(), err)
```
dimalinux, line 804: I just moved the `ac` declaration up higher. `bc` was already getting created outside of any loop and these clients are stateless from one call to the next. (If the http client has shared state, I think it was getting shared even if you created multiple clients to the same endpoint.) I removed the suite level setting of the swap timeout, so the reader can easily see that it's getting called on the taker and so that it used the shared test context.

tests/integration_test.go
```diff
@@ -758,7 +731,6 @@ func (s *IntegrationTestSuite) testErrorShouldOnlyTakeOfferOnce(asset types.EthA
 
 			if status != types.CompletedSuccess {
 				errCh <- fmt.Errorf("0th swap did not exit successfully: got %s", status)
-				cancel()
```
dimalinux, line 761: I think this `cancel()` is the cause of the race condition we have in `TestError_ShouldOnlyTakeOfferOnce`. We were only calling `cancel()` here on Bob's swapd. The identical code for Charlie isn't calling it. The switch statement starting on line 798 below is handling `errCh` before `ctx.Done()` so I'm surprised that this was an issue. I would have expected the `errCh` case to get handled and the `ctx.Done()` case to be ignored.

rpcclient/client.go
noot: that's painful lol

rpcclient/client.go
```diff
@@ -1,13 +1,85 @@
...
+	ctx, cancel := context.WithTimeout(c.ctx, callTimeout)
+	defer cancel()
+	httpReq = httpReq.WithContext(ctx)
+
+	httpResp, err := httpClient.Do(httpReq)
+	if err != nil {
+		return fmt.Errorf("failed to post %q request: %w", method, err)
+	}
+
+	defer func() { _ = httpResp.Body.Close() }()
```
noot, line 74: super nit but split this into multiple lines? lol

noot (approved):
looks good, nice cleanup!

rpcclient/client.go
dimalinux: My thought was actually the opposite.  If you are creating a function body purely so the linter doesn't complain about an ignored error value, maybe we want it on one line so it more closely resembles `defer xxx.Close()`? Think about it. If you want them on multiple lines, I'll fix it anywhere I did it and not do it in the future. I do see lots of other projects adopting the above style though.

PR #242, opened 2022-11-28
add go-dleq; remove all rust dependencies

noot:
finally finished the go-dleq lib, updated to use it. the project is now pure Go :D

- removes cgo-dleq and fake dleq


Makefile
```diff
@@ -44,37 +35,31 @@ format-solidity:
...
-build-go:
-	GOBIN="$(CURDIR)/bin" go build -tags=fakedleq ./cmd/swapd
-	GOBIN="$(CURDIR)/bin" go build -tags=fakedleq ./cmd/swapcli
-
-# WARNING: this should not be used in production, as the DLEq prover has been stubbed out and now proves nothing.
-.PHONY: build-go-darwin
-build-go-darwin:
-	GOOS=darwin GOARCH=arm64 $(MAKE) build-go
+.PHONY: build-darwin
+build-darwin:
```
dimalinux, line 56: Maybe we should add a comment so that no one uses this target when compiling on macos?  Something like:
```
# Test macos/arm build from linux. Use "make build" if compiling on macos.
```
If we want to provide release binaries at some point for linux/arm, linux/intel, macos/arm, macos/intel ..., I can script something up that will replace this target.

noot: good point, added! and yep, that would be nice to have soon

Makefile
```diff
@@ -89,10 +75,6 @@ mock:
...
 # Deletes all executables matching the directory names in cmd/
-.PHONY: clean-go
-clean-go:
-	rm -f bin/
-
 .PHONY: clean
-clean: clean-go
-	rm -f $(DLEQ_LIB)
+clean:
+	rm -r bin/
```
dimalinux, line 80: You may be able to adjust your editor here.  Mine is configured to ensure that files always have a trailing newline (newline at the end of the last, non-empty line).

dleq/go_dleq.go
```diff
@@ -0,0 +1,76 @@
...
+
+// DefaultDLEq is the default DLEq prover.
+type DefaultDLEq = GoDLEq
+
+// GoDLEq is a wrapper around the go-dleq library prover and verifier.
+type GoDLEq struct{}
+
+var (
+	curveA = secp256k1.NewCurve()
+	curveB = ed25519.NewCurve()
```
dimalinux, line 21: `curveEthereum` and `curveMonero` would probably make the code more self-documenting.

dleq/go_dleq.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// Prove generates a secret and a corresponding proof that it has a value
+// on the secp256k1 and ed25519 curves.
+func (d *GoDLEq) Prove() (*Proof, error) {
+	x, err := dleq.GenerateSecretForCurves(curveA, curveB)
+	if err != nil {
+		panic(err)
```
dimalinux, line 29: Can I assume this is panicking because the called function should never fail? If that's the reason, can we add a comment?

dleq/go_dleq.go
```diff
@@ -0,0 +1,76 @@
...
+type GoDLEq struct{}
+
+var (
+	curveA = secp256k1.NewCurve()
+	curveB = ed25519.NewCurve()
+)
+
+// Prove generates a secret and a corresponding proof that it has a value
+// on the secp256k1 and ed25519 curves.
+func (d *GoDLEq) Prove() (*Proof, error) {
```
dimalinux, line 26: The name doesn't seem to reflect what the function is doing. What would be more self documenting?  `GenSecretAndProof`? For the comment, I don't understand what it means for a secret to have a value on 2 curves. That sounds like a point, but I'm assuming the secret is a scalar that only forms a point on the curve when it is multiplied by the curve's generator (`G`).

dleq/go_dleq.go
noot: updated

dleq/go_dleq.go
noot: i forgot to change it, updated

dleq/go_dleq.go
noot: yes it means it has a corresponding point on both curves when multiplied by the curves's generators. will update

dleq/go_dleq.go
```diff
@@ -0,0 +1,76 @@
...
+
+	dleq "github.com/athanorlabs/go-dleq"
+	"github.com/athanorlabs/go-dleq/ed25519"
+	"github.com/athanorlabs/go-dleq/secp256k1"
+
+	dsecp256k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
+)
+
+// DefaultDLEq is the default DLEq prover.
+type DefaultDLEq = GoDLEq
```
dimalinux, line 14: Should we add, in the comment, that there is only one implementation at the current time? Or is it still easy to point this at the Rust version?

dimalinux (approved):
Great to have this! Make sure to look at all the comments before merging.

dleq/go_dleq.go
noot: yeah I'll add that this is the only implementation at the moment

PR #245, opened 2022-11-29
fix: don't re-advertise if we have no offers

### LINKED ISSUE #160: update network offer advertisement if no offers are available
in net/discovery.go, if we have ever made and advertised offers, we continue to do that indefinitely. fix this so we don't continue to advertise if we have no offers, and update what's advertised if our offer list empties or becomes non empty.

noot:
closes #160


net/discovery.go
```diff
@@ -120,6 +133,11 @@ func (d *discovery) advertise() {
 			d.provides = []types.ProvidesCoin{types.ProvidesXMR}
```
dimalinux, line 133: Github won't let me put the comment on line 119, but that line still has the TODO that this PR is closing.

net/discovery.go
```diff
@@ -120,6 +133,11 @@ func (d *discovery) advertise() {
...
 			doAdvertise()
 		case <-time.After(ttl):
+			offers := d.offerAPI.GetOffers()
+			if len(offers) == 0 {
+				continue
+			}
+
 			doAdvertise()
```
dimalinux, line 141: I suspect the average reader, including myself right now, won't know the finer nuances of libp2p DHT caching.  We may need some comments somewhere in here. It makes sense to not advertise if there are no offers, assuming that we don't need to advertise to clear previous offers. Is this an optimization, or is it changing behavior?

This function would be easier to understand if the long nested function was not nested and if this function was called `advertiseLoop`.

net/discovery.go
noot: removed

net/discovery.go
noot: updated, and added comments explaining things more

dimalinux (approved):
Comments are great! Thanks!

net/discovery.go
```diff
@@ -83,50 +96,65 @@ func (d *discovery) stop() error {
...
-		ttl, err = d.rd.Advertise(d.ctx, "")
-		if err != nil {
-			log.Debugf("failed to advertise in the DHT: err=%s", err)
-			ttl = tryAdvertiseTimeout
-			return
-		}
-
-		ttl = defaultAdvertiseTTL
-	}
+	ttl = d.advertise(ttl)
```
dimalinux, line 102: Removing line 100 and passing the constant here might be slightly more readable.

dimalinux: What does `ttl` stand for? It usually means "time to live" and is used for a countdown until something is discarded. This seems more like a `nextRefresh` duration.

net/discovery.go
```diff
@@ -83,50 +96,65 @@ func (d *discovery) stop() error {
...
 			return
 		}
 	}
 }
 
+// advertise advertises that we provide XMR in the DHT.
+// note: we only advertise that we are an XMR provider, but we don't
+// advertise our specific offers.
+// to find what our offers are, peers need to send us a QueryRequest
+// over the query subprotocol.
```
dimalinux, line 132: Can we mention what the return value is?  It looks like we return the duration the caller should pause before advertising again.

net/discovery.go
```diff
@@ -83,50 +96,65 @@ func (d *discovery) stop() error {
...
+func (d *discovery) advertise(ttl time.Duration) time.Duration {
+	log.Debug("advertising in the DHT...")
+	err := d.dht.Bootstrap(d.ctx)
+	if err != nil {
+		log.Warnf("failed to bootstrap DHT: err=%s", err)
+		return ttl
```
dimalinux, line 138: Is this the value you wanted to return? On other errors you are returning `tryAdvertiseTimeout`.

net/discovery.go
dimalinux: Ohhh... I'm seeing where the name came from. They are abusing that name too:
https://github.com/libp2p/go-libp2p/blob/v0.21.0/p2p/discovery/routing/routing.go#L32-L37
It was probably correctly used on the DHT record, and then some programmer thought it was a good idea to use it on a value that has almost nothing in common with the time of the DHT record's expiration.

net/discovery.go
```diff
@@ -83,50 +96,65 @@ func (d *discovery) stop() error {
...
+
+	for _, provides := range d.provides {
+		ttl, err = d.rd.Advertise(d.ctx, string(provides))
+		if err != nil {
+			log.Debugf("failed to advertise in the DHT: err=%s", err)
+			return tryAdvertiseTimeout
+		}
+	}
+
+	ttl, err = d.rd.Advertise(d.ctx, "")
```
dimalinux, line 149: Why do we advertise once with "XMR" on line 142 and then a second time with ""?

net/discovery.go
noot: yeah, this is the same logic as before, but I updated it

net/discovery.go
noot: however there isn't a loop that tries to discover new peers if our peer count is low, that should probably be implemented

dimalinux: > one says that we're in the network and allows others to discover us no matter what, the other says we're XMR providers specifically.

We only call `advertise` when we have XMR offers.  `swapcli discover` and `swapcli query-all` both take a coin. Is there anyone using this?

PR #246, opened 2022-11-30
fix: add monero public spend key check when verifying DLEq proof

noot:
as title says, this should have been checked before.


protocol/common.go
```diff
@@ -76,5 +82,17 @@ func VerifyKeysAndProof(proofStr, secp256k1PubString string) (*secp256k1.PublicK
 		return nil, err
 	}
 
-	return secp256k1Pub, nil
+	ed25519Pub, err := mcrypto.NewPublicKeyFromHex(ed25519PubString)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate XMRMaker's public spend key: %w", err)
+	}
+
+	if res.Ed25519PublicKey().Hex() != ed25519Pub.Hex() {
```
dimalinux, line 90: Would this be better?  `!bytes.Equal(res.Ed25519PublicKey().Bytes(), ed25519Pub.Bytes())`

I'm always hesitant to compare hex, because there is more than one way to represent the same value with upper/lower case and prefixes. Unless it is a unit test, in which case comparing hex can help, as the non-equal strings are more helpful in the test failure message than binary.

dimalinux (approved):
Look at the comment I left before merging.

protocol/common.go
noot: updated

PR #248, opened 2022-11-30
implement get-offers for swapcli

### LINKED ISSUE #243: implement `swapcli get-offers`
we should be able to list our current offers with swapcli.

noot:
closes #243

### DIFF
--- a/cmd/swapcli/main.go
+++ b/cmd/swapcli/main.go
@@ -247,6 +247,14 @@ var (
 				},
 			},
 			{
+				Name:   "get-offers",
+				Usage:  "Get all current offers.",
+				Action: runGetOffers,
+				Flags: []cli.Flag{
+					swapdPortFlag,
+				},
+			},
+			{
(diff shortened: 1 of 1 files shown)


cmd/swapcli/main.go
```diff
@@ -628,6 +636,21 @@ func runClearOffers(ctx *cli.Context) error {
...
+func runGetOffers(ctx *cli.Context) error {
+	c := newRRPClient(ctx)
+	resp, err := c.GetOffers()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Offers:")
+	for _, offer := range resp {
+		fmt.Printf("\t%v\n", offer)
```
dimalinux, line 648: Is that readable?

dimalinux (approved):

cmd/swapcli/main.go
noot: yep, `offer.String()` formats it

PR #251, opened 2022-12-04
libp2p QUIC and NAT traversal features

dimalinux:
This PR enables additional libp2p NAT traversal features and enables nodes to communicate using UDP/QUIC in addition to TCP. In this PR we also switch to using libp2p's "routed host". Most commands that previously accepted multiaddress values now accept a Peer ID instead.


monero/wallet_client.go
```diff
@@ -591,10 +591,10 @@ func getWalletRPCFlags(conf *WalletClientConf) []string {
 	return args
 }
 
-// getFreePort returns an OS allocated and immediately freed port. There is nothing preventing
+// getFreeTCPPort returns an OS allocated and immediately freed port. There is nothing preventing
 // something else on the system from using the port before the caller has a chance, but OS
 // allocated ports are randomised to make the risk negligible.
-func getFreePort() (uint, error) {
+func getFreeTCPPort() (uint, error) {
```
dimalinux, line 597: Since the code now uses 2 different port types, I wanted to preemptively avoid confusion.

net/host.go
```diff
@@ -328,49 +327,36 @@ func (h *host) writeToStream(s libp2pnetwork.Stream, msg Message) error {
 	return nil
 }
 
-// readStream reads from the stream into the given buffer, returning the number of bytes read
-func readStream(stream libp2pnetwork.Stream, buf []byte) (int, error) {
-	if stream == nil {
-		return 0, errNilStream
+// readStreamMessage reads the 4-byte LE size header and message body returning the
+// message body bytes.
+func readStreamMessage(s libp2pnetwork.Stream) ([]byte, error) {
```
dimalinux, line 332: I was running into bugs in this function. I chose to simplify it instead of debug it. We use a fixed, 4-byte size (instead of variable-length 16-byte size) and we allocate the buffers in here instead of passing them it=b. The majority of the callers were dynamically allocating the maximum size buffer before every call, so the new solution is most likely less resource intensive.

net/query_test.go
```diff
@@ -3,16 +3,16 @@ package net
...
-	"github.com/athanorlabs/atomic-swap/common/types"
-
 	"github.com/stretchr/testify/require"
+
+	"github.com/athanorlabs/atomic-swap/common/types"
 )
 
 func TestHost_Query(t *testing.T) {
-	ha := newHost(t, defaultPort)
+	ha := newHost(t, 0) // OS assigned port
```
dimalinux, line 12: Avoiding port in use failures

net/host.go
```diff
@@ -121,38 +112,46 @@ func NewHost(cfg *Config) (*host, error) {
 		return nil, err
 	}
 
+	// format bootnodes
+	bns, err := stringsToAddrInfos(cfg.Bootnodes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to format bootnodes: %w", err)
+	}
```
dimalinux, line 119: This was just moved from further down.

cmd/swapcli/main.go
```diff
@@ -18,7 +18,6 @@ import (
 )
 
 const (
-	defaultSwapdPort              = 5001
```
dimalinux, line 21: `swapcli` was defaulting to 5001 and `swapd` was defaulting to 5005. I changed the default for both to 5000.

cmd/swapd/main_test.go
```diff
@@ -68,12 +68,15 @@ func TestDaemon_DevXMRTaker(t *testing.T) {
 		"test --dev-xmrtaker",
 		map[string]any{
 			flagEnv:         "dev",
+			flagDeploy:      true,
 			flagDevXMRTaker: true,
 			flagDataDir:     t.TempDir(),
+			flagRPCPort:     uint(0),
+			flagLibp2pPort:  uint(0),
```
dimalinux, line 75: The above 2 lines and similar ones below are making the tests use random ports.

common/rpctypes/types.go
```diff
@@ -80,36 +86,47 @@ type MakeOfferRequest struct {
...
+	PiconeroUnlockedBalance uint64   `json:"piconeroUnlockedBalance"`
+	BlocksToUnlock          uint64   `json:"blocksToUnlock"`
+	EthAddress              string   `json:"ethAddress"`
+	WeiBalance              *big.Int `json:"weiBalance"`
+}
+
+// AddressesResponse ...
+type AddressesResponse struct {
+	Addrs []string `json:"addresses"`
+}
```
dimalinux, line 127: This was moved, unchanged, from `rpc/net.go` so we have all the RPC types in one place.

common/types/offer.go
```diff
@@ -35,23 +35,23 @@ func NewOffer(coin ProvidesCoin, minAmount float64, maxAmount float64, exRate Ex
...
+		MaxAmount:    maxAmount,
+		ExchangeRate: exRate,
+		EthAsset:     ethAsset,
 	}
 }
 
 // String ...
 func (o *Offer) String() string {
-	return fmt.Sprintf("Offer ID=%s Provides=%v MinimumAmount=%v MaximumAmount=%v ExchangeRate=%v EthAsset=%v",
+	return fmt.Sprintf("OfferID:%s Provides:%v MinAmount:%v MaxAmount:%v ExchangeRate:%v EthAsset:%v",
```
dimalinux, line 50: Colons get treated as a word divider when double clicking.

common/types/offer_test.go
```diff
@@ -16,7 +16,7 @@ func TestOffer_MarshalJSON(t *testing.T) {
 
 	expected := fmt.Sprintf(`{
 		"version": "0.1.0",
-		"offerId": "%s",
+		"offerID": "%s",
```
dimalinux, line 19: I actually think the original way is prettier, but I was shooting for consistency over beauty.

docs/local.md
```diff
@@ -31,97 +42,145 @@ make build
...
-./swapcli make --min-amount 0.1 --max-amount 1 --exchange-rate 0.05 --swapd-port 5002
-# Published offer with ID cf4bf01a0775a0d13fa41b14516e4b89034300707a1754e0d99b65f6cb6fffb9
+./bin/swapcli make --min-amount 0.1 --max-amount 1 --exchange-rate 0.05 --swapd-port 5001
+```
+Example output:
+```
+Published:
+	Offer ID:  0x09dd41c7b8620cdc3716463dc947a11edf3af45ff07c8b0ff89dd23592e732ca
+	Peer ID:   12D3KooWK7989g6xmAaEsKFPuZTj2CVknRxQuk7dFL55CC1rpEWW
+	Taker Min: 0.005000000000000001 ETH
```
dimalinux, line 112: I'll fix this to what it will look like after we switch to using big.Float and using a precision that is customized for either Monero or Ethereum.

docs/rpc.md
```diff
@@ -17,34 +20,66 @@ Returns:
...
 ### `net_discover`
 
 Discover peers on the network via DHT that have active swap offers.
 
 Parameters:
-- `provides` (optional): one of `ETH` or `XMR`, depending on which offer you are searching for. **Note**: Currently only `XMR` offers are supported. Default is `XMR`.
-- `searchTime` (optional): duration in seconds for which to perform the search. Default is 12s.
+- `provides` (optional): one of `ETH` or `XMR`, depending on which offer you are searching
+  for. **Note**: Currently only `XMR` offers are supported. Default is `XMR`.
+- `searchTime` (optional): time in seconds to perform the search. Default is 12s.
```
dimalinux, line 49: Some of these blocks didn't change.  I just auto-formatted them so they didn't scroll off of the right side of my screen.

net/discovery.go
```diff
@@ -27,50 +27,12 @@ type discovery struct {
...
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	default:
-		// not cancelled, continue on
-	}
-
-	dht, err := dual.New(ctx, h, dhtOpts...)
-	if err != nil {
-		return nil, err
-	}
```
dimalinux, line 61: The code in this function was moved to host.go, as the `dht` created above is needed to initialize libp2p's routedhost.

net/filters.go
```diff
@@ -1,40 +0,0 @@
-package net
```
dimalinux, line 1: I'm sure we'll want this code again in the future, at which point we can restore it from the git history.

rpcclient/get_offers.go
```diff
@@ -1,21 +1,20 @@
 package rpcclient
 
 import (
-	"github.com/athanorlabs/atomic-swap/common/types"
 	"github.com/athanorlabs/atomic-swap/rpc"
 )
 
 // GetOffers calls swap_getOffers.
-func (c *Client) GetOffers() ([]*types.Offer, error) {
+func (c *Client) GetOffers() (*rpc.GetOffersResponse, error) {
```
dimalinux, line 8: We are returning the full response type now, because it now includes the local Peer ID. The Peer ID is not included in the request, so there is a good chance that the caller does not have it.

rpcclient/make_offer.go
```diff
@@ -13,14 +13,14 @@ func (c *Client) MakeOffer(
 	ethAsset types.EthAsset,
 	relayerEndpoint string,
 	relayerCommission float64,
-) (string, error) {
+) (*rpctypes.MakeOfferResponse, error) {
```
dimalinux, line 16: Same on this one. We are returning the response now, because it includes both the Peer ID and the Offer ID. We want to return the Peer ID, as it is not specified in the request, so there is a good chance Bob doesn't have it. If Bob wants to pass the swap information on to Alice, he needs to pass both pieces of information for her to accept the offer.

common/types/offer.go
```diff
@@ -19,13 +19,13 @@ var (
...
-	MinimumAmount float64        `json:"minAmount"`
-	MaximumAmount float64        `json:"maxAmount"`
-	ExchangeRate  ExchangeRate   `json:"exchangeRate"`
-	EthAsset      EthAsset       `json:"ethAsset"`
+	Version      semver.Version `json:"version"`
+	ID           Hash           `json:"offerID"`
+	Provides     ProvidesCoin   `json:"provides"`
+	MinAmount    float64        `json:"minAmount"`
+	MaxAmount    float64        `json:"maxAmount"`
+	ExchangeRate ExchangeRate   `json:"exchangeRate"`
```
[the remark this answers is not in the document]
noot: I left it as a float because it's pretty annoying to have to type 12 decimal places lol. big.Float would be okay though

docs/local.md
```diff
@@ -31,97 +42,145 @@ make build
 
 This creates `swapd` and `swapcli` binaries in the `bin` directory at the top of the project.
 
-To run as Alice, execute in terminal 1:
+### Launch Alice and Bob's swapd Instances
+
+To launch Alice's swapd instance, use this command:
 ```bash
-./swapd --dev-xmrtaker --deploy
+./bin/swapd --dev-xmrtaker --deploy &> alice.log &
```
noot, line 49: do we need to redirect to a log file for just local instructions? I think if people want to that they can do it on their own

net/discovery.go
```diff
@@ -180,13 +143,17 @@ func (d *discovery) discoverLoop() {
 	}
 }
 
-func (d *discovery) findPeers(provides string, timeout time.Duration) ([]peer.AddrInfo, error) {
-	peerCh, err := d.rd.FindPeers(d.ctx, provides)
+func (d *discovery) findPeers(provides string, timeout time.Duration) ([]peer.ID, error) {
+	peerCh, err := d.rd.FindPeers(d.ctx, provides,
+		libp2pdiscovery.Limit(defaultMaxPeers),
+		libp2pdiscovery.TTL(defaultAdvertiseTTL),
```
noot, line 149: what does this do?

net/host.go
```diff
@@ -123,30 +129,15 @@ func NewHost(cfg *Config) (*host, error) {
 
 	// set libp2p host options
 	opts := []libp2p.Option{
-		libp2p.ListenAddrs(addr),
-		libp2p.DisableRelay(),
-		libp2p.Identity(key),
+		libp2p.ListenAddrStrings(
+			fmt.Sprintf("/ip4/%s/tcp/%d", listenIP, cfg.Port),
+			fmt.Sprintf("/ip4/%s/udp/%d/quic-v1", listenIP, cfg.Port),
+		), libp2p.Identity(key),
```
noot, line 135:
```suggestion
		), 
		libp2p.Identity(key),
```

net/host.go
```diff
@@ -155,27 +146,56 @@ func NewHost(cfg *Config) (*host, error) {
 		return nil, fmt.Errorf("failed to format bootnodes: %w", err)
 	}
 
+	if len(bns) > 0 {
+		opts = append(opts, libp2p.EnableAutoRelay(autorelay.WithStaticRelays(bns)))
```
noot, line 150: I assume this requires the bootnodes to have relay enabled?

net/host.go
```diff
@@ -123,30 +129,15 @@ func NewHost(cfg *Config) (*host, error) {
 
 	// set libp2p host options
 	opts := []libp2p.Option{
-		libp2p.ListenAddrs(addr),
-		libp2p.DisableRelay(),
-		libp2p.Identity(key),
+		libp2p.ListenAddrStrings(
+			fmt.Sprintf("/ip4/%s/tcp/%d", listenIP, cfg.Port),
+			fmt.Sprintf("/ip4/%s/udp/%d/quic-v1", listenIP, cfg.Port),
```
noot, line 134: does this actually enable both transports? I would expect you'd need to use `libp2p.Transport` https://pkg.go.dev/github.com/libp2p/go-libp2p#Transport

net/host.go
```diff
@@ -155,27 +146,56 @@ func NewHost(cfg *Config) (*host, error) {
...
 	hst := &host{
 		ctx:        ourCtx,
 		cancel:     cancel,
-		protocolID: fmt.Sprintf("%s/%s/%d", protocolID, cfg.Environment, cfg.EthChainID),
-		h:          h,
+		protocolID: fmt.Sprintf("%s/%s/%s/%d", protocolID, protocolVersion, cfg.Environment, cfg.EthChainID),
+		h:          routedHost,
 		ds:         ds,
 		bootnodes:  bns,
-		queryBuf:   make([]byte, 1024*5),
```
noot, line 172: why was this removed?

net/host.go
```diff
@@ -308,86 +334,113 @@ func (h *host) addrInfo() peer.AddrInfo {
 	}
 }
 
-func (h *host) writeToStream(s libp2pnetwork.Stream, msg Message) error {
+func writeStreamMessage(s io.Writer, msg Message, peerID peer.ID) error {
```
noot, line 337: why change stream to `io.Writer` and add the additional peerID param?

net/host.go
```diff
@@ -308,86 +334,113 @@ func (h *host) addrInfo() peer.AddrInfo {
...
-	for i := 0; i < maxReads; i++ {
-		n, err := stream.Read(buf[tot:])
-		if err != nil {
-			return n + tot, err
-		}
-
-		tot += n
-		if tot == int(length) {
-			break
+	msgBuf := make([]byte, msgLen)
```
noot, line 387: imo it's a bad idea to allocate buffers based on a peer opening a stream as it's a DOS vector. we should fix this in a follow up

net/host.go
```diff
@@ -308,86 +334,113 @@ func (h *host) addrInfo() peer.AddrInfo {
...
+	_, err = io.ReadFull(s, msgBuf)
+	if err != nil {
+		if isEOF(err) {
+			err = io.ErrUnexpectedEOF
```
noot, line 391: what's the point of masking the underlying error here? why not wrap the error?

net/host.go
```diff
@@ -308,86 +334,113 @@ func (h *host) addrInfo() peer.AddrInfo {
...
+		log.Debugf("Bootstrapping to peer: %s (%s)", bn, h.h.Network().Connectedness(bn.ID))
+		wg.Add(1)
+		go func(p peer.AddrInfo) {
+			defer wg.Done()
+			err := h.h.Connect(h.ctx, p)
+			if err != nil {
+				log.Debugf("Failed to bootstrap to peer %s: err=%s", p.ID, err)
+				atomic.AddUint64(&failed, 1)
+			}
+			log.Debugf("Bootstrapped connections to: %s", h.h.Network().ConnsToPeer(p.ID))
```
noot, line 438: what does this log actually look like? seems like printing a `Conn` type might be ugly?

net/initiate.go
```diff
@@ -59,7 +60,7 @@ func (h *host) Initiate(who peer.AddrInfo, msg *SendKeysMessage, s common.SwapSt
 		stream:    stream,
 	}
 
-	go h.handleProtocolStreamInner(stream, s, make([]byte, messageBufferSize))
```
noot, line 62: the reason 1 buffer was allocated per peer is because allocations are expensive, on prev projects allocating buffers for reading network messages used a LOT of CPU. should update this in a follow up imo so the network has a buffer pool

net/message/message.go
```diff
@@ -50,35 +50,27 @@ func DecodeMessage(b []byte) (Message, error) {
 	if len(b) == 0 {
 		return nil, errors.New("invalid message bytes")
 	}
+	msgType := Type(b[0])
+	msgJSON := b[1:]
+	var msg Message
```
noot, line 55: check len(b) > 1 instead?

net/initiate.go
```diff
@@ -138,22 +134,19 @@ func (h *host) handleProtocolStreamInner(stream libp2pnetwork.Stream, s SwapStat
...
-		msg, err := message.DecodeMessage(buf[:tot])
-		if err != nil {
-			log.Debug("failed to decode message from peer, id=", stream.ID(), " protocol=", stream.Protocol(), " err=", err)
-			continue
-		}
-
-		log.Debug(
-			"received message from peer, peer=", stream.Conn().RemotePeer(), " type=", msg.Type(),
-		)
+		log.Debugf("received proto=%s message from peer=%s type=%s",
```
noot, line 148:
```suggestion
		log.Debugf("received protocol=%s message from peer=%s type=%s",
```

net/query.go
```diff
@@ -54,22 +53,16 @@ func (h *host) Query(who peer.AddrInfo) (*QueryResponse, error) {
 
 func (h *host) receiveQueryResponse(stream libp2pnetwork.Stream) (*QueryResponse, error) {
 	h.queryMu.Lock()
-	defer h.queryMu.Unlock()
-
-	buf := h.queryBuf
+	defer h.queryMu.Unlock() // TODO: Do we need this?
```
noot, line 56: since you removed the buffer then now, but I think we should have the pre-allocated buffer

net/utils.go
```diff
@@ -93,54 +80,3 @@ func saveKey(priv crypto.PrivKey, fp string) (err error) {
 	}
 	return f.Close()
 }
-
-func uint64ToLEB128(in uint64) []byte {
```
noot, line 97: why did you remove these again? they *should* save a few bytes on most messages, if that matters

rpc/net.go
```diff
@@ -196,17 +169,12 @@ func (s *NetService) takeOffer(multiaddr, offerID string, providesAmount float64
 	skm.OfferID = offerID
 	skm.ProvidedAmount = providesAmount
 
-	if err = s.net.Initiate(who, skm, swapState); err != nil {
+	if err = s.net.Initiate(peer.AddrInfo{ID: who}, skm, swapState); err != nil {
```
noot, line 172: `(peer.AddrInfo{ID: who}`

this works? interesting

noot (commented):
overall looks fine, some minor comments only, will do some local testing as well!

common/types/offer.go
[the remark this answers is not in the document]
dimalinux: Yeah, I want to update all of these to `*big.Float` assuming that the built-in marshaling works well for it.  I'll look into that today.  The underlying type of `ExchangeRate` is `float64` and we don't have any custom marshaling for it, so changing the type here doesn't affect anything. It just puts it in a type that the rest of our code expects.

net/host.go
dimalinux: Yes, it does.  Relay is enabled by default in libp2p.  We were previously disabling it.

net/host.go
dimalinux: Yes. The default transports include websockets, tcp, and 2 versions of QUIC, so it might be more correct to say that this filters out the transports which we don't use.

net/host.go
dimalinux: The peer ID was only used in a log message. I changed the type to make the function unit testable.

net/host.go
dimalinux: There isn't any additional context in the underlying varieties of EOF that we care about. We want to treat them as the same regardless of what type of reader or stream is being used. The important distinction that we care about is: Was the connection closed between messages (frequently not an error), or was it closed in the middle of a message (almost always an error).

net/host.go
dimalinux: It definitely is ugly and maybe we want to remove it long term. On the flip side, to get the less ugly pieces of information, I'd have to do a lot of ugly casting here.
```
Bootstrapped connections to: [<swarm.Conn[*libp2pquic.transport] /ip4/0.0.0.0/udp/9900/quic-v1 (12D3KooWQQWDJ7KA1Fwdf2ejWz9VXHKvY8cC5PB7Sf34fbEGbsgV) <-> /ip4/149.28.254.150/udp/9900/quic-v1 (12D3KooWGHwHJt6uYDC8fJaeGrduDYwjMBG86Lv1XmopnygwJBzN)>]
```

net/utils.go
dimalinux: The existing solution had bugs. I thought it was better to do something simple and obvious that anyone could quickly understand. The maximum message size constant was more than 16 bits, so I went with uint32 for the size. You can think of it as having a 4-byte fixed size message header.

rpc/net.go
dimalinux: Yeah, it works.  libp2p does it in a bunch of places in their code:
https://github.com/libp2p/go-libp2p/blob/master/p2p/host/routed/routed.go#L205

net/discovery.go
dimalinux: I meant to explicitly ask for your feedback on this one.  I added this code when trying to debug a different problem, but this change wasn't what solved the problem. It should work without it, but it may add value.  Here's the documentation on them:
https://github.com/libp2p/go-libp2p/blob/v0.24.0/core/discovery/options.go#L27-L41

net/initiate.go
dimalinux: I was dealing with bugs and I was just shooting for simple.  I'm not opposed to moving back to a fixed size buffer.  For the non DDOS case, we are only allocating the needed message size (verses the max message size above). The go runtime (like libc) pre-allocates pools of smaller sized buffers. The CPU used for the volume of messages we exchange should be so small that you would have difficulty measuring it (verses noise from everything else we do).

docs/local.md
dimalinux: I was trying to simplify the instructions about using multiple terminals. Ideally, people are only looking at the logs to troubleshoot problems. I was assuming the same thing.  The user is free to not redirect the logs if they want them on the terminal.  They'd just cut-and-paste a short part of the line.

net/host.go
dimalinux: The allocations for swap messages and query messages are all in one place now. The buffers are the same size as the messages creating a lot of code simplifications.

net/host.go
dimalinux: I agree that it is a potential DDOS issue.  We don't hold on to the buffers long, and if the message is bad, we close the connection. I didn't think these changes were a degradation from what the code was already doing, as we were already allocating the buffer on every new stream here:
https://github.com/AthanorLabs/atomic-swap/blob/master/net/initiate.go#L62

net/message/message.go
dimalinux: The above code is safe.  We really only need to check that `len(b)` is non-zero before referencing `b[0]`.  If `b[1:]` is empty, that will get caught by `json.Unmarshal`.  The Unmarshal will definitely fail if we don't have at least `{}`.  I'll update it to `< 3`.

net/host.go
noot: yeah, I think this is fine for this PR but should be fixed still. for example a peer could connect to us an open as many `query` streams as possible (since we can have multiple streams open with a peer with the same protocol id), which would result in us going OOM.

net/discovery.go
noot: I'm wondering if this option actually does anything, looking at what `Advertise` does here it doesn't look like it's actually passed to the DHT: https://github.com/libp2p/go-libp2p/blob/v0.24.1/p2p/discovery/routing/routing.go#L25

net/host.go
dimalinux: We should definitely put some time into testing/attempting/preventing DDOS attacks. I'm trying to test your DDOS idea right now, but I can't open more than 256 streams on the same connection. It looks like the max number of streams is configurable per direction:
https://github.com/libp2p/go-libp2p/blob/master/p2p/host/resource-manager/limit.go#L25-L26

net/host.go
dimalinux: I figured out how to clean it up.  It looks better now:
```
Bootstrapped connection to /ip4/164.92.217.101/udp/9900/quic-v1/p2p/12D3KooWQLZjDoEtjuVJs2DFsFbfasdxpJpJ5Q7jVztoDGzUa3Qz
```

net/discovery.go
noot: yes, I think that's good

net/host.go
noot: good to know, maybe just limiting the number of open streams a peer can have would be okay, since we have 50 max peers. if we limit open streams per peer to say, 10, the worst that can happen is they force us to do many large allocations in sequentially (which is still bad, but probably won't cause OOM?)

net/host.go
noot: nice :D

noot (approved):
looks good, only the one very minor comment left. will open an issue so we can look into network buffers/limiting streams per peer after! really nice work :)

net/host.go
dimalinux: I think we'd be fine. The max buffer size right now is "1 << 17" (131072).
```
(50 peers * 10 connections * 131072 bytes) / (1024 * 1024) = 62.5 MB
```
Because the streams are multiplexed over the same connection, and we would only be reading one message at a time, we are probably no longer referencing the previous buffer when reading the next one (i.e. GC can recycle it).  libp2p is limiting how much unread data can be cached on the receiver's stream as well.

net/discovery.go
dimalinux: done

PR #252, opened 2022-12-05
implement discoverLoop to find peers if count is low

### LINKED ISSUE #247: try to discover peers if peer count is low
if our peer count gets low (<5 perhaps?) we should initiate DHT discovery, searching for peers that provide the empty string (as that's what currently used to advertise that we're in the network).

noot:
closes #247

### DIFF
--- a/net/discovery.go
+++ b/net/discovery.go
@@ -18,6 +18,7 @@ import (
 const (
 	tryAdvertiseTimeout = time.Second * 30
 	defaultAdvertiseTTL = time.Minute * 5
+	defaultMinPeers     = 3  // TODO: make this configurable
 	defaultMaxPeers     = 50 // TODO: make this configurable
 )
 
@@ -86,6 +87,7 @@ func (d *discovery) start() error {
 	// wait to connect to bootstrap peers
 	time.Sleep(time.Second)
 	go d.advertiseLoop()
+	go d.discoverLoop()
 
 	log.Debug("discovery started!")
 	return nil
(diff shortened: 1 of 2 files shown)


net/discovery.go
```diff
@@ -154,27 +156,46 @@ func (d *discovery) advertise() time.Duration {
...
+		case <-d.ctx.Done():
+			timer.Stop()
+			return
+		case <-timer.C:
+			if len(d.h.Network().Peers()) >= defaultMinPeers {
+				continue
+			}
+
+			// if our peer count is low, try to find some peers
+			timer := time.NewTimer(discoverLoopDuration)
```
dimalinux, line 174: This is a shadowed variable that isn't even the same type as the outer scope version. I wonder why the linter didn't flag it.

net/discovery.go
```diff
@@ -196,3 +217,17 @@ func (d *discovery) discover(provides types.ProvidesCoin,
 		}
 	}
 }
+
+func (d *discovery) discover(
+	provides types.ProvidesCoin,
+	searchTime time.Duration,
+) ([]peer.AddrInfo, error) {
+	log.Debugf("attempting to find DHT peers that provide [%s] for %vs...",
```
dimalinux, line 225: Normally, an ellipsis at the end of a log message message is used to indicate that there will be a matching log message after the long operation completes.

net/host.go
```diff
@@ -218,7 +218,7 @@ func (h *host) logPeers() {
 		}
 
 		log.Debugf("peer count: %d", len(h.h.Network().Peers()))
-		time.Sleep(time.Minute)
+		time.Sleep(time.Minute * 5)
```
dimalinux, line 221: Can we update this to use `common.SleepWithContext`?

net/discovery.go
```diff
@@ -154,27 +156,46 @@ func (d *discovery) advertise() time.Duration {
...
 
-	peerCh, err := d.rd.FindPeers(d.ctx, string(provides))
+func (d *discovery) findPeers(provides string, done <-chan time.Time) ([]peer.AddrInfo, error) {
+	peerCh, err := d.rd.FindPeers(d.ctx, provides)
 	if err != nil {
 		return nil, err
 	}
 
-	timer := time.NewTicker(searchTime)
 	peers := []peer.AddrInfo{}
```
dimalinux, line 192: If we are retesting this code right now, it would be a good time to update this declaration to `var peers []peer.AddrInfo`. Hopefully, we don't have any code that is treating nil and empty slices differently.

net/discovery.go
noot: i think it's the same type, but yeah i'll rename it to avoid confusion

net/host.go
noot: updated

net/discovery.go
noot: updated

net/discovery.go
noot: updated

net/discovery.go
```diff
@@ -154,27 +156,46 @@ func (d *discovery) advertise() time.Duration {
...
+			if err != nil {
+				log.Errorf("failed to find peers: %s", err)
+			}
+
+			findPeersTimer.Stop()
+		}
+	}
+}
+
+func (d *discovery) findPeers(provides string, done <-chan time.Time) ([]peer.AddrInfo, error) {
```
dimalinux, line 186: Just an idea to reduce code complexity:

You could pass in the timeout as a duration instead of a channel. This function could  then just wrap `d.ctx` using `context.WithTimeout` instead of creating a new channel.

noot: it would be nice to have the two cases be handled differently imo:
```
		select {
		case <-d.ctx.Done():
			return peers, d.ctx.Err()
		case <-done:
			return peers, nil
               // code omitted
```

what do you think?

dimalinux (approved):
Looks good! I had one last question if you have time to answer it.

net/discovery.go
dimalinux: Sorry, I edited my comment after seeing that the 2 cases were handled differently and you had time to reply before I finished. :) See above.

net/discovery.go
dimalinux: The returned types are similar, but slightly different:
```
func NewTicker(d Duration) *Ticker
func NewTimer(d Duration) *Timer
```

net/discovery.go
noot: oh lol, no worries! yeah, I guess we expect the timeout to always be hit basically, so we don't want to return an error

PR #254, opened 2022-12-08
validate swap timeouts set by counterparty

### LINKED ISSUE #161: validate swap timeouts set (as an xmr-maker)
since the xmr-taker can decide what timeouts to set in the swap contract, the xmr-maker needs to validate these and check that they're not unreasonably short or long.

noot:
closes #161


protocol/xmrmaker/checks.go
```diff
@@ -171,3 +172,29 @@ func (s *swapState) checkContract(txHash ethcommon.Hash) error {
 
 	return nil
 }
+
+func (s *swapState) checkAndSetTimeouts(t0, t1 *big.Int) error {
+	s.setTimeouts(t0, t1)
+
+	if s.Backend.Env() == common.Development {
+		return nil
+	}
```
dimalinux, line 181: Is there a way we can set this up to have the full body executed in all environments?

.github/workflows/checks.yml
```diff
@@ -12,6 +12,9 @@ jobs:
   linter-check:
     runs-on: ubuntu-22.04
     steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.x'
```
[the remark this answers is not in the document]
noot: yes

protocol/xmrmaker/checks.go
```diff
@@ -171,3 +172,29 @@ func (s *swapState) checkContract(txHash ethcommon.Hash) error {
...
+
+	expectedTimeout := common.SwapTimeoutFromEnvironment(s.Backend.Env())
+	allowableTimeDiff := time.Minute * 3
+
+	if time.Now().Add(expectedTimeout).Sub(s.t0).Abs() > allowableTimeDiff {
```
dimalinux, line 186: Does this work?  I thought `time.Now()` was the local time and `t0` was in UTC.  If we are going to depend on the wall-time of the client being accurate, maybe we should do a sanity check on it first? Compare it with the most recent block time? Then we can give a better error message when the client's time is off, which will definitely happen.

protocol/xmrmaker/checks.go
noot: yeah, it's just that the `allowableTimeDiff` is longer than the actual timeout for development, which is janky

protocol/xmrmaker/checks.go
```diff
@@ -171,3 +172,29 @@ func (s *swapState) checkContract(txHash ethcommon.Hash) error {
 
 	return nil
 }
+
+func (s *swapState) checkAndSetTimeouts(t0, t1 *big.Int) error {
```
dimalinux, line 176: Can we add a short description of what we are checking and why?  I'm not sure it will be immediately obvious to code auditors.

protocol/xmrmaker/checks.go
noot: they're both local time, `s.t0` is calculated with `time.Unix`: https://pkg.go.dev/time#Unix

noot: the alternative is to convert `time.Now()` to unix and compare to `t0`/`t1`, but these should be the same

dimalinux: That makes sense. I wasn't looking at the code in my IDE when I commented. `t0` at the top is `*big.Int`, but `s.t0` is `time.Time`. Thanks.

protocol/xmrmaker/checks.go
dimalinux: Just throwing out possible ideas:
* Divide by 20.  Your using 1/20th in stagenet/mainnet where the duration is an hour. For dev's 2 minutes, that would be 6 seconds.
* Have `SwapTimeoutFromEnvironment` return both the expected `T1-T0` duration as the max `T0` delta from the current time.

protocol/xmrmaker/checks.go
```diff
@@ -171,3 +172,29 @@ func (s *swapState) checkContract(txHash ethcommon.Hash) error {
...
+	}
+
+	expectedTimeout := common.SwapTimeoutFromEnvironment(s.Backend.Env())
+	allowableTimeDiff := time.Minute * 3
+
+	if time.Now().Add(expectedTimeout).Sub(s.t0).Abs() > allowableTimeDiff {
+		return errInvalidT0
+	}
+
+	if s.t1.Sub(s.t0).Abs() > allowableTimeDiff {
```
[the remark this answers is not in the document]
noot: hmm, I suppose it should be exact. updated

protocol/xmrmaker/checks.go
noot: yes, responded and added a comment here

protocol/xmrmaker/checks.go
noot: I updated to just divide by 20, I worry that may be too short of a diff time, but I'll see from the CI

protocol/xmrmaker/checks.go
dimalinux: Here's another possible description.  Pick and chose what you like out of it:
```
// checkAndSetTimeouts validates and sets our local timeout values. The XMR
// taker, when initiated the swap, must set the timeout to an exact value that
// depends on the network and environment. We also verify that the XMR taker
// didn't wait an excessive amount of time after locking the ETH asset, but
// before telling us.

```

protocol/xmrmaker/checks.go
```diff
@@ -171,3 +172,34 @@ func (s *swapState) checkContract(txHash ethcommon.Hash) error {
...
+
+	if s.Backend.Env() == common.Development {
+		allowableTimeDiff = time.Minute
+	}
+
+	if time.Now().Add(expectedTimeout).Sub(s.t0).Abs() > allowableTimeDiff {
+		return errInvalidT0
+	}
+
+	if s.t1.Sub(s.t0).Abs() == expectedTimeout {
```
dimalinux, line 195: Was that supposed to be `!=`?

protocol/xmrmaker/errors.go
```diff
@@ -21,6 +21,8 @@ var (
 	errLockTxReverted              = errors.New("other party failed to lock ETH asset (transaction reverted)")
 	errInvalidETHLockedTransaction = errors.New("eth locked tx was not to correct contract address")
 	errRelayerCommissionTooHigh    = errors.New("relayer commission must be less than 0.1 (10%)")
+	errInvalidT0                   = errors.New("invalid t0 value set by counterparty")
+	errInvalidT1                   = errors.New("invalid t1 value set by counterparty")
```
dimalinux, line 25: Should these messages be something like:

`ETH asset locked too far in the past`
`swap timeout set to unexpected value`

To get the clearest error messages, we should probably check the T1-T0 delta (the swap timeout) before checking the T0 value.

protocol/xmrmaker/checks.go
dimalinux: If the time deltas don't turn out to be exact enough for an exact comparison, we can subtract the big Int values instead.

noot: yes, fixed

protocol/xmrmaker/errors.go
noot: `ETH asset locked too far in the past` isn't necessarily true, the counterparty could have just put an invalid timeout. should I change the `errInvalidT1` message to say `delta between t0 and t1 is invalid`?

dimalinux: Alice doesn't set the T1-T0 delta directly.  We are using the delta as a proxy to determine if she passed the correct `_timeoutDuration` when calling `newSwap`. That's why I thought a message about the "timeout" or "timeout duration" being incorrect would be more helpful to users.

> ETH asset locked too far in the past isn't necessarily true, the counterparty could have just put an invalid timeout.

That's why I suggested that we flip the order of the checks and eliminate the possibility that the delta was set incorrectly first before performing the 2nd check.

noot:
@dimalinux I added back the ignoring timeouts if on development. The integration tests were all setting different timeouts to simulate refunds, and increasing that timeout to 2m will just increase the integration test time unnecessarily. The way to fix that would be to set the expected timeout on the xmrmaker in the tests, which I can add if you think it's necessary.

dimalinux:
> @dimalinux I added back the ignoring timeouts if on development. The integration tests were all setting different timeouts to simulate refunds, and increasing that timeout to 2m will just increase the integration test time unnecessarily. The way to fix that would be to set the expected timeout on the xmrmaker in the tests, which I can add if you think it's necessary.

That's fine. Maybe add a short comment above the early function exit stating that the integration tests change the timeout. Otherwise someone else will try to fix the code only to learn what you just learned.

dimalinux (approved):
I was hoping we might still update the error messages to describe what caused the problem, instead of describing how we detected the problem. Something that will help the end user troubleshoot the root cause quicker.

PR #257, opened 2022-12-11
Updated local.md

phazejeff:
I was deploying locally following the local.md instructions and found a few things that needed to be updated

### DIFF
--- a/docs/local.md
+++ b/docs/local.md
@@ -29,7 +29,7 @@ Build binary:
 make build
 ```
 
-This creates `swapd` and `swapcli` binaries in the `/bin` directory.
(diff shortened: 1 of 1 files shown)


docs/local.md
```diff
@@ -29,11 +29,11 @@ Build binary:
 make build
 ```
 
-This creates `swapd` and `swapcli` binaries in the root directory.
+This creates `swapd` and `swapcli` binaries in the `/bin` directory.
```
dimalinux, line 32: That looks like `/bin` in the root of the filesystem. Maybe say something like: "the `bin` folder at the top of the project"?

Thanks for the changes! Let's make that tiny fix and we'll get it merged.

phazejeff: Sure, just made that adjustment

PR #262, opened 2022-12-19
implement suggested exchange rate

### LINKED ISSUE #205: implement suggested exchange rate feature
currently, to make an offer, you must provide an exchange rate for the swap. this requires the user to manually calculate it themselves. it would be nice to add a subcommand for fetching a suggested exchange rate from an external API (coingecko or something like that?) this *would* be pretty centralized (unless we hit up many APIs and average them), so this should also notify the user to double check that they're okay with the rate.

noot:
Implement suggested exchange rate feature by querying the on-chain Chainlink price feeds. The needed feeds are only available on ethereum mainnet, so stagenet users, to get the full experience, get pushed to a public mainnet endpoint only for the feed data. Dev users get fixed price values.

The `swapcli` output currently looks like this:
```
./bin/swapcli suggested-exchange-rate
Exchange rate: 0.119497
XMR/USD Price: 170.9978      (2023-01-12 14:22:23 -0600 CST)
ETH/USD Price: 1430.98158542 (2023-01-12 14:55:35 -0600 CST)
```
closes #205

Misc changes:
* cleaned up script to generate go bindings from solidity
* updated swap factory byte code (hadn't been updated since last change)


common/get_prices.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, err
+	}
+
+	var to ethcommon.Address
+
+	switch chainID.Uint64() {
+	case 1:
+		// see https://data.chain.link/ethereum/mainnet/crypto-usd/eth-usd
+		to = ethcommon.HexToAddress("0x5f4ec3df9cbd43714fe2740f5e3616155c5b8419")
+	case 5, 1337, 31337:
```
dimalinux, line 76: Maybe we should have constants so people know what these are? Is this Goerli, Ganache, and Hardhat? We'll need at least some of them for the startup check you created a ticket for.

common/get_prices.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+		{
+			Type: uint80Ty,
+		},
+	}
+}
+
+// GetETHUSDPrice returns the current ETH/USD price from the Chainlink oracle.
+// It errors if the chain ID is not the Ethereum mainnet.
+func GetETHUSDPrice(ctx context.Context, ec *ethclient.Client) (*big.Int, error) {
```
dimalinux, line 64: The description says we are returning the ETH/USD price, but the value we are returning is a `*big.Int`. Can we add information on how the ratio is represented?

I added `t.Logf("ETH/USD price: %s", price)` to `TestGetETHUSDPrice_mainnet` and that gives `116704640000`. Do we need to be returning the number of decimal places as a 2nd parameter? Or is it a constant `8`?

noot: it's just the price of ETH in USD, and yes I believe it's a constant 8 decimals

common/get_prices.go
```diff
@@ -0,0 +1,130 @@
...
+
+	var to ethcommon.Address
+
+	switch chainID.Uint64() {
+	case 1:
+		// see https://data.chain.link/ethereum/mainnet/crypto-usd/eth-usd
+		to = ethcommon.HexToAddress("0x5f4ec3df9cbd43714fe2740f5e3616155c5b8419")
+	case 5, 1337, 31337:
+		// for goerli or test clients, just return 1
+		return big.NewInt(1), nil
```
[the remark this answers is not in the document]
noot: sure, I can change it to 1000 and 100 for ETH and XMR respectively

common/get_prices.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+		{
+			Type: uint256Ty,
+		},
+		{
+			Type: uint256Ty,
+		},
+		{
+			Type: uint80Ty,
+		},
```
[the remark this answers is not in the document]
dimalinux: `answeredInRound` is a legacy variable. Now always equal to `roundId`.

cmd/swapd/main.go
```diff
@@ -262,14 +262,17 @@ func setLogLevelsFromContext(c *cli.Context) error {
...
 	_ = logging.SetLogLevel("xmrmaker", level)
+	_ = logging.SetLogLevel("coins", level)
 	_ = logging.SetLogLevel("common", level)
+	_ = logging.SetLogLevel("contracts", level)
 	_ = logging.SetLogLevel("cmd", level)
+	_ = logging.SetLogLevel("extethclient", level)
+	_ = logging.SetLogLevel("monero", level)
 	_ = logging.SetLogLevel("net", level)
 	_ = logging.SetLogLevel("offers", level)
+	_ = logging.SetLogLevel("pricefeed", level)
```
dimalinux, line 273: `coins` and `pricefeed` were added. The others were just alphabetized.

common/config.go
```diff
@@ -76,7 +76,7 @@ var StagenetConfig = Config{
 			Port: 38081,
 		},
 	},
-	ContractAddress: ethcommon.HexToAddress("0xd2B5d6252D0645E4cF4Bb547E82A485F527BEFb7"),
+	ContractAddress: ethcommon.HexToAddress("0x01EeB71A63853fc89Ef26493bbdB7829F72b40d4"),
```
dimalinux, line 79: SwapFactory bytecode was updated and a new goerli contract was deployed.

pricefeed/pricefeed.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			return nil, err
+		}
+		defer ec.Close()
+	case common.GanacheChainID, common.HardhatChainID:
+		return &PriceFeed{
+			Description: "ETH / USD (fake)",
+			Price:       apd.New(123412345678, -8), // 1234.12345678
+			UpdatedAt:   time.Now(),
+		}, nil
```
dimalinux, line 67: Output from swapcli in development mode looks like this:
```
$ ./bin/swapcli suggested-exchange-rate
Exchange rate: 0.099766
XMR/USD Price: 123.12345678  (2023-01-12 15:32:36.542548651 -0600 CST)
ETH/USD Price: 1234.12345678 (2023-01-12 15:32:36.542929 -0600 CST)
```

rpc/swap.go
```diff
@@ -252,3 +257,44 @@ func (s *SwapService) Cancel(_ *http.Request, req *CancelRequest, resp *CancelRe
...
+	ETHPrice     *apd.Decimal        `json:"ethPrice"`
+	XMRUpdatedAt time.Time           `json:"xmrUpdatedAt"`
+	XMRPrice     *apd.Decimal        `json:"xmrPrice"`
+	ExchangeRate *coins.ExchangeRate `json:"exchangeRate"`
+}
+
+// SuggestedExchangeRate returns the current mainnet exchange rate, expressed as the XMR/ETH price.
+func (s *SwapService) SuggestedExchangeRate(_ *http.Request, _ *interface{}, resp *SuggestedExchangeRateResponse) error { //nolint:lll
+
+	ctx := context.Background() // TODO: There should be some parent context that would be better to use
```
dimalinux, line 273: This bothered me, but I couldn't find anywhere to grab it from.

scripts/generate-bindings.sh
```diff
@@ -10,29 +10,24 @@ if [[ -z "${SOLC_BIN}" ]]; then
...
+	echo "Generating go bindings for ${solidity_type_name}"
+
+	"${SOLC_BIN}" --abi "ethereum/contracts/${solidity_type_name}.sol" -o ethereum/abi/ --overwrite
+	"${SOLC_BIN}" --bin "ethereum/contracts/${solidity_type_name}.sol" -o ethereum/bin/ --overwrite
+	"${ABIGEN}" \
+		--abi "ethereum/abi/${solidity_type_name}.abi" \
+		--bin "ethereum/bin/${solidity_type_name}.bin" \
+		--pkg contracts \
+		--type "${go_type_name}" \
+		--out "ethereum/${go_file_name}.go"
```
dimalinux, line 27: I didn't want to do it in this PR, but I could easily update this to use the official solc docker images to compile the contract and then developers would not need a specific version of `solc` installed if that sounds like a good idea.

dimalinux (approved):

pricefeed/pricefeed.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch chainID.Uint64() {
+	case common.MainnetChainID:
+		// No extra work to do
+	case common.GoerliChainID:
+		// Push stagenet/goerli users to a mainnet endpoint
+		ec, err = ethclient.Dial(mainnetEndpoint)
```
noot, line 57: why? just to make it more like mainnet?

rpc/swap.go
noot: we could set a timeout for this?

pricefeed/pricefeed.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	decimals, err := chainlinkPriceFeedProxy.Decimals(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	price := apd.NewWithBigInt(new(apd.BigInt).SetMathBigInt(roundData.Answer), -int32(decimals))
+	_, _ = price.Reduce(price) // push even multiples of 10 to the exponent
+	updatedAt := time.Unix(roundData.StartedAt.Int64(), 0)
```
noot, line 130: why `roundData.StartedAt` instead of `UpdatedAt`?

pricefeed/pricefeed.go
```diff
@@ -0,0 +1,143 @@
...
+	default:
+		return nil, errUnsupportedNetwork
+	}
+
+	return getChainlinkPriceFeed(ctx, chainlinkXMRToUSDProxy, ec)
+}
+
+// getChainlinkPriceFeed retries the latest price feed data from the given contract address.
+func getChainlinkPriceFeed(ctx context.Context, feedAddress string, ec *ethclient.Client) (*PriceFeed, error) {
+
```
noot, line 108:
```suggestion
```

scripts/generate-bindings.sh
noot: does the solc docker image automatically handle versioning? ie will it use the right version for a given contract?

noot (commented):
looks good, not sure if I like having the `pricefeed` package in the project root though since it only has two files. move it to common maybe?

pricefeed/pricefeed.go
dimalinux: > why?

Because we want people testing the full functionality, and stagenet is the only option at the current time. If you try to start `swapd` with a mainnet endpoint, it will fail, because we don't have a SwapFactory contract deployed on mainnet yet.

> just to make it more like mainnet?

I think there is value in making the stagenet experience more or less identical to the mainnet experience, but without real money at stake. Stagenet is where they practice before going and doing the real thing.

pricefeed/pricefeed.go
dimalinux: They are identical, but you are right, that should have been UpdatedAt.  I'll fix it. Thanks.

dimalinux:
> looks good, not sure if I like having the `pricefeed` package in the project root though since it only has two files. move it to common maybe?

The `common` hierarchy is for things like global constants that are "common", in that all the other packages need access to them. In that sense, price feeds, doesn't seem like a fit with `common`'s purpose.  

The code in `pricefeed` is largely self contained and has a strong sense of specific purpose. I wouldn't let the fact that it only has 2 files in it bother you.

rpc/swap.go
dimalinux: I updated the code to pass the context in to `NewSwapService(...)`. I think that makes sense. The same context that cancels the local RPC services should also cancel requests made on behalf of those services.

scripts/generate-bindings.sh
dimalinux: They use separate docker containers for the different solc versions.  All the containers are in the same dockerhub repo, but you set the docker image tag to the solc version you want to use. If you need to dynamicly pick between different solc versions, you can parse the needed version from the pragma in the solidity file, and then use that version as the image tag when invoking docker. Doing that only makes sense if you need to compile different contracts with different versions though.

Here's some sample code from a personal project that I created:
```
	solidity_version="$(
		grep --max-count=1 '^pragma solidity' "${solidity_file_name}" |
			sed 's/.*\([[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+\).*/\1/'
	)"
	local solc_exec=(
		docker run --rm
		--volume "${PWD}/contracts:/contracts"
		-u "$(id -u):$(id -g)"
		"ethereum/solc:${solidity_version}"
	)
	"${solc_exec[@]}" --abi "${solidity_file_name}" -o contracts/abi/ --overwrite
	"${solc_exec[@]}" --bin "${solidity_file_name}" -o contracts/bin/ --overwrite
```

PR #265, opened 2022-12-20
support fallback monero nodes in default configs

dimalinux:
This PR allows us to configure multiple `monerod` nodes in our default configs.  They'll get tried sequentially and the first one that works gets used. Note: This PR doesn't solve the issue of fail-over to a new `monerod` node if one goes down while `swapd` is running. In that scenario, you'll need to restart `swapd`.

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -41,7 +41,19 @@ var MainnetConfig = Config{
 	DataDir: path.Join(baseDir, "mainnet"),
 	MoneroNodes: []*MoneroNode{
 		{
-			Host: "127.0.0.1",
+			Host: "node.sethforprivacy.com",
+			Port: 18089,
+		},
+		{
+			Host: "xmr-node.cakewallet.com",
+			Port: DefaultMoneroDaemonMainnetPort,
+		},
+		{
+			Host: "node.monerodevs.org",
+			Port: 18089,
+		},
+		{
+			Host: "node.community.rino.io",
 			Port: DefaultMoneroDaemonMainnetPort,
 		},
 	},
--- a/net/host_test.go
+++ b/net/host_test.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/binary"
 	"io"
-	"os"
 	"path"
 	"testing"
 
@@ -19,10 +18,8 @@ import (
 	"github.com/athanorlabs/atomic-swap/tests"
 )
 
-func TestMain(m *testing.M) {
+func init() {
 	logging.SetLogLevel("net", "debug")
-	m.Run()
-	os.Exit(0)
 }
 
 var testID = types.Hash{99}
--- a/monero/wallet_client_test.go
+++ b/monero/wallet_client_test.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"testing"
 
+	logging "github.com/ipfs/go-log"
 	"github.com/stretchr/testify/require"
 
 	"github.com/athanorlabs/atomic-swap/common"
@@ -13,6 +14,10 @@ import (
 )
 
 var moneroWalletRPCPath = path.Join("..", "monero-bin", "monero-wallet-rpc")
+
+func init() {
+	logging.SetLogLevel("monero", "debug")
+}
 
 func TestClient_Transfer(t *testing.T) {
 	amount := common.MoneroToPiconero(10) // 1k monero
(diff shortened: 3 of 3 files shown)


common/config.go
```diff
@@ -22,28 +22,49 @@ const (
...
+			Port: 38089,
+		},
+		{
+			Host: "node.monerodevs.org",
+			Port: 38089,
+		},
+		{
+			Host: "stagenet.community.rino.io",
+			Port: 38081,
+		},
```
dimalinux, line 65: I just pulled some nods from here: https://monero.fail/?nettype=stagenet
Suggestions on the best ones to pick are welcome.

common/config.go
```diff
@@ -22,28 +22,49 @@ const (
...
 // MainnetConfig is the mainnet ethereum and monero configuration
 var MainnetConfig = Config{
-	DataDir:          path.Join(baseDir, "mainnet"),
-	MoneroDaemonHost: "127.0.0.1",
-	MoneroDaemonPort: DefaultMoneroDaemonMainnetPort,
+	DataDir: path.Join(baseDir, "mainnet"),
+	MoneroNodes: []*MoneroNode{
+		{
+			Host: "127.0.0.1",
+			Port: DefaultMoneroDaemonMainnetPort,
```
dimalinux, line 45: Obviously, we want to encourage people to use their own nodes, but I could put some default nodes in here.

monero/wallet_client.go
```diff
@@ -175,11 +174,11 @@ func (c *walletClient) GetBalance(idx uint64) (*wallet.GetBalanceResponse, error
...
-// WaitForTransReceipt waits for the passed monero transaction ID to receive
-// numConfirmations and returns the transfer information. While this function will always
-// wait for the transaction to leave the mem-pool even if zero confirmations are
-// requested, it is the caller's responsibility to request enough confirmations that the
-// returned transfer information will not be invalidated by a block reorg.
+// WaitForReceipt waits for the passed monero transaction ID to receive numConfirmations
+// and returns the transfer information. While this function will always wait for the
+// transaction to leave the mem-pool even if zero confirmations are requested, it is the
+// caller's responsibility to request enough confirmations that the returned transfer
+// information will not be invalidated by a block reorg.
```
dimalinux, line 181: Only the first word (the function name) changed, the rest was auto-formatted to the new width.

monero/wallet_client.go
```diff
@@ -377,10 +376,27 @@ func (c *walletClient) Close() {
...
-// validateMonerodConfig validates the monerod node before we launch monero-wallet-rpc, as
+func findWorkingNode(env common.Environment, nodes []*common.MoneroNode) (*common.MoneroNode, error) {
+	if len(nodes) == 0 {
+		return nil, errors.New("no monero nodes")
+	}
+	var err error
+	for _, n := range nodes {
+		err = validateMonerodNode(env, n)
+		if err != nil {
+			log.Warnf("Non-working node: %s", err)
```
dimalinux, line 387: `validateMonerodNode` already includes the node name in the error message.

monero/wallet_client.go
```diff
@@ -415,32 +431,22 @@ func validateMonerodConfig(env common.Environment, monerodHost string, monerodPo
...
-		switch conf.Env {
-		case common.Mainnet, common.Development:
-			conf.MonerodPort = common.DefaultMoneroDaemonMainnetPort
-		case common.Stagenet:
-			conf.MonerodPort = common.DefaultMoneroDaemonStagenetPort
-		default:
-			panic("unhandled environment value")
-		}
+	if len(conf.MonerodNodes) == 0 {
+		conf.MonerodNodes = common.ConfigDefaultsForEnv(conf.Env).MoneroNodes
```
dimalinux, line 445: `swapd`, at least right now, is always setting `conf.MonerodNodes`, but tests that don't want to configure all the values use this.

net/host_test.go
```diff
@@ -19,10 +18,8 @@ import (
 	"github.com/athanorlabs/atomic-swap/tests"
 )
 
-func TestMain(m *testing.M) {
+func init() {
 	logging.SetLogLevel("net", "debug")
-	m.Run()
-	os.Exit(0)
 }
```
dimalinux, line 23: I think I wrote this, but I don't know why I didn't just use `init`. I think I was searching for a way of linking our logger to the test logger, but didn't find it.

cmd/swapd/main.go
```diff
@@ -643,17 +648,26 @@ func newBackend(
...
+				return nil, errFlagValueEmpty(flagMoneroDaemonHost)
+			}
+		}
+		if c.IsSet(flagMoneroDaemonPort) {
+			node.Port = c.Uint(flagMoneroDaemonPort)
+			if node.Port == 0 {
+				return nil, errFlagValueZero(flagMoneroDaemonPort)
+			}
+		}
+		cfg.MoneroNodes = []*common.MoneroNode{node}
```
noot, line 668: do you think it would be useful to allow users to specify multiple monero daemons via cli? I guess since we only end up using one anyways, it probably isn't super useful.

noot (approved):
looks good!

cmd/swapd/main.go
dimalinux: Hopefully, people are using the flag to specify their own local server, in which case it wouldn't be a useful feature.  If we figure out that it is a valid use case in the future, I think we'd need to switch to a host:port syntax that possibly defaults the port if no colon is given.

PR #266, opened 2023-01-02
bump go-dleq version

noot:


dimalinux (approved):
Should we start tagging the `go-dleq` changes with release tags (for versioning)?

noot:
@dimalinux sure, I don't expect the API to change though

Issue #209, opened 2022-10-23
relayer support: add relayer discovery via network


noot:
relies on https://github.com/AthanorLabs/go-relayer/issues/1

noot:
this can be done after #271

noot:
possibly part of this issue: if a maker makes an offer but they have no ETH balance, they should be alerted of this, and a relayer should automatically be used for the claim transaction.

PR #268, opened 2023-01-03
refactor net package to separate generic and swap-specific functionality

noot:
- create `net/swapnet` package which implements a swap-specific host; ie. a host that supports network messages used in the swap.
- update `net` to be a more generic libp2p host that supports discovery and the ability to register stream handlers

related to #209


net/host.go
```diff
@@ -181,42 +169,24 @@ func NewHost(cfg *Config) (*host, error) {
 	hst := &host{
 		ctx:        ourCtx,
 		cancel:     cancel,
-		protocolID: fmt.Sprintf("%s/%s/%s/%d", protocolID, protocolVersion, cfg.Environment, cfg.EthChainID),
+		protocolID: fmt.Sprintf("%s/%s/%d", cfg.ProtocolID, cfg.Environment, cfg.EthChainID),
```
dimalinux, line 172: How are we preventing hosts with incompatible API versions from talking to each other?

dimalinux: Answer: The version is now part of the ProtocolID.

dimalinux:
Can we add package comments to the `swapnet` package?  Besides being helpful, we'll need them when we update the linter version.

net/swapnet/host.go
```diff
@@ -0,0 +1,134 @@
...
+const (
+	protocolID     = "/atomic-swap/0.1"
+	maxMessageSize = 1 << 17
+)
+
+var log = logging.Logger("host")
+var _ Host = &host{}
+
+// Host represents a peer-to-peer node (ie. a host)
+type Host interface {
```
dimalinux, line 26: I can't keep these types straight in my head.  Both the `swapnet` package and the `net` package above it have a file named `host.go`. In that file, they both have an interface named `Host` and the comment above both interfaces is 100% identical. Can we reduce the cognitive load with comments and maybe some file/type-name differentiation?

net/host_test.go
```diff
@@ -13,52 +13,16 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/athanorlabs/atomic-swap/common"
-	"github.com/athanorlabs/atomic-swap/common/types"
 	"github.com/athanorlabs/atomic-swap/net/message"
 	"github.com/athanorlabs/atomic-swap/tests"
 )
 
+const defaultMaxMessageSize = 1 << 17
```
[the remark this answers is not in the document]
noot: no it's only passed in to the read stream function now (in this package). it's not used anywhere else.

net/swapnet/host.go
noot: okay I can updated the comment, what do you suggest renaming the hosts? should this be `SwapHost` or something (although I feel that's kind of redundant)

dimalinux: Would it be correct to say that `net.Host` provides the raw libp2p "node" functionality and that `swapnet.Host` extends the functionality of this "node" to make it a "peer"? Then you could have `net.Node` and `swapnet.Peer` as possible names.

I'm just throwing out ideas. Since I didn't come up with the abstraction, my ideas may be less than perfect. The goal is to have more distinct (and hopefully, intuitive) naming, so readers/developers can quickly keep the pieces separate in their heads and understand how the pieces interact with each other.

net/host_test.go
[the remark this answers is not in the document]
dimalinux: If they need to be the same value, as the name implies, you could put the `MaxMessageSize` constant in this package, and then have the `swapnet` package reference it here. If the value here in tests does not need to be linked, you could just call the constant here `testMessageSize` to make it clear that the values are independent.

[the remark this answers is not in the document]
noot: I'll just name it `testMessageSize`, the message size should be set by the package that uses `net.Host`

PR #269, opened 2023-01-04
Fixed Point Math

dimalinux:
This PR updates our code to use the fixed point match library `apd.Decimal`. Using floats for money, especially ethereum, is problematic because:
* Conversion back and forth between base2 to base10 is lossy.
* ETHER requires 18 decimal places of precision, but float64
  only handles around 16 decimals.

Note:
Before this PR, the piconero, wei and ERC20 types where in `common` and the exchange rate type was in `common/types`. This was causing problems, because support methods for all 4 types required access to the same constants and math helper functions. To solve this issue, I pulled the types out into a new, top-level `coins` package.


cmd/swapcli/main.go
```diff
@@ -772,20 +792,34 @@ func runGetSwapTimeout(ctx *cli.Context) error {
...
+
+	xRate := o.ExchangeRate
+	minETH, err := xRate.ToETH(o.MinAmount)
+	if err != nil {
+		return err
+	}
+	maxETH, err := xRate.ToETH(o.MaxAmount)
+	if err != nil {
+		return err
+	}
```
dimalinux, line 808: It's a little ugly that we have to do error handling here. In theory, we will never receive values, like a zero-valued exchange rate triggering division by zero, that would cause this error handling to be invoked.

cmd/swapcli/main.go
```diff
@@ -379,12 +381,12 @@ func runBalances(ctx *cli.Context) error {
...
 
 	fmt.Printf("Ethereum address: %s\n", balances.EthAddress)
-	fmt.Printf("Balance: %s\n", common.FmtFloat((*common.WeiAmount)(balances.WeiBalance).AsEther()))
+	fmt.Printf("ETH Balance: %s\n", balances.WeiBalance.AsEther().Text('f'))
 	fmt.Println()
 	fmt.Printf("Monero address: %s\n", balances.MoneroAddress)
-	fmt.Printf("Balance: %s\n", common.FmtFloat(common.PiconeroAmount(balances.PiconeroBalance).AsMonero()))
-	fmt.Printf("Unlocked balance: %s\n",
-		common.FmtFloat(common.PiconeroAmount(balances.PiconeroUnlockedBalance).AsMonero()))
+	fmt.Printf("XMR Balance: %s\n", balances.PiconeroBalance.AsMonero().Text('f'))
```
dimalinux, line 387: On `apd.Decimal` we have to use `Text('f')` instead of `String()` or it will sometimes use exponent syntax. With the `String()` method on our own types like `PiconeroAmount`, `WeiAmount` and `ExchangeRate`, it is not needed.

cmd/swapcli/main.go
```diff
@@ -451,30 +456,41 @@ func runQueryAll(ctx *cli.Context) error {
...
 
-	exchangeRate := ctx.Float64(flagExchangeRate)
-	if exchangeRate == 0 {
-		return errNoExchangeRate
+	exchangeRateDec, err := readUnsignedDecimalFlag(ctx, flagExchangeRate)
+	if err != nil {
+		return err
+	}
+	exchangeRate := coins.ToExchangeRate(exchangeRateDec)
+	// TODO: How to handle this if the other asset is not ETH?
```
noot, line 485: this should check if the asset is an ERC20, and if it is, then read the decimals from the contract and set the `otherMin/Max` from that.

cmd/swapcli/main.go
```diff
@@ -643,13 +657,16 @@ func runGetPastSwap(ctx *cli.Context) error {
...
 	}
 
-	fmt.Printf("Provided: %s\n ProvidedAmount: %v\n ReceivedAmount: %v\n ExchangeRate: %v\n Status: %s\n",
-		info.Provided,
-		info.ProvidedAmount,
-		info.ReceivedAmount,
-		info.ExchangeRate,
-		info.Status,
-	)
+	receivedCoin := "ETH"
```
noot, line 660: it might not be ETH necessarily, check the info for the asset instead?

noot: also if we're a maker this will be XMR

coins/coins.go
```diff
@@ -0,0 +1,252 @@
+// Package coins provides types, conversions and exchange calculations for dealing
+// with cryptocurrency coins and ECR20 tokens.
```
noot, line 2:
```suggestion
// with cryptocurrency coins and ERC20 tokens.
```

coins/coins.go
```diff
@@ -0,0 +1,252 @@
...
+func (a *PiconeroAmount) Decimal() *apd.Decimal {
+	return (*apd.Decimal)(a)
+}
+
+// Uint64 converts piconero amount to uint64. Errors if a is negative or larger than 2^63-1.
+func (a *PiconeroAmount) Uint64() (uint64, error) {
+	// Hopefully, the rest of our code is doing input validation and the error below
+	// never gets triggered.
+	if a.Negative {
+		return 0, fmt.Errorf("can not convert %s to unsigned", a.String())
```
noot, line 38:
```suggestion
		return 0, fmt.Errorf("cannot convert %s to unsigned", a.String())
```

coins/coins.go
```diff
@@ -0,0 +1,252 @@
...
+	"fmt"
+	"math/big"
+	"strconv"
+
+	"github.com/cockroachdb/apd/v3"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// PiconeroAmount represents some amount of piconero (the smallest denomination of monero)
+type PiconeroAmount apd.Decimal
```
noot, line 15: why does this need to be `apd.Decimal`? it'll always be a positive integer value, so uint64 or *big.Int would also be fine

coins/coins.go
```diff
@@ -0,0 +1,252 @@
...
+// AsMonero converts the piconero PiconeroAmount into standard units
+func (a *PiconeroAmount) AsMonero() *apd.Decimal {
+	xmrAmt := new(apd.Decimal).Set(a.Decimal())
+	decreaseExponent(xmrAmt, NumMoneroDecimals)
+	_, _ = xmrAmt.Reduce(xmrAmt)
+	return xmrAmt
+}
+
+// WeiAmount represents some amount of ether in the smallest denomination (wei)
+type WeiAmount apd.Decimal
```
noot, line 107: same here, why not *big.Int since this is always an integer?

coins/coins.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// String returns the wei amount as a base10 string
+func (a *WeiAmount) String() string {
+	return a.Decimal().Text('f')
+}
+
+// ERC20TokenAmount represents some amount of an ERC20 token in the smallest denomination
+type ERC20TokenAmount struct {
+	amount      *apd.Decimal
```
noot, line 190: same here

monero/test_support.go
```diff
@@ -116,17 +117,20 @@ func TestBackgroundMineBlocks(t *testing.T) {
...
 // MineMinXMRBalance enables mining for the passed wc wallet until it has an unlocked balance greater
 // than or equal to minBalance.
-func MineMinXMRBalance(t *testing.T, wc WalletClient, minBalance common.PiconeroAmount) {
+func MineMinXMRBalance(t *testing.T, wc WalletClient, minBalance *coins.PiconeroAmount) {
 	daemonCli := monerorpc.New(MonerodRegtestEndpoint, nil).Daemon
 	addr, err := wc.GetAddress(0)
 	require.NoError(t, err)
 	t.Log("mining to address:", addr.Address)
 
+	minBal64Bit, err := minBalance.Decimal().Int64()
```
noot, line 126:
```suggestion
	minBal64Bit, err := minBalance.Uint64()
```

monero/wallet_client_test.go
```diff
@@ -6,29 +6,38 @@ import (
...
 	"github.com/stretchr/testify/require"
 
+	"github.com/athanorlabs/atomic-swap/coins"
 	"github.com/athanorlabs/atomic-swap/common"
 	mcrypto "github.com/athanorlabs/atomic-swap/crypto/monero"
 )
 
 var moneroWalletRPCPath = path.Join("..", "monero-bin", "monero-wallet-rpc")
 
+// package "tests" depends on the "monero", so StrToDecimal is in "common" so both can use it.
```
noot, line 20:
```suggestion
// package "tests" depends on the "monero" package, so StrToDecimal is in "common" so both can use it.
```

net/discovery.go
```diff
@@ -28,7 +28,7 @@ type discovery struct {
 	dht         *dual.DHT
 	h           libp2phost.Host
 	rd          *libp2prouting.RoutingDiscovery
-	provides    []types.ProvidesCoin // set to a single item slice of XMR when we make an offer
+	provides    []coins.ProvidesCoin // set to a single item slice of XMR when we make an offer
```
noot, line 31: I need to update this to be `[]string` lol. will do when I move this to a separate repo

net/message/message.go
```diff
@@ -106,14 +107,14 @@ func (m *QueryResponse) Type() Type {
...
 type SendKeysMessage struct {
-	OfferID            types.Hash
-	ProvidedAmount     float64
-	PublicSpendKey     string
-	PublicViewKey      string
-	PrivateViewKey     string
-	DLEqProof          string
-	Secp256k1PublicKey string
-	EthAddress         string
+	OfferID            types.Hash   `json:"offerID"`
```
noot, line 110: why are JSON tags added to only this message? seems unnecessary

protocol/xmrmaker/claim_test.go
```diff
@@ -18,6 +19,7 @@ import (
 	"github.com/athanorlabs/go-relayer/relayer"
 	rrpc "github.com/athanorlabs/go-relayer/rpc"
 
+	"github.com/athanorlabs/atomic-swap/coins"
 	//"github.com/athanorlabs/atomic-swap/common"
```
noot, line 23: remove pls

protocol/xmrmaker/net.go
```diff
@@ -99,22 +103,30 @@ func (inst *Instance) HandleInitiateMessage(msg *net.SendKeysMessage) (net.SwapS
...
 
-	providedAmount := offer.ExchangeRate.ToXMR(msg.ProvidedAmount)
+	providedAmount, err := offer.ExchangeRate.ToXMR(msg.ProvidedAmount)
+	if err != nil {
+		return nil, nil, err
+	}
 
-	if providedAmount < offer.MinAmount {
+	if providedAmount.Cmp(offer.MinAmount) < 0 {
+		// TODO: This message will be confusing to the end-user, since they provided ETH, not XMR
```
noot, line 121: I think it's fine since it's returned to the taker (ETH-provider) saying the amount they provided is too low?

protocol/xmrmaker/errors.go
```diff
@@ -31,49 +30,49 @@ var (
...
 
 type errAmountProvidedTooLow struct {
-	providedAmount float64
-	minAmount      float64
+	providedAmount *apd.Decimal
+	minAmount      *apd.Decimal
 }
 
 func (e errAmountProvidedTooLow) Error() string {
 	return fmt.Sprintf("%s XMR provided by taker is under offer minimum of %s XMR",
```
noot, line 50:
```suggestion
	return fmt.Sprintf("%s ETH provided by taker is under offer minimum of %s XMR",
```

protocol/xmrmaker/errors.go
```diff
@@ -31,49 +30,49 @@ var (
...
 
 type errAmountProvidedTooHigh struct {
-	providedAmount float64
-	maxAmount      float64
+	providedAmount *apd.Decimal
+	maxAmount      *apd.Decimal
 }
 
 func (e errAmountProvidedTooHigh) Error() string {
 	return fmt.Sprintf("%s XMR provided by taker is over offer maximum of %s XMR",
```
noot, line 62:
```suggestion
	return fmt.Sprintf("%s ETH provided by taker is over offer maximum of %s XMR",
```
these are my bad lol

protocol/xmrtaker/message_handler.go
```diff
@@ -81,8 +81,14 @@ func (s *swapState) setNextExpectedEvent(event EventType) error {
...
 
 func (s *swapState) handleSendKeysMessage(msg *net.SendKeysMessage) (net.Message, error) {
-	if msg.ProvidedAmount < s.info.ReceivedAmount {
-		return nil, fmt.Errorf("receiving amount is not the same as expected: got %v, expected %v",
+
+	if msg.ProvidedAmount == nil {
+		return nil, errMissingProvidedAmount
+	}
+
+	// TODO: THIS CHECK LOOKS BACKWARDS?!?
```
noot, line 89: I think it's fine? we're checking that what the counterparty is claiming they will provide is the same (or greater than) what we expected them to, as calculated when the swap state was created

protocol/xmrtaker/message_handler.go
```diff
@@ -224,9 +230,10 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) error {
 	log.Debugf("checking locked wallet, address=%s balance=%d blocks-to-unlock=%d",
 		lockedAddr, balance.Balance, balance.BlocksToUnlock)
 
-	if balance.Balance < uint64(s.receivedAmountInPiconero()) {
+	// TODO: I'm confused. If it is received (past tense), how is this possible here?
```
noot, line 233: `receivedAmountInPiconero` could be named `expectedAmountToReceiveInPiconero`

protocol/xmrtaker/message_handler.go
```diff
@@ -224,9 +230,10 @@ func (s *swapState) handleNotifyXMRLock(msg *message.NotifyXMRLock) error {
...
+	if s.receivedAmountInPiconero().CmpU64(balance.Balance) > 0 {
```
noot, line 234: this will error if the amounts are the same which is bad

noot (commented):
seems mostly good, a few comments and questions about the type choices

coins/coins.go
dimalinux: It doesn't have to be, but I think there is value in treating all the currencies the same way. I didn't want to take it on in this PR, but XMR, ETH and ERC20 tokens could all be handled with a single class that tracks the decimals, unit-name/currency-name, etc. if we want to do it that way in the future. It's convenient to go back and forth with XMR just by adjusting the exponent. After working with `apd.Decimal` for a while, I really like it. For money, it seems like a definite improvement over `big.Int`.

coins/coins.go
dimalinux: `apd.Decimal` is basically an improved `big.Int` combined with an exponent. It has much better performance characteristics than `big.Int`, because it will use built-in types if the value of the coefficient is small enough. Pushing the zeros out of the coefficient and into the exponent also improves its performance. That said, we don't care much about performance. It is the fact that the math operations are better, are lossless in base10, provide any type of rounding you may want, and provide very precise error reasons that make the package nice to work with. It is maintained by Cockroach DB, which has a $5 billion valuation and is by far the best relational database I've ever used/administrated, so support-wise it seem like a safe choice. If we are going to use `apd.Decimal` for ETH and developers have to learn the basics of it anyways, we are best off using it it for WEI, XMR and all the other currencies was my opinion. If we stored WEI as a `big.Int`, we'd still have to convert it to `apd.Decimal` before doing any kind of calculation with the exchange rate or even converting it back into ETH. Play around with it some. After learning the basics, you'll probably agree. If you have question, I can probably ask answer them quickly.

monero/wallet_client_test.go
dimalinux: Thanks for catching that. That was legacy code before I broke the coins off into their own package. I just updated the references here to use `coins.StrToDecimal` directly.

net/message/message.go
dimalinux: I thought our goal was to use consistent naming in all of our JSON handling. I noticed that this struct didn't have tags and I didn't want to write "ProvidedAmount" (with a capital "P") in the code that checks and generates an error message if this field is invalid.

cmd/swapcli/main.go
dimalinux: That is what I tried to do first, but info's type doesn't have the needed fields:
```
type GetPastResponse struct {
	Provided       coins.ProvidesCoin  `json:"provided"`
	ProvidedAmount *apd.Decimal        `json:"providedAmount"`
	ReceivedAmount *apd.Decimal        `json:"expectedAmount"`
	ExchangeRate   *coins.ExchangeRate `json:"exchangeRate"`
	Status         string              `json:"status"`
}
```
We'll have to fix it in a future PR.

cmd/swapcli/main.go
dimalinux: We are in the swapcli client here.  Would we implement that as a separate API call?  I'm asking for future development, as these are not minor changes that would fit into this PR.

net/message/message.go
dimalinux: I'll go through all the message types in a future PR, make sure they have consistent JSON tags, and, more importantly, make sure we are doing all needed input validations on all the fields.

protocol/xmrmaker/net.go
dimalinux: The error messages are fixed and I removed the TODO.

protocol/xmrmaker/errors.go
```diff
@@ -31,49 +30,49 @@ var (
...
+	maxOfferAmount  *apd.Decimal
+	unlockedBalance *apd.Decimal
 }
 
 func (e errUnlockedBalanceTooLow) Error() string {
 	return fmt.Sprintf("balance %s XMR is too low for maximum offer amount of %s XMR",
-		strconv.FormatFloat(e.minAmount, 'f', -1, 64),
-		strconv.FormatFloat(e.unlockedBalance, 'f', -1, 64),
+		e.maxOfferAmount.String(),
+		e.unlockedBalance.String(),
```
noot, line 76:
```suggestion
		e.unlockedBalance.String(),
		e.maxOfferAmount.String(),
```

shouldn't these be switched?

rpc/swap.go
```diff
@@ -54,11 +56,11 @@ type GetPastRequest struct {
...
 // GetPastResponse ...
 type GetPastResponse struct {
-	Provided       types.ProvidesCoin `json:"provided"`
-	ProvidedAmount float64            `json:"providedAmount"`
-	ReceivedAmount float64            `json:"receivedAmount"`
-	ExchangeRate   types.ExchangeRate `json:"exchangeRate"`
-	Status         string             `json:"status"`
+	Provided       coins.ProvidesCoin  `json:"provided"`
+	ProvidedAmount *apd.Decimal        `json:"providedAmount"`
+	ReceivedAmount *apd.Decimal        `json:"expectedAmount"`
```
noot, line 61:
```suggestion
	ReceivedAmount *apd.Decimal        `json:"receivedAmount"`
```

since the swap was in the past, so the amount was received already?

noot (approved):
looks good - only a couple minor comments

PR #270, opened 2023-01-08
remove previous net/* contents, move net/swapnet/* to net/

noot:
as title says, diff looks messed up because net/swapnet was moved to net/. previous net/ contents were moved here: https://github.com/AthanorLabs/go-p2p-net 

also feel free to suggest on package naming, right now the package in net/ is still called swapnet because the package in the above repo is called net

related to #209


cmd/swapd/main.go
```diff
@@ -23,13 +23,13 @@ import (
...
 	"github.com/athanorlabs/atomic-swap/monero"
-	"github.com/athanorlabs/atomic-swap/net"
-	"github.com/athanorlabs/atomic-swap/net/swapnet"
+	swapnet "github.com/athanorlabs/atomic-swap/net"
 	"github.com/athanorlabs/atomic-swap/protocol/backend"
 	"github.com/athanorlabs/atomic-swap/protocol/swap"
 	"github.com/athanorlabs/atomic-swap/protocol/xmrmaker"
 	"github.com/athanorlabs/atomic-swap/protocol/xmrtaker"
 	"github.com/athanorlabs/atomic-swap/rpc"
+	net "github.com/athanorlabs/go-p2p-net"
```
dimalinux, line 32: What if we called the repo `go-p2pnet`, and then imported it without changing the name? Unless AnthorLabs will be providing the code in another language, even the `go-` prefix is probably unnecessary. We want to differentiate the library from the standard library's `net` package which users may be importing in the same file. `p2pnet` doesn't currently (and probably never will) conflict with any `go-libp2p` package name.

noot: I like it, updated!

PR #281, opened 2023-01-19
fix: update offer ID to be hash of offer fields

### LINKED ISSUE #273: OffersID should be a hash of the offer
The offerID needs to be a hash of the fields, including a nonce, and the taker need to check the hash.

noot:
- update offer type to have nonce
- update offer ID to be hash of all its field
- add ID validation on json unmarshalling; unmarshal will fail if the hash of the fields isn't the same as the ID provided

closes #273

### DIFF
--- a/common/types/offer.go
+++ b/common/types/offer.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"crypto/rand"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -32,6 +33,7 @@ type Offer struct {
 	MaxAmount    *apd.Decimal        `json:"maxAmount"` // Max XMR amount
 	ExchangeRate *coins.ExchangeRate `json:"exchangeRate"`
 	EthAsset     EthAsset            `json:"ethAsset"`
+	Nonce        uint64              `json:"nonce"`
 }
 
 // NewOffer creates and returns an Offer with an initialised ID and Version fields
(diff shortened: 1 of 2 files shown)


common/types/offer.go
```diff
@@ -42,30 +44,53 @@ func NewOffer(
...
+}
+
+func (o *Offer) hash() Hash {
+	b := append([]byte(o.Version.String()), []byte(o.Provides)...)
+	b = append(b, []byte(o.MinAmount.Text('f'))...)
+	b = append(b, []byte(o.MaxAmount.Text('f'))...)
+	b = append(b, []byte(o.ExchangeRate.String())...)
+	b = append(b, o.EthAsset[:]...)
+	b = append(b, []byte(fmt.Sprintf("%d", o.Nonce))...)
+	return sha3.Sum256(b)
```
dimalinux, line 81: Should we have comma separators to make it a human readable string?

dimalinux (approved):

common/types/offer.go
noot: I don't think we need it to be human readable. I wanted to have the minimum amount of bytes in here as possible, so there's less that can be manipulated

dimalinux: The commas, assuming no field can actually contain a comma, actually make it harder to manipulate. Without the commas, a min value of 111 and a max value of 222 becomes the same hash as a min value of 1 with a max value of 11222.

PR #282, opened 2023-01-22
feat: implement swap relayer cmd package

### LINKED ISSUE #271: implement swap relayer cmd package
building off the "generic" relayer [here](https://github.com/AthanorLabs/go-relayer/pull/9), we should implement a swap-specific relayer that lives in this repo.

specifically, it needs:
- p2p configuration; it should advertise on a swap-specific protocol, so they can be discovered easily (relayer discovery should be implemented in another PR)
- validation function impl; the relayer should only accept transactions that call the swap contract, and with a certain commission rate (specific by the relayer via CLI)

noot:
- implement cmd/relayer which contains a swap-specific relayer program.
- it advertises on the libp2p protocol `/atomic-swap/0.1/<chain-id>/relayer/0.1`
- it has a default commission of 1%
- it implements a validation function which allows only calls to the `claimRelayer()` function in a swap contract, and also checks that the fee paid is >= `<commission-percentage>`

closes #271


cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+		&cli.UintFlag{
+			Name:  flagLibp2pPort,
+			Usage: "libp2p port to listen on",
+			Value: defaultLibp2pPort,
+		},
+		&cli.StringSliceFlag{
+			Name:    flagBootnodes,
+			Aliases: []string{"bn"},
+			Usage:   "libp2p bootnode, comma separated if passing multiple to a single flag",
+			EnvVars: []string{"SWAPD_BOOTNODES"},
```
dimalinux, line 114: Are these the same as swapd's bootnodes?

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+	"github.com/athanorlabs/atomic-swap/common"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+	swapnet "github.com/athanorlabs/atomic-swap/net"
+
+	logging "github.com/ipfs/go-log"
+)
+
+const (
+	flagDataDir          = "datadir"
+	flagEndpoint         = "endpoint"
```
dimalinux, line 36: Is this the ETH endpoint?  We call this flag `ethereum-endpoint` in swapd, which is a little on the long side to type. Would `eth-endpoint` in both places be less confusing, but easy to type and consistent across programs?

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+	"github.com/athanorlabs/atomic-swap/cliutil"
+	"github.com/athanorlabs/atomic-swap/common"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+	swapnet "github.com/athanorlabs/atomic-swap/net"
+
+	logging "github.com/ipfs/go-log"
+)
+
+const (
+	flagDataDir          = "datadir"
```
dimalinux, line 35: This is `data-dir` in swapd. We should be consistent across programs or end users will be using curse words about us.  :)

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+			Value: "{HOME}/.atomicswap/{ENV}", // For --help only, actual default replaces variables
+		},
+		&cli.StringFlag{
+			Name:  flagEndpoint,
+			Value: "http://localhost:8545",
+			Usage: "Ethereum RPC endpoint",
+		},
+		&cli.StringFlag{
+			Name:  flagKey,
+			Value: "eth.key",
```
dimalinux, line 69: Do we want this to be in the current working directory?  This is what we use with `swapd`: https://github.com/AthanorLabs/atomic-swap/blob/master/cmd/swapd/main.go#L104-L108

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+
+	err = h.Start()
+	if err != nil {
+		return nil, err
+	}
+
+	return h, nil
+}
+
+func deployOrGetForwarder(
```
dimalinux, line 351: Is there another package that this might fit better in? It certainly doesn't look like a function that belongs in `main.go`.

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+		return nil, ethcommon.Address{}, fmt.Errorf("failed to make transactor: %w", err)
+	}
+
+	if addressString == "" {
+		address, tx, _, err := rcontracts.DeployForwarder(txOpts, ec) //nolint:govet
+		if err != nil {
+			return nil, ethcommon.Address{}, err
+		}
+
+		_, err = bind.WaitMined(context.Background(), ec, tx)
```
dimalinux, line 368: Don't we have a context that should have been passed in for this?

cmd/relayer/validate.go
```diff
@@ -0,0 +1,156 @@
+package main
```
dimalinux, line 1: This doesn't seem like the right package for this.  Maybe we should create a hierarchy under the top level `relayer` package to separate server and client code? Code in `cmd/PROG_NAME` is for the command line shell. Parsing the command line, setting up the signal handler, reading config files, etc. are all in scope, but it is generally not where you would expect to find the entire program's implementation.

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
+package main
```
dimalinux, line 1: Can you add package comments.  See swapd's main as as example. Just a brief description of what the relayer is, and a mention that this is only the entrypoint for the relayer (defining the package's scope).

cmd/relayer/validate.go
```diff
@@ -0,0 +1,156 @@
...
+		0,
+	)
+
+	percentage := new(apd.Decimal)
+	_, err = coins.DecimalCtx().Quo(percentage, feeD, valueD)
+	if err != nil {
+		return err
+	}
+
+	if percentage.Cmp(minFeePercentage) == -1 {
```
dimalinux, line 148: I would change this to `< 0`. This particular `Cmp` implementation may return `-1`, but the general API of compare methods is only that they return a negative value if the left hand side is less than the right hand side.

cmd/relayer/validate.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create bytes32 type: %w", err)
+	}
+
+	addressTy, err := abi.NewType("address", "", nil)
+	if err != nil {
+		return fmt.Errorf("failed to create address type: %w", err)
+	}
+
+	arguments := abi.Arguments{
```
dimalinux, line 78: Is there a contract, with generated code, that we can use to make this easier to cross check and auto-fail if something is incorrect?

cmd/swapd/main_test.go
```diff
@@ -150,14 +151,14 @@ func Test_expandBootnodes(t *testing.T) {
 		"node7",
 		"node8",
 	}
-	require.EqualValues(t, expected, expandBootnodes(cliNodes))
```
dimalinux, line 153: This whole test belongs in the package where the function it is testing is implemented.

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+		},
+		&cli.UintFlag{
+			Name:  flagRPCPort,
+			Value: 7799,
+			Usage: "Relayer RPC server port",
+		},
+		&cli.BoolFlag{
+			Name:  flagRPC,
+			Value: false,
+			Usage: "Run the relayer RPC server. Defaults false",
```
dimalinux, line 84: We run the relayer server regardless, if I'm understanding this correctly. I don't think the end user needs to understand that accepting RPC requests is also a "server". Maybe the description should say something like "Accept RPC requests on localhost"? What are the requests that we accept?

cmd/relayer/main.go
```diff
@@ -0,0 +1,412 @@
...
+		},
+		&cli.BoolFlag{
+			Name:  flagWithNetwork,
+			Value: true,
+			Usage: "Run the relayer with p2p network capabilities",
+		},
+		&cli.StringFlag{
+			Name:  flagLibp2pKey,
+			Usage: "libp2p private key",
+			Value: fmt.Sprintf("%s/node.key", os.TempDir()),
```
dimalinux, line 103: Doesn't this belong in the the "data dir" like we do in swapd?

cmd/relayer/main.go
noot: yes, the relayer will be on the same network as the swap nodes, they will just support different stream protocols

cmd/relayer/main.go
noot: the relayer is eth-only, so I omitted the "ethereum" part for a few of the flags. do you think I should change the flags, or make it clearer what the relayer is for?

cmd/relayer/main.go
noot: we don't run the relayer HTTP RPC server if this is set to false. there are now 2 ways for the relayer to accept incoming requests, first is through the p2p network, second is through the RPC server. since users will be discovering the relayers via p2p now (most likely), I don't think it's necessary to have swap relayers also run the RPC server by default. 

the requests it's accepting is only the request to submit a transaction.

noot: I'll update the description though

cmd/relayer/validate.go
noot: I'd argue since the code is only used by the relayer server and nothing else, it's fine to be here, but I can move it out to a different package if you'd like. the "validation func" is also passed into the constructor for the relayer, so I'd say it's part of setting up the relayer server

cmd/relayer/main.go
dimalinux: There are 3 endpoints associated with the relayer, even if the flags for the other 2 don't use the phrase "endpoint", so I think having a prefix would be more clear:

1. Ethereum endpoint
2. RPC endpoint
3. libp2p endpoint

cmd/relayer/main.go
dimalinux: I was thinking we might integrate the relayer with swapd, so we could have transaction locking on the wallet. If the relayer is a separate daemon, we'll need to make sure the end user knows that they need to use separate wallet keys for the relayer and any swapd instances.

cmd/relayer/validate.go
noot: I tried using the swapABI but it failed to unpack for some reason :/ I slightly updated the code to unpack into a map, it should be clearer now what the args are

cmd/relayer/main.go
noot: there's a `getOrDeploySwapFactory` inside cmd/swapd that's analogous, I can move this to a diff file to separate it?

cmd/relayer/main.go
noot: updated!

cmd/relayer/main.go
noot: updated

cmd/relayer/main.go
noot: updated

cmd/relayer/main.go
noot: updated

cmd/relayer/validate.go
noot: updated

cmd/swapd/main_test.go
noot: updated

cmd/relayer/main.go
noot: updated

cmd/relayer/main.go
```diff
@@ -0,0 +1,426 @@
...
+	if err != nil {
+		return err
+	}
+
+	if relayerCommission.Cmp(apd.New(1, -1)) > 0 {
+		return errors.New("relayer commission is too high: must be less than 0.1 (10%)")
+	}
+
+	// TODO: do we need to restrict potential commission values? eg. 1%, 1.25%, 1.5%, etc
+	// or should we just require a fixed value for now?
```
dimalinux, line 246: I still like the idea of a fixed-value commission, since the relayer's work is independent of the amount being transferred.  Just like miner fees only depend on the size of the transaction, not the amount of money being transferred.

cmd/relayer/main.go
dimalinux: Yeah, a different file from `main.go` would be good.

cmd/relayer/validate.go
dimalinux: Okay. I like the way that it was broken off into a separate file. If we do ever move it somewhere else in the future, that will make it a lot easier since the code and tests are separate from `main.go`..

cmd/relayer/validate.go
dimalinux: I can take a stab at using the generated code this evening. If nothing else, it would be educational to understand why it didn't work.

cmd/relayer/main.go
```diff
@@ -0,0 +1,426 @@
...
+			Value: "{HOME}/.atomicswap/{ENV}", // For --help only, actual default replaces variables
+		},
+		&cli.StringFlag{
+			Name:  flagEndpoint,
+			Value: "http://localhost:8545",
+			Usage: "Ethereum RPC endpoint",
+		},
+		&cli.StringFlag{
+			Name:  flagKey,
+			Value: "{HOME}/.atomicswap/{ENV}/relayer/eth.key",
```
dimalinux, line 76: Can you update  `docs/default-file-locations.md` with all the new relayer files? For this entry, we should have a note that the wallet key must be different from `swapd`'s ETH wallet key.

cmd/relayer/main.go
```diff
@@ -0,0 +1,426 @@
...
+		},
+		&cli.BoolFlag{
+			Name:  flagWithNetwork,
+			Value: true,
+			Usage: "Run the relayer with p2p network capabilities",
+		},
+		&cli.StringFlag{
+			Name:  flagLibp2pKey,
+			Usage: "libp2p private key",
+			Value: common.DefaultLibp2pKeyFileName,
```
dimalinux, line 110: Can we put this in the data dir like swapd does? Would it be reasonable to use the same location as `swapd` by default, or do we want to keep these identifiers separate?

cmd/relayer/validate_test.go
```diff
@@ -0,0 +1,95 @@
...
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+	"github.com/cockroachdb/apd/v3"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateRelayerFee(t *testing.T) {
+	buf := new(bytes.Buffer)
+	_, err := buf.Write([]byte(contracts.SwapFactoryABI))
```
dimalinux, line 17: New code should be using `contracts.SwapFactoryMetaData.ABI`.

cmd/relayer/validate_test.go
```diff
@@ -0,0 +1,95 @@
...
+	require.NoError(t, err)
+
+	swapABI, err := abi.JSON(buf)
```
dimalinux, line 20: You can just write:
```
swapABI, err := abi.JSON(strings.NewReader(contracts.SwapFactoryMetaData.ABI))
```
and then remove the `bytes.Buffer` above.  There is also `bytes.NewReader` if you ever need to turn a byte slice into a reader.

cmd/relayer/validate.go
```diff
@@ -0,0 +1,156 @@
...
+	relayerCommission *apd.Decimal
+	forwarderAddress  ethcommon.Address
+}
+
+func (v *validator) validateTransactionFunc(req *rcommon.SubmitTransactionRequest) error {
+	// validate that:
+	// 1. the `to` address is a swap contract;
+	// 2. the function being called is `claimRelayer`;
+	// 3. the fee passed to `claimRelayer` is equal to or greater
+	// than our desired commission percentage.
```
dimalinux, line 84: Is there any easy way we can validate that the commission on the swap amount will exceed the commission fees paid by the relayer? (Maybe in a follow-up PR, if not here?)

One idea would be to handle it by setting a minimum fee to use the relayer.  Another option would be to have a minimum ETH amount for a swap. (These options are not mutually exclusive.)  On the monero side, we definitely need to institute a minimum swap amount to avoid ugly support questions. If the swap amount is too small, the monero gets trapped in the A+B wallet, as the wallet doesn't have enough funds to sweep any portion of the swap amount out of it.

cmd/relayer/validate.go
dimalinux: Yeah, I couldn't see how to deserialize a slice of 3 differently typed values using the ABI.  My only thought was that maybe we should separate deserialization from validation so we can easily unit test the deserialization.

noot: I tried loading the `SwapFactoryABI` into an `abi.ABI` type and doing `Unpack` from there, but it failed, which seems weird considering this method *should* be the same and it worked

cmd/relayer/main.go
noot: I agree, I think it might make more sense potentially, let's discuss it. I went with percentages to start off since that's what other relayer networks tended to use

cmd/relayer/main.go
noot: should the default value be `{DATA-DIR}/relayer` actually? for some reason, `{DATA-DIR}` didn't seem to be properly recognized when I tried running this locally

cmd/relayer/main.go
noot: I tried to put it there but it wasn't being recognized when I ran it locally (with the integration tests), is there a trick to getting `{DATA-DIR}` to resolve? I thought it would work with the script since swapd does, probably missing something obvious

cmd/relayer/validate.go
noot: yeah, we'd have to add something in the relayer code that would allow for a gas estimate to happen before actual execution, then we could find the max gas fee from there. however since the call should always take the same amount of gas (for ETH, ERC20s would need to be estimated probably) we can determine the price just by multiplying with the gas price.

I think having a fixed relayer commission could also be a solution. there's not really any "attack" possible on relayers thankfully, since instantiating the swap costs significantly more than calling `claim()` does.

overall, there probably should be a minimum swap amount, the amount should at least be higher than tx fees like you said

cmd/relayer/validate.go
```diff
@@ -0,0 +1,156 @@
...
+
+	err = validateRelayerFee(req.Data[4:], v.relayerCommission)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateRelayerFee(data []byte, minFeePercentage *apd.Decimal) error {
```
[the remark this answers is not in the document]
noot: sure, which part do you want to take out? mainly the `UnpackIntoMap` part?

cmd/relayer/main.go
dimalinux: The values in braces (`{...}`) are like that for the `--help`. The true default value can depend on other flags, so we don't try to resolve it for the purposes of the help. When you later check the value, use `IsSet` to detect when a user has set the flag instead of using `String`. If it is not set, then you resolve the data-dir value and create the path.

cmd/relayer/main.go
dimalinux: See the answer for `flagKey`. You can look at swapd's main to see how we are handling values that depend on the `--data-dir` and `--env` flags.

cmd/relayer/main.go
dimalinux: The worse scenario, that we'd like to avoid is:
* User wants to do large swap and doesn't like the percentage based commission.
* User does tiny swap with a relayer to get gas fees for a subsequent swap. The swap is so small that the relayer loses money on it.
* User now has gas fees to do a large swap without a relayer.

cmd/relayer/validate.go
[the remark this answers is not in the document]
dimalinux: Anything that interacts with `data` could be handled/de-serialized before we call this function is what I was thinking. Decoupling de-serialization from fee validation.

cmd/relayer/main.go
noot: figured it out lol, updated

cmd/relayer/main.go
noot: updated

cmd/relayer/validate.go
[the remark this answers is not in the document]
noot: updated, also updated the test

cmd/relayer/main.go
```diff
@@ -0,0 +1,388 @@
+// Package main is the entrypoint for the swap-specific Ethereum transaction relayer.
+// It's purpose is to allow swaps users (ETH-takers in particular) to submit calls to the
```
dimalinux, line 2: `Its`. Ironically, "its" is only possessive when it doesn't have the apostrophe. :)

noot: updated

dimalinux (approved):

PR #283, opened 2023-01-22
fix: add commas to offer hash preimage

noot:


common/types/offer.go
```diff
@@ -73,10 +73,15 @@ func (o *Offer) setID() {
...
 func (o *Offer) hash() Hash {
 	b := append([]byte(o.Version.String()), []byte(o.Provides)...)
+	b = append(b, []byte(",")...)
 	b = append(b, []byte(o.MinAmount.Text('f'))...)
+	b = append(b, []byte(",")...)
 	b = append(b, []byte(o.MaxAmount.Text('f'))...)
+	b = append(b, []byte(",")...)
 	b = append(b, []byte(o.ExchangeRate.String())...)
+	b = append(b, []byte(",")...)
 	b = append(b, o.EthAsset[:]...)
```
dimalinux, line 83: Can we use the `String()` or `Hex()` method here to be consistent with every other field that is using a text representation.

dimalinux (approved):
Approved, but see the request to make ETH asset consistent with the other fields.

PR #285, opened 2023-01-23
reduce monero wallet contention

dimalinux:
This PR creates a new `monero-wallet-rpc` instance for temporary swap wallets. This eliminates the need for monero wallet locking, as `monero-wallet-rpc` instances now never switch between wallets. They just use one wallet from the time they are created until they are closed. Simultaneous transactions from the same primary wallet for different swaps is safe. We only required the locking to prevent wallet switching.

### DIFF
--- a/coins/coins.go
+++ b/coins/coins.go
@@ -105,6 +105,14 @@ func (a *PiconeroAmount) AsMonero() *apd.Decimal {
 	return xmrAmt
 }
 
+// AsMoneroString converts a PiconeroAmount into a formatted XMR amount string.
+func (a *PiconeroAmount) AsMoneroString() string {
+	xmrAmt := new(apd.Decimal).Set(a.Decimal())
+	decreaseExponent(xmrAmt, NumMoneroDecimals)
+	_, _ = xmrAmt.Reduce(xmrAmt)
+	return xmrAmt.Text('f')
+}
+
 // FmtPiconeroAmtAsXMR takes piconeros as input and produces a formatted string of the
 // amount in XMR.
 func FmtPiconeroAmtAsXMR(piconeros uint64) string {
--- a/coins/coins.go
+++ b/coins/coins.go
@@ -107,16 +107,13 @@ func (a *PiconeroAmount) AsMonero() *apd.Decimal {
 
 // AsMoneroString converts a PiconeroAmount into a formatted XMR amount string.
 func (a *PiconeroAmount) AsMoneroString() string {
-	xmrAmt := new(apd.Decimal).Set(a.Decimal())
-	decreaseExponent(xmrAmt, NumMoneroDecimals)
-	_, _ = xmrAmt.Reduce(xmrAmt)
-	return xmrAmt.Text('f')
+	return a.AsMonero().Text('f')
 }
 
 // FmtPiconeroAmtAsXMR takes piconeros as input and produces a formatted string of the
 // amount in XMR.
 func FmtPiconeroAmtAsXMR(piconeros uint64) string {
-	return NewPiconeroAmount(piconeros).AsMonero().Text('f')
+	return NewPiconeroAmount(piconeros).AsMoneroString()
 }
 
 // WeiAmount represents some amount of ether in the smallest denomination (wei)
--- a/coins/coins.go
+++ b/coins/coins.go
@@ -105,6 +105,12 @@ func (a *PiconeroAmount) AsMonero() *apd.Decimal {
 	return xmrAmt
 }
 
+// FmtPiconeroAmtAsXMR takes piconeros as input and produces a formatted string of the
+// amount in XMR.
+func FmtPiconeroAmtAsXMR(piconeros uint64) string {
+	return NewPiconeroAmount(piconeros).AsMonero().Text('f')
+}
+
 // WeiAmount represents some amount of ether in the smallest denomination (wei)
 type WeiAmount apd.Decimal
 
--- a/monero/test_support.go
+++ b/monero/test_support.go
@@ -60,10 +60,8 @@ func CreateWalletClient(t *testing.T) WalletClient {
 }
 
 // GetBalance is a convenience method for tests that assumes you want the primary
-// address, that you want to refresh, and that errors should fail the test.
+// address and that errors should fail the test.
 func GetBalance(t *testing.T, wc WalletClient) *wallet.GetBalanceResponse {
-	err := wc.Refresh()
-	require.NoError(t, err)
 	balance, err := wc.GetBalance(0)
 	require.NoError(t, err)
 	return balance
@@ -127,7 +125,6 @@ func MineMinXMRBalance(t *testing.T, wc WalletClient, minBalance *coins.Piconero
 	require.NoError(t, err)
 
 	for {
-		require.NoError(t, wc.Refresh())
 		balance, err := wc.GetBalance(0)
 		require.NoError(t, err)
 		if balance.UnlockedBalance > minBalU64 {
--- a/monero/utils.go
+++ b/monero/utils.go
@@ -61,24 +61,22 @@ func CreateWallet(
 	client WalletClient,
 	kpAB *mcrypto.PrivateKeyPair,
 	restoreHeight uint64,
-) (mcrypto.Address, error) {
-	t := time.Now().Format(common.TimeFmtNSecs)
-	walletName := fmt.Sprintf("%s-%s", name, t)
-	if err := client.GenerateFromKeys(kpAB, restoreHeight, walletName, "", env); err != nil {
-		return "", err
-	}
-
-	log.Info("created wallet: ", walletName)
-
-	if err := client.Refresh(); err != nil {
-		return "", err
-	}
-
-	balance, err := client.GetBalance(0)
+) (WalletClient, error) {
+	conf := client.CreateABWalletConf()
+	abCli, err := CreateSpendWalletFromKeys(conf, kpAB, restoreHeight)
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+
+	if err = abCli.Refresh(); err != nil {
+		return nil, err
+	}
+
+	balance, err := abCli.GetBalance(0)
+	if err != nil {
+		return nil, err
 	}
 
 	log.Info("wallet balance: ", balance.Balance)
-	return kpAB.Address(env), nil
+	return abCli, nil
 }
--- a/monero/utils.go
+++ b/monero/utils.go
@@ -21,7 +21,8 @@ var (
 // WaitForBlocks waits for `count` new blocks to arrive.
 // It returns the height of the chain.
 func WaitForBlocks(ctx context.Context, client WalletClient, count int) (uint64, error) {
-	startHeight, err := client.GetChainHeight()
+	c := client.(*walletClient)
+	startHeight, err := c.getChainHeight()
 	if err != nil {
 		return 0, fmt.Errorf("failed to get height: %w", err)
 	}
@@ -29,16 +30,16 @@ func WaitForBlocks(ctx context.Context, client WalletClient, count int) (uint64,
 	endHeight := startHeight + uint64(count)
 
 	for {
-		if err := client.Refresh(); err != nil {
-			return 0, err
-		}
-
-		height, err := client.GetChainHeight()
+		height, err := c.getChainHeight()
 		if err != nil {
 			return 0, err
 		}
 
 		if height >= endHeight {
+			// ensure wallet height is refreshed to the chain height
+			if err = client.(*walletClient).refresh(); err != nil {
+				return 0, err
+			}
 			return height, nil
 		}
 
--- a/monero/utils.go
+++ b/monero/utils.go
@@ -37,7 +37,7 @@ func WaitForBlocks(ctx context.Context, client WalletClient, count int) (uint64,
 
 		if height >= endHeight {
 			// ensure wallet height is refreshed to the chain height
-			if err = client.(*walletClient).refresh(); err != nil {
+			if err = c.refresh(); err != nil {
 				return 0, err
 			}
 			return height, nil
--- a/monero/utils.go
+++ b/monero/utils.go
@@ -5,10 +5,9 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/athanorlabs/atomic-swap/common"
-	mcrypto "github.com/athanorlabs/atomic-swap/crypto/monero"
-
 	logging "github.com/ipfs/go-log"
+
+	"github.com/athanorlabs/atomic-swap/common"
 )
 
 var (
@@ -52,31 +51,4 @@ func WaitForBlocks(ctx context.Context, client WalletClient, count int) (uint64,
 			return 0, err
 		}
 	}
-}
-
-// CreateWallet creates a monero wallet from a private keypair.
-func CreateWallet(
-	name string,
-	env common.Environment,
-	client WalletClient,
-	kpAB *mcrypto.PrivateKeyPair,
-	restoreHeight uint64,
-) (WalletClient, error) {
-	conf := client.CreateABWalletConf()
-	abCli, err := CreateSpendWalletFromKeys(conf, kpAB, restoreHeight)
-	if err != nil {
-		return nil, err
-	}
-
-	if err = abCli.Refresh(); err != nil {
-		return nil, err
-	}
-
-	balance, err := abCli.GetBalance(0)
-	if err != nil {
-		return nil, err
-	}
-
-	log.Info("wallet balance: ", balance.Balance)
-	return abCli, nil
 }
--- a/monero/wallet_client.go
+++ b/monero/wallet_client.go
@@ -412,8 +412,8 @@ func createWalletFromKeys(
 	c.walletAddr = mcrypto.Address(acctResp.Address)
 	if c.walletAddr != address {
 		c.Close()
-		log.Errorf("Provided address %s does not match wallet computed address %s", address, c.walletAddr)
-		return nil, errors.New("monero address mismatch")
+		return nil, fmt.Errorf("provided address %s does not match monero-wallet-rpc computed address %s",
+			address, c.walletAddr)
 	}
 
 	bal, err := c.GetBalance(0)
--- a/monero/wallet_client.go
+++ b/monero/wallet_client.go
@@ -43,7 +43,7 @@ type WalletClient interface {
 	Transfer(to mcrypto.Address, accountIdx uint64, amount *coins.PiconeroAmount) (*wallet.TransferResponse, error)
 	SweepAll(to mcrypto.Address, accountIdx uint64) (*wallet.SweepAllResponse, error)
 	WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Transfer, error)
-	CreateABWalletConf() *WalletClientConf
+	CreateABWalletConf(walletNamePrefix string) *WalletClientConf
 	WalletName() string
 	GetHeight() (uint64, error)
 	GetChainHeight() (uint64, error)
@@ -262,8 +262,8 @@ func (c *walletClient) SweepAll(to mcrypto.Address, accountIdx uint64) (*wallet.
 	})
 }
 
-func (c *walletClient) CreateABWalletConf() *WalletClientConf {
-	walletName := fmt.Sprintf("ab-swap-wallet-%s", time.Now().Format(common.TimeFmtNSecs))
+func (c *walletClient) CreateABWalletConf(walletNamePrefix string) *WalletClientConf {
+	walletName := fmt.Sprintf("%s-%s", walletNamePrefix, time.Now().Format(common.TimeFmtNSecs))
 	walletPath := path.Join(path.Dir(c.conf.WalletFilePath), walletName)
 	conf := &WalletClientConf{
 		Env:                 c.conf.Env,
--- a/monero/wallet_client.go
+++ b/monero/wallet_client.go
@@ -40,7 +40,13 @@ type WalletClient interface {
 	GetAddress(idx uint64) (*wallet.GetAddressResponse, error)
 	PrimaryAddress() mcrypto.Address
 	GetBalance(idx uint64) (*wallet.GetBalanceResponse, error)
-	Transfer(to mcrypto.Address, accountIdx uint64, amount *coins.PiconeroAmount) (*wallet.TransferResponse, error)
+	Transfer(
+		ctx context.Context,
+		to mcrypto.Address,
+		accountIdx uint64,
+		amount *coins.PiconeroAmount,
+		numConfirmations uint64,
+	) (*wallet.Transfer, error)
 	SweepAll(to mcrypto.Address, accountIdx uint64) (*wallet.SweepAllResponse, error)
 	WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Transfer, error)
 	CreateABWalletConf(walletNamePrefix string) *WalletClientConf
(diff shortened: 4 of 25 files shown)


protocol/xmrmaker/swap_state_test.go
```diff
@@ -38,7 +38,7 @@ func newTestSwapStateAndDB(t *testing.T) (*Instance, *swapState, *offers.MockDat
 		types.NewOffer("", new(apd.Decimal), new(apd.Decimal), new(coins.ExchangeRate), types.EthAssetETH),
 		&types.OfferExtra{},
 		xmrmaker.offerManager,
-		coins.NewPiconeroAmount(33),
```
dimalinux, line 41: 33 piconeros won't cover the fees to transfer back in a refund scenario. We should probably have a "minimum-min" XMR offer amount that would prevent funds from getting stranded in a swap wallet, because the amount is too small to transfer to any other wallet.

monero/utils.go
```diff
@@ -53,32 +52,3 @@ func WaitForBlocks(ctx context.Context, client WalletClient, count int) (uint64,
 		}
 	}
 }
-
-// CreateWallet creates a monero wallet from a private keypair.
-func CreateWallet(
```
dimalinux, line 58: The places that were calling this now just call `CreateSpendWalletFromKeys` directly.

protocol/swap/manager.go
```diff
@@ -148,13 +148,15 @@ func (m *manager) GetOngoingSwap(id types.Hash) (Info, error) {
 }
 
 // GetOngoingSwaps returns all ongoing swaps.
-func (m *manager) GetOngoingSwaps() ([]Info, error) {
+func (m *manager) GetOngoingSwaps() ([]*Info, error) {
```
dimalinux, line 151: It's kind of a big data structure to be passing around by value. Some places were passing it by value, others by reference. All the places are passing it by reference now.

monero/test_support.go
```diff
@@ -60,10 +60,8 @@ func CreateWalletClient(t *testing.T) WalletClient {
 }
 
 // GetBalance is a convenience method for tests that assumes you want the primary
-// address, that you want to refresh, and that errors should fail the test.
+// address and that errors should fail the test.
 func GetBalance(t *testing.T, wc WalletClient) *wallet.GetBalanceResponse {
-	err := wc.Refresh()
-	require.NoError(t, err)
```
dimalinux, line 66: `GetBalance` below now does the refresh. I removed `Refresh` from the outside-package wallet API. If someone calls `GetBalance` or `GetHeight`, the refresh is taken care of for them.

monero/wallet_client.go
```diff
@@ -36,37 +36,29 @@ const (
...
+		numConfirmations uint64,
+	) ([]*wallet.Transfer, error)
+	CreateABWalletConf(walletNamePrefix string) *WalletClientConf
+	WalletName() string
 	GetHeight() (uint64, error)
-	GetChainHeight() (uint64, error)
-	Refresh() error
-	CreateWallet(filename, password string) error
-	OpenWallet(filename, password string) error
-	CloseWallet() error
```
dimalinux, line 67: `Lock`, `Unlock`, `Refresh`, `OpenWallet`, `CloseWallet`, `CreateWallet`, `GetChainHeight`, and `WaitForReceipt` were all removed from this package external interface.

monero/wallet_client.go
```diff
@@ -116,33 +105,72 @@ func NewWalletClient(conf *WalletClientConf) (WalletClient, error) {
...
+		conf.MoneroWalletRPCPath, err = getMoneroWalletRPCBin()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if len(conf.MonerodNodes) == 0 {
+		conf.MonerodNodes = common.ConfigDefaultsForEnv(conf.Env).MoneroNodes
+	}
+	validatedNode, err := findWorkingNode(conf.Env, conf.MonerodNodes)
```
dimalinux, line 118: I moved the primary-wallet validation code (above and below) out of `createWalletRPCService` and into this `NewWalletClient` function. The configurations for the A+B wallets are generated by `CreateABWalletConf` and don't need to be re-validated.

monero/wallet_client.go
```diff
@@ -372,17 +546,48 @@ func (c *walletClient) Endpoint() string {
...
 }
 
+// Close kills the monero-wallet-rpc process closing the wallet. It is designed to only be
+// called a single time from a single go process.
 func (c *walletClient) Close() {
-	c.Lock()
-	defer c.Unlock()
-	if c.rpcProcess != nil {
-		p := c.rpcProcess
-		c.rpcProcess = nil
```
dimalinux, line 380: We are no longer using locks, so I decided to remove that line and add a comment that `Close()` should only be called once. `c.rpcProcess` will only be nil if it was never assigned. Having said that, if you call `Close()` twice, nothing bad should happen other than a warning log that `StopWallet` failed.

protocol/xmrmaker/swap_state_test.go
dimalinux: Created an issue for it: https://github.com/AthanorLabs/atomic-swap/issues/286

coins/coins.go
```diff
@@ -105,6 +105,20 @@ func (a *PiconeroAmount) AsMonero() *apd.Decimal {
 	return xmrAmt
 }
 
+// AsMoneroString converts a PiconeroAmount into a formatted XMR amount string.
+func (a *PiconeroAmount) AsMoneroString() string {
+	xmrAmt := new(apd.Decimal).Set(a.Decimal())
+	decreaseExponent(xmrAmt, NumMoneroDecimals)
+	_, _ = xmrAmt.Reduce(xmrAmt)
+	return xmrAmt.Text('f')
```
dimalinux, line 113: TODO: This function should just be `return a.AsMonero().Text('f')` and the function below should call this one.  I'll fix it before merging.

monero/utils.go
```diff
@@ -22,24 +21,25 @@ var (
 // WaitForBlocks waits for `count` new blocks to arrive.
 // It returns the height of the chain.
 func WaitForBlocks(ctx context.Context, client WalletClient, count int) (uint64, error) {
-	startHeight, err := client.GetChainHeight()
+	c := client.(*walletClient)
```
noot, line 24: is `*walletClient` the only implementation of `WalletClient`?

monero/utils.go
```diff
@@ -22,24 +21,25 @@ var (
...
-
-		height, err := client.GetChainHeight()
+		height, err := c.getChainHeight()
 		if err != nil {
 			return 0, err
 		}
 
 		if height >= endHeight {
+			// ensure wallet height is refreshed to the chain height
+			if err = client.(*walletClient).refresh(); err != nil {
```
noot, line 40:
```suggestion
			if err = c.refresh(); err != nil {
```

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+			coins.FmtPiconeroAmtAsXMR(receipt.Amount),
+			coins.FmtPiconeroAmtAsXMR(receipt.Fee),
+			receipt.Height,
+		)
+		transfers = append(transfers, receipt)
+	}
+	return transfers, nil
+}
+
+func (c *walletClient) CreateABWalletConf(walletNamePrefix string) *WalletClientConf {
```
noot, line 354:
```suggestion
func (c *walletClient) CreateWalletConf(walletNamePrefix string) *WalletClientConf {
```
since there's nothing in the function that's A+B specific?

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+		return nil, fmt.Errorf("transfer failed: %w", err)
+	}
+	log.Infof("Transfer of %s XMR initiated, TXID=%s", amountStr, reqResp.TxHash)
+	transfer, err := c.waitForReceipt(&waitForReceiptRequest{
+		Ctx:              ctx,
+		TxID:             reqResp.TxHash,
+		NumConfirmations: numConfirmations,
+		AccountIdx:       accountIdx,
+	})
+	log.Warnf("Transfer TXID=%s failed to get receipt: %s", reqResp.TxHash, err)
```
noot, line 283: do we need a warn here since we also return an error?

noot: also this should be in the `if err != nil` block

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	from := addrResp.Address
+	balance, err := c.GetBalance(accountIdx)
+	if err != nil {
+		return nil, fmt.Errorf("sweep operation failed to get balance: %w", err)
+	}
+	log.Infof("Starting sweep of %s XMR from %s to %s", coins.FmtPiconeroAmtAsXMR(balance.Balance), from, to)
+	if balance.BlocksToUnlock > 0 {
+		log.Infof("Sweep operation waiting %d blocks for balance to fully unlock", balance.BlocksToUnlock)
+		if _, err = WaitForBlocks(ctx, c, int(balance.BlocksToUnlock)); err != nil {
+			log.Warnf("Sweep operation from %s failed waiting to unlock balance", from)
```
noot, line 314: same here, probably don't need a warn cause we return an error

noot: same with others in this func

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	}
+	reqResp, err := c.wRPC.SweepAll(&wallet.SweepAllRequest{
 		AccountIndex: accountIdx,
 		Address:      string(to),
 	})
+	if err != nil {
+		log.Warnf("Sweep transaction(s) from %s failed, %s", from, err)
+		return nil, fmt.Errorf("sweep_all failed: %w", err)
+	}
+	log.Infof("Sweep transaction started, TX IDs: %s", strings.Join(reqResp.TxHashList, ", "))
```
noot, line 330: didn't know SweepAll could return mutliple txs, any idea what the diff txs are doing?

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+			log.Warnf("Sweep TX ID %s failed waiting for receipt: %s", txID, err)
+			return nil, fmt.Errorf("sweep failed waiting for receipt: %w", err)
+		}
+		log.Infof("Sweep transfer ID=%s of %s XMR (%s XMR fees) completed at height %d",
+			txID,
+			coins.FmtPiconeroAmtAsXMR(receipt.Amount),
+			coins.FmtPiconeroAmtAsXMR(receipt.Fee),
+			receipt.Height,
+		)
+		transfers = append(transfers, receipt)
```
noot, line 349: put this in a go func maybe? might make error handling more annoying, but then it can return on failure immediately if any of the txs fails without having to wait for the earlier ones to complete

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+		return nil, err
+	}
+	acctResp, err := c.GetAddress(0)
+	if err != nil {
+		c.Close()
+		return nil, err
+	}
+	c.walletAddr = mcrypto.Address(acctResp.Address)
+	if c.walletAddr != address {
+		panic("addresses do not match")
```
noot, line 419: this panic seems risky, is there any way the address passed might not match that generated?

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+		return nil, err
+	}
+
+	log.Infof("Created wallet %s, balance is %s XMR (%d blocks to unlock), address is %s",
+		c.WalletName(),
+		coins.FmtPiconeroAmtAsXMR(bal.Balance),
+		bal.BlocksToUnlock,
+		c.PrimaryAddress(),
+	)
+	return c, err
```
noot, line 434:
```suggestion
	return c, nil
```

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
 }
 
-// GenerateFromKeys creates a wallet from a given wallet address, view key, and optional spend key
-func (c *walletClient) GenerateFromKeys(
-	kp *mcrypto.PrivateKeyPair,
+// CreateSpendWalletFromKeys creates a new monero-wallet-rpc process and wallet from a given wallet address,
+// view key, and optional spend key
```
noot, line 438: is the spend key optional? shouldn't `CreateViewOnlyWalletFromKeys` be used in that case

noot: also does `*mcrypto.PrivateKeyPair` allow for nil spend keys?

monero/wallet_client.go
```diff
@@ -372,17 +550,48 @@ func (c *walletClient) Endpoint() string {
...
+		log.Warnf("StopWallet errored: %s", err)
+		err = p.Kill() // uses, SIG-TERM, which monero-wallet-rpc has a handler for
+	}
+	// If err is nil at this point, the process existed, and we block until the child
+	// process exits. (Note: kill does not error when signaling an exited, but non-reaped
+	// child.)
+	if err == nil {
+		_, _ = p.Wait()
+	}
+
```
noot, line 571: add a `c.rpcProcess == nil` in case this function is ever called twice? is there ever a chance of that happening?

protocol/xmrmaker/swap_state.go
```diff
@@ -424,9 +425,28 @@ func (s *swapState) reclaimMonero(skA *mcrypto.PrivateSpendKey) (mcrypto.Address
...
+	)
+	if err != nil {
+		return "", fmt.Errorf("failed to sweep refund back to primary wallet: %w", err)
+	}
+	for _, transfer := range transfers {
+		// It is unlikely that anyone will ever see more than one of these messages
+		log.Infof("Refunded %s XMR from swap to primary wallet (%s XMR lost to fees)",
+			coins.FmtPiconeroAmtAsXMR(transfer.Amount), coins.FmtPiconeroAmtAsXMR(transfer.Fee))
+	}
+	return kpAB.Address(s.Env()), nil
```
noot, line 449: this seems to be basically the same code as `func (inst *Instance) recoverRefund`, can you update these two to use the same function

protocol/xmrtaker/claim.go
```diff
@@ -86,24 +88,27 @@ func (s *swapState) claimMonero(skB *mcrypto.PrivateSpendKey) (mcrypto.Address,
...
-	s.XMRClient().Lock()
-	defer s.XMRClient().Unlock()
-
-	addr, err := monero.CreateWallet("xmrtaker-swap-wallet", s.Env(), s.Backend.XMRClient(), kpAB, s.walletScanHeight)
+	conf := s.XMRClient().CreateABWalletConf("xmrtaker-swap-wallet-claim")
+	abWalletCli, err := monero.CreateSpendWalletFromKeys(conf, kpAB, s.walletScanHeight)
 	if err != nil {
 		return "", err
 	}
+	defer abWalletCli.CloseAndRemoveWallet()
```
noot, line 104: remove this, it's covered in the if/else below

protocol/xmrtaker/claim.go
```diff
@@ -113,44 +118,49 @@ func (s *swapState) claimMonero(skB *mcrypto.PrivateSpendKey) (mcrypto.Address,
...
 	log.Infof("monero claimed in account %s; transferring to original account %s",
-		addr, depositAddr)
+		abAddr, depositAddr)
 
 	err = mcrypto.ValidateAddress(string(depositAddr), s.Env())
 	if err != nil {
-		log.Errorf("failed to transfer to original account, address %s is invalid", addr)
-		return addr, nil
+		log.Errorf("failed to transfer to original account, address %s is invalid", abAddr)
+		return abAddr, nil
```
noot, line 126: wait why do we just log here instead of returning an error?

protocol/xmrtaker/instance.go
```diff
@@ -100,22 +100,36 @@ func (inst *Instance) createOngoingSwap(s swap.Info) error {
...
-		inst.backend.XMRClient().Lock()
-		defer inst.backend.XMRClient().Unlock()
-
-		// TODO: do we want to transfer this back to the original account?
-		addr, err := monero.CreateWallet(
-			"xmrmaker-swap-wallet",
-			inst.backend.Env(),
-			inst.backend.XMRClient(),
-			kp,
+		conf := inst.backend.XMRClient().CreateABWalletConf("xmrtaker-swap-wallet-db-restored")
```
noot, line 103: maybe put everything in this `if err == nil` block into its own func? there's a lot going on

protocol/xmrtaker/instance.go
```diff
@@ -100,22 +100,36 @@ func (inst *Instance) createOngoingSwap(s swap.Info) error {
...
+		abWalletCli, err := monero.CreateSpendWalletFromKeys(
+			conf,
+			kp, // TODO: Fix the key here?
```
noot, line 106: yes!

noot (commented):
overall looks good! mostly smaller comments and suggestions

monero/utils.go
dimalinux: Yes. I could even see an argument for removing the interface and capitalizing the struct type.

monero/wallet_client.go
dimalinux: If there are too many outputs, it can take more than one transaction to sweep them all. We need to handle it, but it is unlikely that there will ever be more than one.

protocol/xmrtaker/claim.go
dimalinux: I'll update it to return the error.

monero/wallet_client.go
dimalinux: I'll add an error level log and return an error. A mismatched address should have been detected long before we get here, but better to be safe and not have the server crash.

monero/utils.go
noot: yeah I'd be fine with that. or move to interface to the packages where it's used if you want to do any mock testing

monero/wallet_client.go
dimalinux: The code in this PR will never assign `rpcProcess` nil. The only way it can be nil, is if it was never assigned from the start (handled at the beginning of the function). There is nowhere that should be calling `Close()` more than once, but if it happens by accident, the 2nd `StopWallet()` will fail causing the `p.Kill()` line above to execute. Then the `Kill()` would also fail, as the process ID has already been reaped by `Wait()`, so we won't call `Wait()` a second time. What I describe above is the scenario where one `Close()` completed before the next one started. Even if both `Close()` calls were executed simultaneously, the above code should be safe from any adverse behavior.

monero/wallet_client.go
dimalinux: In all likelihood, there will only be one transaction. Even if there is more than one transaction, it will most likely end up in the same block, so waiting for the 2nd receipt won't block. To prevent `SweepAll` calls from blocking for too long, I made the compromise of setting the number of confirmations to 2 (verses 10 for the transfer that Bob does to the A+B wallet).

monero/utils.go
dimalinux: Let's play with the idea in a different PR. I think it is too big of a change to add into this PR at this point.

monero/wallet_client.go
dimalinux: `CreateSpendWalletFromKeys` only takes a `PrivateKeyPair`. I don't think your design for `PrivateKeyPair` ever intended the type to allow the view key to be unset.  `NewPrivateKeyPairFromBytes` will throw an error, for example, if the view key is not set. We could document that and maybe panic if a nil value is passed to `NewPrivateKeyPair`. My intention was that package external users would only use `CreateSpendWalletFromKeys` for spend wallets and only use `CreateViewOnlyWalletFromKeys` for view-only wallets.

protocol/xmrmaker/swap_state.go
dimalinux: I wanted to do that originally, and it is part of the reason I realized that `SweepAll` should be handling the receipt (which it now does). The issue was that these functions are for objects of different types. One is an `Instance` member function, while the other is a `swapState` function. I made a procedural (non-object-oriented) function named `sweepRefund` and put it in `xmrmaker/utils.go` so the code could be shared. It is a bit of a monster, in that it requires 7 parameters, but it was good to consolidate the code.

protocol/xmrtaker/instance.go
dimalinux: This is the code that is generating the view key from the spend key. It needs a major rewrite and better unit test coverage in a follow-up PR. Let's break it up, but do it in the next PR.

monero/wallet_client.go
noot: okay, I guess update the comment then to say the spend key is required?

monero/wallet_client.go
noot: @dimalinux this shouldn't be logging here?

monero/wallet_client.go
```diff
@@ -221,49 +250,211 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	reqResp, err := c.wRPC.Transfer(&wallet.TransferRequest{
 		Destinations: []wallet.Destination{{
 			Amount:  amt,
 			Address: string(to),
 		}},
 		AccountIndex: accountIdx,
 	})
+	if err != nil {
+		log.Warnf("Transfer of %s XMR failed: %s", amountStr, err)
+		return nil, fmt.Errorf("transfer failed: %w", err)
```
noot, line 274: don't think we need the warn log and returned error?

monero/wallet_client.go
noot: @dimalinux thoughts?

monero/wallet_client.go
noot: I don't see this as updated?

dimalinux: That code was updated and pushed on the 26th:
https://github.com/AthanorLabs/atomic-swap/blame/dimalinux/reduce-xmr-wallet-contention/monero/wallet_client.go#L412-L417

monero/wallet_client.go
dimalinux: The warning message was removed days before your last comment. That's why I marked it as resolved:
https://github.com/AthanorLabs/atomic-swap/blame/dimalinux/reduce-xmr-wallet-contention/monero/wallet_client.go#L312-L314

monero/wallet_client.go
dimalinux: Those changes were pushed before well before this thread was resolved the first time:
https://github.com/AthanorLabs/atomic-swap/commit/91af4ae32337f1d28d79540b93bcb161c07c9325

monero/wallet_client.go
dimalinux: Obviously, it can be taken too far, and I removed a lot of warnings based on your feedback, but if something super bad happens that must be logged, it is good to have some possibly redundant logging closest to the failure point, especially if the early points have additional context they can provide that would be too verbose pass up the error chain.

For errors that relate to things that might cause loss of funds (or losing track of funds that might be recoverable), I think we want as much context as possible and minor redundancy doesn't hurt.

monero/wallet_client.go
```diff
@@ -221,49 +250,208 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	addrResp, err := c.GetAddress(accountIdx)
+	if err != nil {
+		return nil, fmt.Errorf("sweep operation failed to get address: %w", err)
+	}
+	from := addrResp.Address
+	balance, err := c.GetBalance(accountIdx)
+	if err != nil {
+		return nil, fmt.Errorf("sweep operation failed to get balance: %w", err)
+	}
+	log.Infof("Starting sweep of %s XMR from %s to %s", coins.FmtPiconeroAmtAsXMR(balance.Balance), from, to)
```
noot, line 309:
```suggestion

	log.Infof("Starting sweep of %s XMR from %s to %s", coins.FmtPiconeroAmtAsXMR(balance.Balance), from, to)
```

monero/wallet_client.go
```diff
@@ -221,49 +250,208 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	ctx context.Context,
+	to mcrypto.Address,
+	accountIdx uint64,
+	numConfirmations uint64,
+) ([]*wallet.Transfer, error) {
+	addrResp, err := c.GetAddress(accountIdx)
+	if err != nil {
+		return nil, fmt.Errorf("sweep operation failed to get address: %w", err)
+	}
+	from := addrResp.Address
```
noot, line 304:
```suggestion

	from := addrResp.Address
```

monero/wallet_client.go
```diff
@@ -221,49 +250,208 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	if balance.BlocksToUnlock > 0 {
+		log.Infof("Sweep operation waiting %d blocks for balance to fully unlock", balance.BlocksToUnlock)
+		if _, err = WaitForBlocks(ctx, c, int(balance.BlocksToUnlock)); err != nil {
+			return nil, fmt.Errorf("sweep operation failed waiting to unlock balance: %w", err)
+		}
+	}
+	if balance.Balance == 0 {
+		return nil, fmt.Errorf("sweep from %s failed, no balance to sweep", from)
+	}
+	reqResp, err := c.wRPC.SweepAll(&wallet.SweepAllRequest{
```
noot, line 319:
```suggestion

	reqResp, err := c.wRPC.SweepAll(&wallet.SweepAllRequest{
```

nit but can you add spaces after all the `if err != nil` and other error checks in this file? it's harder to read when everything is bunched together

monero/wallet_client.go
```diff
@@ -221,49 +250,208 @@ func (c *walletClient) WaitForReceipt(req *WaitForReceiptRequest) (*wallet.Trans
...
+	log.Infof("Starting sweep of %s XMR from %s to %s", coins.FmtPiconeroAmtAsXMR(balance.Balance), from, to)
+	if balance.BlocksToUnlock > 0 {
+		log.Infof("Sweep operation waiting %d blocks for balance to fully unlock", balance.BlocksToUnlock)
+		if _, err = WaitForBlocks(ctx, c, int(balance.BlocksToUnlock)); err != nil {
+			return nil, fmt.Errorf("sweep operation failed waiting to unlock balance: %w", err)
+		}
+	}
+	if balance.Balance == 0 {
+		return nil, fmt.Errorf("sweep from %s failed, no balance to sweep", from)
+	}
```
noot, line 318: nit: this check could be above the previous check, would be slighter cleaner imo

noot (approved):
looks good, please resolve minor comments before merging

PR #291, opened 2023-01-29
feat: add relayer discovery and transaction submission via network 

### LINKED ISSUE #209: relayer support: add relayer discovery via network

noot:
- add relayer network host to protocol backend; add functions for relayer discovery and submission of tx messages
- update xmrmaker's claim functionality to discover and submit via relayer if their eth wallet is empty
- todo: add integration tests

for the future, it would be useful to have a cache of recently used relayers (that succeeded), then for future txs, use relayers from this cache, as we know they've been successful in the past. 

closes #209


cmd/swapd/main.go
```diff
@@ -433,6 +434,7 @@ func (d *daemon) make(c *cli.Context) error { //nolint:gocyclo
 		return err
 	}
 	d.host = host
+	rhost := setupRelayerNetwork(d.ctx, host.P2pHost())
```
dimalinux, line 437: Not critical: I like terse variable names in short functions, but in a 181 line function should we call this relayHost or relayerHost to avoid confusion with routedHost?

coins/coins.go
```diff
@@ -180,7 +180,7 @@ func (a *WeiAmount) BigInt() *big.Int {
 	}
 	if cond.Rounded() {
 		// We round when converting from Ether to Wei, so we shouldn't see this
-		log.Warn("Converting WeiAmount=%s to big.Int required rounding", a)
+		log.Warn("converting WeiAmount=%s to big.Int required rounding", a.String())
```
dimalinux, line 183: I thought we were using the convention that error messages start with a lower case letter and log messages start with an upper case letter. `String()` is called automatically when using `%s`.

net/host.go
```diff
@@ -74,6 +74,11 @@ func NewHost(cfg *p2pnet.Config) (*Host, error) {
 	}, nil
 }
 
+// P2pHost returns the underlying go-p2p-net host.
+func (h *Host) P2pHost() P2pnetHost {
```
dimalinux, line 78: Does the word "Net" add anything to that interface type name? I was going to suggest a capital `N`, since the source package, `go-p2p-net`, is treating them as separate words, but if the function is called `P2pHost`, maybe `net` isn't adding any value to the type name either.

coins/coins.go
dimalinux: Whatever we decide, we should probably harmonize all the logs to use the same standard. I'd been using first letter capitalized in log messages due to previous places of employment requiring it.

protocol/xmrmaker/claim.go
```diff
@@ -65,15 +68,21 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
 		if err != nil {
-			return ethcommon.Hash{}, err
+			log.Warnf("failed to claim using relayer at %s, trying relayers via p2p: err: %s",
+				s.offerExtra.RelayerEndpoint,
+				err,
+			)
+			txHash, err = s.discoverRelayersAndClaim()
 		}
+	} else if ethBalance.Uint64() == 0 {
+		txHash, err = s.discoverRelayersAndClaim()
```
dimalinux, line 78: This logic is going to be difficult to document, but it is the type of information that end users will want to know. 

I'll try to describe what I think I'm reading. Let me know what I get right/wrong:
* To use a relayer, you should specify a non p2p http or https endpoint, or you should have an ethereum balance of exactly zero.
* If the specified non p2p relayer does not work, or if you did not specify a relayer and have a zero ethereum balance, all p2p relayers will be tried until one works.
* If you have a low ethereum balance and want to use a relayer, you must find one with an http or https endpoint and specify it when making the offer.

protocol/xmrmaker/claim.go
```diff
@@ -99,6 +108,53 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
 }
 
+// discoverRelayersAndClaim discovers available relayers on the network,
+func (s *swapState) discoverRelayersAndClaim() (ethcommon.Hash, error) {
+	relayers, err := s.Backend.DiscoverRelayers()
+	if err != nil {
+		return ethcommon.Hash{}, err
+	}
+
+	forwarderAddress, err := s.Contract().TrustedForwarder(&bind.CallOpts{})
```
dimalinux, line 118: `&bind.CallOpts{Context: s.ctx}`  (otherwise you could just pass `nil` and get the same result).

protocol/xmrmaker/claim.go
```diff
@@ -38,12 +40,13 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
 		}
 	}
 
+	ethBalance, err := s.ETHClient().Balance(s.ctx)
```
dimalinux, line 43: I would call this `weiBalance` to avoid confusion.  Or you could convert it into an ETH balance out here instead of just in the log on line 49. One of the improvements that `apd.Decimal` provides over `big.Int` is an `IsZero` function.

protocol/xmrmaker/claim.go
```diff
@@ -65,15 +68,21 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
 		txHash, err = s.claimRelayer()
 		if err != nil {
-			return ethcommon.Hash{}, err
+			log.Warnf("failed to claim using relayer at %s, trying relayers via p2p: err: %s",
+				s.offerExtra.RelayerEndpoint,
+				err,
+			)
+			txHash, err = s.discoverRelayersAndClaim()
 		}
+	} else if ethBalance.Uint64() == 0 {
```
dimalinux, line 77: The maximum amount of ETH that you can represent as wei in `uint64` is a little under 18.5. The documentation for `Uint64()` says the value is undefined if the value is too large.
```
$ python3 -c "print((2**64-1)/1E18)"
18.446744073709553
```

protocol/xmrmaker/claim.go
```diff
@@ -99,6 +108,53 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
+		s.ETHClient().Raw(),
+		forwarderAddress,
+		s.contractAddr,
+		calldata,
+	)
+	if err != nil {
+		return ethcommon.Hash{}, err
+	}
+
+	for _, relayer := range relayers {
```
dimalinux, line 139: Do we know how these relayers are sorted?

protocol/xmrmaker/claim.go
```diff
@@ -135,41 +191,63 @@ func claimRelayer(
...
 	// wait for inclusion
 	receipt, err := block.WaitForReceipt(ctx, ec, txHash)
 	if err != nil {
-		return ethcommon.Hash{}, err
+		return err
 	}
 
 	if receipt.Status == 0 {
-		return ethcommon.Hash{}, fmt.Errorf("transaction failed")
+		return fmt.Errorf("transaction failed")
```
dimalinux, line 220: `WaitForReceipt` above already did this status check: https://github.com/AthanorLabs/atomic-swap/blob/master/ethereum/block/wait_for_receipt.go#L44

protocol/xmrmaker/claim.go
```diff
@@ -99,6 +108,53 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
+	}
+
+	for _, relayer := range relayers {
+		resp, err := s.Backend.SubmitTransactionToRelayer(relayer, req)
+		if err != nil {
+			log.Warnf("failed to submit tx to relayer with peer ID %s, trying next relayer: err: %s", relayer, err)
+			continue
+		}
+
+		err = waitAndCheck(s.ctx, s.ETHClient().Raw(), resp.TxHash)
```
dimalinux, line 146: Is it expected that the relayer waited for at least one confirmation before the request returns? How long do we wait for a receipt? We may need to give up much sooner than we would when getting the receipt for a transaction that we submitted ourselves.

coins/coins.go
noot: ah I need to change this to `Warnf` lol

I'm used to not capitalizing log messages, and I agree we should pick one and stick with it (I vote for lowercase) but as long as it's consistent that's what's important

protocol/xmrmaker/claim.go
noot: yep that's right, also if you put some random string in `relayerEndpoint` it'll also fallback to discovering/trying relayers. I was thinking of adding another option inside the RPC `MakeOfferRequest` that's just a bool `useRelayer` so that the flow is more explicit (and maybe removing the HTTP endpoint, like I mentioned to you before), what do you think?

protocol/xmrmaker/claim.go
noot: they're not sorted, it's effectively random. the real answer is it's in order of records returned to us when we do the dht lookup.

protocol/xmrmaker/claim.go
dimalinux: Yeah, I don't see any advantage in supporting remote https relayers when everything else, besides localhost communication, is using libp2p.  If we only support libp2p relayers, would the flag be `--relayer-peer-id`?  And that flag would automatically set `--use-relayer` to `true` if it was not already set?  (Just tossing out ideas, I haven't thought about it for more than a minute.)

net/host.go
noot: yeah I wasn't sure what to call this either, I felt `P2pNetHost` would be too redundant

protocol/xmrmaker/claim.go
```diff
@@ -65,15 +68,21 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
 		// TODO: eventually update when relayer discovery is implemented
```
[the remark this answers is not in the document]
noot: no, unless we decide to remove the `relayerEndpoint` logic

protocol/xmrmaker/claim.go
noot: yeah, currently relayer options are set per-offer, but I could see having the daemon have a `--use-relayer` flag be useful. and yeah we could change the endpoint to a libp2p endpoint (or array of endpoints)

protocol/xmrmaker/claim.go
noot: no, the relayer doesn't wait at all, it just submits the tx to the mempool.

dimalinux: I asked the question above, because I wanted to understand how we handle the scenario where the relayer submits the transaction, but sets the gas price so low that the transaction won't get picked up quickly, or might never get mined.

scripts/run-integration-tests.sh
```diff
@@ -35,10 +35,9 @@ create-eth-keys() {
 		echo "${GANACHE_KEYS[${i}]}" >"${key_file}"
 	done
 }
-
 # This is the local multiaddr created when using ./tests/alice-libp2p.key on the default libp2p port
-ALICE_MULTIADDR=/ip4/127.0.0.1/udp/9933/quic-v1/p2p/12D3KooWAAxG7eTEHr2uBVw3BDMxYsxyqfKvj3qqqpRGtTfuzTuH
-ALICE_LIBP2PKEY=./tests/alice-libp2p.key
```
dimalinux, line 41: The variable name and key were renamed to "relayer", but it is referenced 3 times as Alice's key and only once as the relayer's key. Would it make sense to use separate keys?

tests/relayer_integration_test.go
```diff
@@ -24,3 +29,22 @@ func (s *IntegrationTestSuite) TestERC20_Success_ClaimRelayer() {
...
+
+func (s *IntegrationTestSuite) TestXMRMaker_DiscoverRelayer() {
+	ctx := context.Background()
+	c := rpcclient.NewClient(ctx, defaultXMRMakerSwapdEndpoint)
+	time.Sleep(time.Second * 30)
+
+	// see https://github.com/AthanorLabs/go-relayer/blob/master/net/host.go#L20
+	peerIDs, err := c.Discover("", defaultDiscoverTimeout)
+	require.NoError(s.T(), err)
+	s.T().Log(len(peerIDs))
```
dimalinux, line 41: If the log is valuable, can you change it to `Logf` and add a context message?

noot: it's not, removed

tests/relayer_integration_test.go
```diff
@@ -24,3 +29,22 @@ func (s *IntegrationTestSuite) TestERC20_Success_ClaimRelayer() {
 		relayerCommission,
 	)
 }
+
+func (s *IntegrationTestSuite) TestXMRMaker_DiscoverRelayer() {
+	ctx := context.Background()
+	c := rpcclient.NewClient(ctx, defaultXMRMakerSwapdEndpoint)
+	time.Sleep(time.Second * 30)
```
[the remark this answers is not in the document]
noot: added a comment about what I explained to you on the call

protocol/xmrmaker/claim.go
noot: right, in that case we can try all the relayers concurrently - I think we discussed this but didn't fully come to a conclusion? if the relayer has enough of a reward for the small amount of gas lost to potentially reverted txs (which they should), I think that's a decent solution

scripts/run-integration-tests.sh
noot: @dimalinux let me know if you wanna take a stab at the script like we discussed

the new flow would be:

- Alice starts first, deploys the forwarder contract and then swap contract which references that
- Bob/Charlie/relayer can all start after, Bob/Charlie use the swap contract deployed by Alice and the relayer uses the forwarder contract deployed by Alice

I didn't update it b/c I didn't want to mess with the log parsing, but feel free to take a stab at it.

net/host.go
noot: updated to `P2pHost`

protocol/xmrmaker/claim.go
dimalinux: That could work. There is still the question of the first attempt though, if we don't go to all of them immediately.

Our existing `WaitForReceipt` method inside `waitAndCheck` will block for 12 minutes at minimum before failing. It would be nice to fail much faster in 2 scenarios:
* Relayer gives us a made up, random transaction ID
* Relayer set the gas price so low that the transaction won't get mined (or won't get mined anytime soon)

tests/integration_test.go
```diff
@@ -221,7 +221,8 @@ func (s *IntegrationTestSuite) testSuccessOneSwap(
 	awsc := s.newSwapdWSClient(ctx, defaultXMRTakerSwapdWSEndpoint)
 
 	// TODO: implement discovery over websockets (#97)
-	peerIDs, err := ac.Discover(coins.ProvidesXMR, defaultDiscoverTimeout)
+	time.Sleep(time.Second * 3)
```
dimalinux, line 224: Can you add a short comment stating what we are waiting for?

dimalinux (approved):
Approved. We'll need follow-up work for the relayer receipt handling.

protocol/xmrmaker/claim.go
noot: @dimalinux I've updated the `waitAndCheck` function so it now checks to see if the tx is in the mempool
- if it's not in the mempool in 6s, it errors
- if it's not included in a block in 1min, it errors
let me know what you think

cmd/swapcli/main.go
```diff
@@ -398,12 +398,8 @@ func runBalances(ctx *cli.Context) error {
 }
 
 func runDiscover(ctx *cli.Context) error {
-	provides, err := providesStrToVal(ctx.String(flagProvides))
-	if err != nil {
-		return err
-	}
-
 	c := newRRPClient(ctx)
+	provides := ctx.String(flagProvides)
```
dimalinux, line 402: Do we need a `swapcli` feature to discover other relayers?

protocol/xmrmaker/claim.go
```diff
@@ -135,41 +193,93 @@ func claimRelayer(
...
+		// into the node we're using
+		err := common.SleepWithContext(ctx, waitDuration)
+		if err != nil {
+			return err
+		}
+
+		_, isPending, err := ec.TransactionByHash(ctx, txHash)
+		if err != nil {
+			// allow up to 5 NotFound errors, in case there's some network problems
+			if errors.Is(err, ethereum.NotFound) && notFoundCount != maxNotFound {
```
dimalinux, line 232: Safe programming practices recommend using `>=` in the 2nd half of the condition above.

protocol/xmrmaker/claim.go
```diff
@@ -135,41 +193,93 @@ func claimRelayer(
...
 		return ethcommon.Hash{}, err
 	}
 
-	txHash, err := rc.SubmitTransaction(contractAddr, calldata)
+	return txHash, nil
+}
+
+func waitAndCheck(ctx context.Context, ec *ethclient.Client, txHash ethcommon.Hash) error {
+	waitDuration := time.Second
+	maxNotFound := 5
```
dimalinux, line 216: You can declare these 2 as constants and add a 3rd constant for the total wait.  Here's what I came up with:
```
	const (
		checkInterval = time.Second // time between transaction polls
		maxWait       = time.Minute // max wait for first TX confirmation
		maxNotFound   = 5           // max failures where the TX is not even found in the mempool
	)
```

protocol/xmrmaker/claim.go
```diff
@@ -135,41 +193,93 @@ func claimRelayer(
...
+	err = waitAndCheck(ctx, ec, txHash)
 	if err != nil {
 		return ethcommon.Hash{}, err
 	}
 
-	txHash, err := rc.SubmitTransaction(contractAddr, calldata)
+	return txHash, nil
+}
+
+func waitAndCheck(ctx context.Context, ec *ethclient.Client, txHash ethcommon.Hash) error {
```
dimalinux, line 214: This is a large package making the scope of this function large.  What do you think of the name `waitForClaimReceipt` or something similar?

dimalinux: How do we know that the transaction hash that the relayer gave us was for our claim?

tests/integration_test.go
dimalinux: How about?
```
// Give the maker's p2p offer advertisement time to propagate
time.Sleep(time.Second * 3)
```
I'm going to try to describe the full problem, that doesn't need to be in the comment, and you can tell me where I'm wrong:

When the maker creates an offer, he then advertises in the libp2p distributed hash table that he has one or more offers. This can take a few seconds to propagate even when all the peers are on the same localhost.

protocol/xmrmaker/claim.go
dimalinux: Looks good.  I added some suggestions on how we might improve the code and a question on how we know that the transaction the relayer gave us is for our claim.

cmd/swapcli/main.go
noot: if you pass the "isrelayer" string into the discover `provides` field, it'll return relayers. we can document that, I just don't expect it to be something users will use

protocol/xmrmaker/claim.go
noot: I can parse the tx calldata field to check it's ours, will add

noot: @dimalinux updated so that the `Claimed` log is checked

protocol/xmrmaker/claim.go
noot: done

protocol/xmrmaker/claim.go
noot: done

tests/integration_test.go
noot: comment updated

protocol/xmrmaker/claim.go
noot: it's not really useful to parse the calldata, there's too many fields and the tx might revert anyways, better to check the logs emitted from the tx

protocol/xmrmaker/claim.go
```diff
@@ -135,41 +202,134 @@ func claimRelayer(
...
-	txHash, err := rc.SubmitTransaction(contractAddr, calldata)
+	return txHash, nil
+}
+
+func waitForClaimReceipt(
+	ctx context.Context,
+	ec *ethclient.Client,
+	txHash ethcommon.Hash,
+	contractAddr ethcommon.Address,
+	contractSwapID, secret [32]byte,
```
dimalinux, line 235: Maybe we should start using `gofumpt` to format.  It would have fixed that line.

tests/integration_test.go
dimalinux: Was my description above correct?

protocol/utils.go
```diff
@@ -45,24 +45,31 @@ func AssetSymbol(b backend.Backend, asset types.EthAsset) (string, error) {
...
-	}
+	// data := log.Data
+	// res, err := abiSF.Unpack(eventName, data)
+	// if err != nil {
+	// 	return err
+	// }
+
+	// if len(res) < 1 {
+	// 	return errLogMissingParams
+	// }
```
dimalinux, line 62: Can this commented out code be deleted? Or do we need it for some near term additional change?

PR #292, opened 2023-01-30
rounding fixes

dimalinux:
* When checking conditions after an APD (arbitrary precision decimal) math operation, `Rounded` is set even when zeros are chopped from a coefficient. Our checks needed to check for `Inexact` instead.
* `calculateRelayerCommission` now rounds the returned relayer commission to the nearest whole Wei value.
* Updated our pricefeed tests to read environment variable for the ETH mainnet endpoint if set

### DIFF
--- a/cmd/relayer/validate.go
+++ b/cmd/relayer/validate.go
@@ -7,14 +7,14 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/athanorlabs/atomic-swap/coins"
-	contracts "github.com/athanorlabs/atomic-swap/ethereum"
 	rcommon "github.com/athanorlabs/go-relayer/common"
-
 	"github.com/cockroachdb/apd/v3"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
+
+	"github.com/athanorlabs/atomic-swap/coins"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


cmd/relayer/validate.go
```diff
@@ -7,14 +7,14 @@ import (
...
-	contracts "github.com/athanorlabs/atomic-swap/ethereum"
 	rcommon "github.com/athanorlabs/go-relayer/common"
-
 	"github.com/cockroachdb/apd/v3"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
+
+	"github.com/athanorlabs/atomic-swap/coins"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
```
dimalinux, line 17: Our `format-go` target should probably be updated to execute:
```
goimports -local github.com/athanorlabs/atomic-swap -w .
```

pricefeed/pricefeed_test.go
```diff
@@ -16,12 +17,22 @@ func init() {
 	logging.SetLogLevel("pricefeed", "debug")
 }
 
+func getMainnetEndpoint() string {
+	endpoint := os.Getenv("ETH_MAINNET_ENDPOINT")
+	if endpoint == "" {
+		endpoint = mainnetEndpoint
+	}
+	return endpoint
+}
```
dimalinux, line 26: Feel free to suggest a different environment variable name. I wanted to lower the friction to getting a reliable ethereum mainnet endpoint set by github action secrets, since the existing tests are failing from time to time.

noot (approved):
thanks!

PR #294, opened 2023-02-01
relayer discovery extras

dimalinux:
Extras to merge into the existing `noot/relayer-discovery` branch.

`contract-address.json` is now `contract-addresses.json` and it looks like this:
```
contract-addresses.json
{
  "swapFactory": "0xcfeb869f69431e42cdb54a4f4f105c19c080a601",
  "forwarder": "0xe78a0f7e598cc8b0bb87894b0f60dd2a88d6a8ab"
}
```

In integration tests, Alice's swapd is now the bootnode for the relayer.


cmd/swapd/contract.go
```diff
@@ -79,37 +84,38 @@ func deploySwapFactory(
...
 
 // writeContractAddressToFile writes the contract address to the given file
-func writeContractAddressToFile(fp, addr string) error {
-	bz, err := json.MarshalIndent(&struct {
-		ContractAddress string
-	}{
-		ContractAddress: addr,
-	}, "", "\t")
+func writeContractAddressToFile(filePath string, addresses *contractAddresses) error {
+	jsonData, err := json.MarshalIndent(addresses, "", "\t")
```
dimalinux, line 116: It doesn't really matter, but I think I like 2 space indenting better than tabs for JSON after looking at it.

scripts/cleanup-test-processes.sh
```diff
@@ -1,35 +1,37 @@
...
+
 # Take note of monero-wallet-rpc instances being killed here if the instances
-# were started by swapd
+# were started by swapd, as killing swapd should have already shut them down.
 echo "Stopping any monero-wallet-rpc instances"
 "${pkill_cmd[@]}" '/monero-wallet-rpc '
 
+echo "Stopping any monerod regest instances"
+if "${pkill_cmd[@]}" '/monerod .* --regtest '; then
+	sleep 2 # we don't want to exit the script while it is still running
```
dimalinux, line 29: I put this sleep here and removed it from `setup-env.sh`. The other changes were just reordering, so that we kill dependent processes first before the processes they depend on.

scripts/run-integration-tests.sh
```diff
@@ -2,14 +2,13 @@
 
 PROJECT_ROOT="$(dirname "$(dirname "$(realpath "$0")")")"
 cd "${PROJECT_ROOT}" || exit 1
+./scripts/cleanup-test-processes.sh
```
dimalinux, line 5: I was hesitant to do this originally, as there is some minor risk of killing a process that you wanted, but if you agree that the benefit exceeds the risk, I'll leave this addition here and in the unit test script.

noot: seems okay to me. does that script still need to manually kill `monero-wallet-rpc` now that it's controlled by swapd?

noot (approved):

scripts/run-integration-tests.sh
dimalinux: For the integration tests, it doesn't need to, as the integration tests are using randomized JSON-RPC port for the `monero-wallet-rpc` instances and the wallets are created in a temporary directory.  On Linux, I haven't seen `monero-wallet-rpc` instances killed (since killing `swapd` sends them SIGTERM), but if you did a forced shutdown of `swapd` on MacOS, the wallet instances probably won't stop. If you see the script killing `monero-wallet-rpc` instances, let me as much information about how it happened as you can so I can look into it. In answer to your question, at least on Linux, we can probably remove that kill.

PR #295, opened 2023-02-01
prevent startup failure due to outdated database entries

dimalinux:


db/database.go
```diff
@@ -108,20 +114,24 @@ func (db *Database) GetAllOffers() ([]*types.Offer, error) {
...
+		if len(id) > idLength {
 			break
 		}
 
 		encodedOffer := iter.Value()
 		offer, err := types.UnmarshalOffer(encodedOffer)
 		if err != nil {
-			return nil, err
+			log.Warnf("removing invalid offer with ID=0x%X from database: %s", id, err)
+			log.Warnf("invalid offer JSON was: %s", string(encodedOffer))
```
dimalinux, line 128: I started the log messages with lower case to match you. :)

db/database.go
```diff
@@ -186,21 +196,26 @@ func (db *Database) GetAllSwaps() ([]*swap.Info, error) {
...
 		// value is the encoded swap
 		encodedSwap := iter.Value()
 		s, err := swap.UnmarshalInfo(encodedSwap)
 		if err != nil {
-			return nil, err
+			log.Warnf("removing invalid swap info with offerID=0x%X from database: %s", id, err)
+			log.Warnf("invalid offer JSON was: %s", string(encodedSwap))
+			if err = db.swapTable.Del(id[:]); err != nil {
+				return nil, fmt.Errorf("failed to remove invalid offer from database: %w", err)
+			}
```
dimalinux, line 214: I'm not sure what the behavior should be here. Deserialization failed, so we don't know if the invalid swap was ongoing or not. Options are:
* Remove the entry as coded above
* Don't return the entry but leave it in the database (giving warnings on every startup)
* Prevent startup

dimalinux: What is the data consistency "agreement" between the `swapTable` and the `offerTable`?  When we remove an offer from the `offerTable`, does any matching Swap Info entry with the same ID need to removed as well?

db/database.go
noot: haha nice, thanks!

db/database.go
noot: hm, I guess this would be a bad case we would like to prevent with versioning. potentially, all swap keys would need to be prefixed by `swap/<version>` so that deserialization can't fail (unless the db got corrupted or something)  I would remove it, we should try to ensure this case doesn't happen.

once the offer is taken it's deleted from the db and a `swap.Info` struct is created and written to the db. however this does raise a problem (unrelated to this PR), if a swap refunds/aborts, the `swap.Info` stays in the db (with the exit status saved), but the offer is also re-added to the db.

dimalinux: Can you add an issue for the problem mentioned above? I'd create it, but I'm not sure I can describe it adequately.

noot: yes will do!

noot: #298

noot (approved):
looks good, in regards to your comment about not deleting any swaps ever, we should probably add versioning or something to ensure that never happens.

PR #296, opened 2023-02-01
cache monero download

dimalinux:
This PR should significantly decrease 2 common CI failures:
(1) Issues downloading the monero binaries (we now cache them). This also reduces build time and bandwidth costs for the monero community.
(2) We use a reliable mainnet endpoint from a github actions secret for unit tests that require it.

This PR also switches us from `go fmt` to `go imports` for formatting the code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@ lint: lint-go lint-shell lint-solidity
 
 .PHONY: format-go
 format-go:
-	go fmt ./...
+	test -x $(GOPATH)/bin/goimports || go install golang.org/x/tools/cmd/goimports@latest
+	$(GOPATH)/bin/goimports -local github.com/athanorlabs/atomic-swap -w .
 
 .PHONY: format-shell
 format-shell:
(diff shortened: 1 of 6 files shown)


pricefeed/pricefeed_test.go
```diff
@@ -17,16 +18,21 @@ func init() {
...
-func getMainnetEndpoint() string {
+func getMainnetEndpoint(t *testing.T) string {
 	endpoint := os.Getenv("ETH_MAINNET_ENDPOINT")
 	if endpoint == "" {
 		endpoint = mainnetEndpoint
 	}
+	eURL, err := url.Parse(endpoint)
+	require.NoError(t, err)
+	// path and fragments may have API keys, so don't log them
+	t.Logf("mainnet endpoint is %s://%s", eURL.Scheme, eURL.Host)
```
dimalinux, line 29: Interestingly, CI redacted this even though the secret part of the secret was omitted:
```
pricefeed_test.go:29: mainnet endpoint is ***
```
Due to the fact that it was redacted, it looks like we are using the correct value, so the log isn't useless.

Makefile
```diff
@@ -73,9 +74,9 @@ bindings:
 .PHONY: mock
 mock:
 	go generate -run mockgen ./...
+	$(MAKE) format-go
```
dimalinux, line 77: This was easier than figuring out how to tell goimports to ignore the mockgen files.

noot (approved):

Issue #210, opened 2022-10-25
Internal protocol document


dimalinux:
We should create a document of the internal swapd message flows.

PR #297, opened 2023-02-02
upgrade linter

dimalinux:
Upgrades linter from 1.48.0 to 1.50.1.

Since making linter fixes required adding code comments on various exported constants, let's use this PR as a chance to get the status and event types properly documented, as tracking through the code to find out their precise meanings is very time consuming.

@noot feel free to edit this branch. It might even be easiest to do it from the github website. I can clean it up if the information is in place.


.golangci.yml
```diff
@@ -37,29 +43,32 @@ linters-settings:
...
-    - megacheck
     - misspell
     - nolintlint
+    - predeclared
     - revive
     - staticcheck
+    - typecheck
     - unconvert
     - unparam
-    - varcheck
```
dimalinux, line 62: `varcheck` was abandoned and superseded by `unused`.

cmd/swaptester/main.go
```diff
@@ -206,7 +206,7 @@ var (
 )
 
 func getRandomExchangeRate() *coins.ExchangeRate {
-	rateFl := minExchangeRate + mrand.Float64()*(maxExchangeRate-minExchangeRate) //nolint:gosec
+	rateFl := minExchangeRate + mrand.Float64()*(maxExchangeRate-minExchangeRate)
```
dimalinux, line 209: Don't know why these changed. Maybe because it is a float value, which wouldn't normally be using for cryptography.

common/consts.go
```diff
@@ -3,30 +3,36 @@ package common
 
 import "github.com/cockroachdb/apd/v3"
 
+// Daemon default ports and URLs
```
dimalinux, line 6: Apparently, we can group the `const` blocks around related values and then just give the group one comment and the linter is happy. It has to be multiple blocks, because if the comment is inside the `const` block, it is associated only with one value.

common/types/status.go
```diff
@@ -4,18 +4,31 @@ package types
 // Status represents the stage that a swap is at.
 type Status byte
 
+// Status values
```
dimalinux, line 7: Check these carefully as I didn't write the code they document. If we can get good comments on all the status and event types, we can close https://github.com/AthanorLabs/atomic-swap/issues/210.

protocol/swap/manager.go
```diff
@@ -37,10 +35,12 @@ type manager struct {
 	past    map[types.Hash]*Info
 }
 
+var _ Manager = (*manager)(nil)
```
dimalinux, line 38: I just moved below the type declarations it uses.

protocol/swap/manager.go
```diff
@@ -37,10 +35,12 @@ type manager struct {
...
+
 // NewManager returns a new Manager that uses the given database.
 // It loads all ongoing swaps into memory on construction.
 // Completed swaps are not loaded into memory.
-func NewManager(db Database) (*manager, error) {
+func NewManager(db Database) (Manager, error) {
```
dimalinux, line 43: We could move the `Manager` interface out of this package and capitalize the `manager` struct, but it seemed out of scope for upgrading the linter. This was to get rid of the complaint that a public function was returning a private type.

protocol/xmrmaker/event.go
```diff
@@ -11,8 +11,9 @@ import (
 // EventType ...
 type EventType byte
 
+// ...
```
dimalinux, line 14: I'd like to get good descriptions on both the maker and taker event constants before we merge. All events should clearly state if they are triggered by the counterparty, contract logs, or something in our own code.

cmd/swapd/profile.go
```diff
@@ -21,8 +21,14 @@ func maybeStartProfiler(c *cli.Context) error {
 	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 
+	server := &http.Server{
+		Addr:              bindIPAndPort,
+		ReadHeaderTimeout: time.Second,
```
dimalinux, line 26: This is just the profiler, but I decided to add the timeout here instead of disabling the lint complaint.

common/types/status.go
```diff
@@ -4,18 +4,33 @@ package types
...
 	// CompletedSuccess represents a successful swap.
 	CompletedSuccess
 	// CompletedRefund represents a swap that was refunded.
 	CompletedRefund
-	// CompletedAbort represents the case where the swap aborts before any funds are locked.
+	// CompletedAbort represents the case where the swap aborts before any funds
+	// are locked.
 	CompletedAbort
+	// UnknownStatus is a placeholder for unmatched status strings.
+	// TODO: Why is it an indicator that a swap is ongoing?
```
dimalinux, line 33: I was trying to understand this line when I wrote that TODO:
https://github.com/AthanorLabs/atomic-swap/blob/master/common/types/status.go#L100

tests/integration_test.go
```diff
@@ -693,6 +694,7 @@ func (s *IntegrationTestSuite) testAbortXMRMakerCancels(asset types.EthAsset) {
 	default:
 	}
 
+	common.SleepWithContext(ctx, 2*time.Second) // give some extra time for the offer to be re-added
```
dimalinux, line 697: Our `TestAbort_ERC20_XMRMakerCancels` integration test is flaky. Interestingly, I'm only seeing the ERC20 version of this test fail sporadically, not the mostly-shared-code `TestAbort_XMRMakerCancels` test. On line 700 below, the length is zero when it is supposed to be one. Do you agree that there is a delay between the maker exiting his go process above and the offer being re-added? Since the failure is sporadic, I can't be certain this fix is helping.

common/types/status.go
noot: I think these are good!

common/types/status.go
noot: I wanted that function to only return false if the status was explicitly a completed one, but I think that `UnknownStatus` case should never be hit. we can change to a panic/err instead?

dimalinux: Done: https://github.com/AthanorLabs/atomic-swap/commit/92d0ad25cdc0788dcfdabe968861dd1ff909e6e5

protocol/swap/manager.go
noot: I think we should do that, in another PR

tests/integration_test.go
noot: yeah, it's probably a timing issue, does increasing this timeout help?

noot (approved):
looks good!

tests/integration_test.go
dimalinux: I think so.

PR #302, opened 2023-02-06
fix: update derivation of shared swap view key to be the same for all paths

dimalinux:
The shared AB Monero wallet does not use the standard view key derivation. It is created by summing Alice's and Bob's view keys and can not be derived from the AB spend key, as is the case in most wallets. The data we were storing to the recovery database in case swapd is restarted mid-swap only had the spend key and was attempting to do standard view key derivation from it. This PR fixes the issue storing both keys to the recovery database.

This PR is still a draft. Not yet done:
1. Determine some unit tests that would have caught the original issue.
2. Simplify some recovery code, breaking it into less monolithic functions.


.github/workflows/checks.yml
```diff
@@ -30,6 +30,17 @@ jobs:
...
+  mockgen-check:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.x'
+      - uses: actions/checkout@v3
+
+      - name: Generate mocks and verify no changes
+        run: make mock && diff -u <(echo -n) <(git status --short --untracked-files=no)
```
dimalinux, line 42: A bunch of these diffs, like this one, will go away when https://github.com/AthanorLabs/atomic-swap/pull/303 is merged to master.

noot:
my additions:
- update recovery DB to have `Put/GetCounterpartySwapPrivateKey` functions
- update both maker/taker to use the same `GetClaimKeypair` and `ClaimMonero` functions, which ensures their swap key/address derivations will be the same and they're using the same logic for obtaining xmr.
- add unit tests for `ClaimMonero`
- break down recovery code to have a specific function (`completeSwap`) for when the counterparty's private swap key is known

monero/wallet_client.go
```diff
@@ -76,6 +76,46 @@ type WalletClientConf struct {
 	LogPath             string               // optional, default is dir(WalletFilePath)/../monero-wallet-rpc.log
 }
 
+// Fill fills in the optional configuration values (Port, MonerodNodes, MoneroWalletRPCPath,
+// and LogPath) if they are not set.
```
dimalinux, line 80: It is also picking a the first successfully validated monero node.

monero/wallet_client.go
```diff
@@ -76,6 +76,46 @@ type WalletClientConf struct {
...
+func (conf *WalletClientConf) Fill() error {
+	if conf.WalletFilePath == "" {
+		panic("WalletFilePath is a required conf field") // should have been caught before we were invoked
+	}
```
dimalinux, line 84: This check is still in `NewWalletClient` before it calls `Fill()`. In `NewWalletClient`, you probably want the check to see if the wallet file exists to be after the call to `Fill()`.

protocol/claim_monero.go
```diff
@@ -0,0 +1,115 @@
...
+		log.Infof("transferred %s XMR to primary wallet (%s XMR lost to fees)",
+			coins.FmtPiconeroAmtAsXMR(transfer.Amount),
+			coins.FmtPiconeroAmtAsXMR(transfer.Fee),
+		)
+	}
+
+	return abAddr, nil
+}
+
+// TODO: Put this in monero package? Unit test.
```
dimalinux, line 94: If it is only used by the `ClaimMonero` function above, here is probably the best place (easiest to find for someone reviewing the code). Its reuse potential is only with AB swap wallets, as the primary wallet's funds may be in a constant state of flux, were some portion of its funds are always locked.

protocol/claim_monero_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	err = conf.Fill()
+	require.NoError(t, err)
+
+	moneroCli, err := monero.CreateSpendWalletFromKeys(conf, kp, 0)
+	require.NoError(t, err)
+	height, err := moneroCli.GetHeight()
+	require.NoError(t, err)
+	amt := coins.StrToDecimal("1")
+	pnamt := coins.PiconeroAmount(*amt)
```
dimalinux, line 41: xmrAmt := coins.StrToDecimal("1")
pnAmt := coins.MoneroToPiconero(xmrAmt)

protocol/claim_monero_test.go
```diff
@@ -0,0 +1,95 @@
...
```
dimalinux, line 77: xmrAmt := coins.StrToDecimal("1")
pnAmt := coins.MoneroToPiconero(xmrAmt)

protocol/xmrmaker/instance.go
```diff
@@ -198,6 +188,48 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
 	return nil
 }
 
+func (inst *Instance) completeSwap(s *swap.Info, skA *mcrypto.PrivateSpendKey) error {
+	// calc counterparty's swap private view key
+	// TODO: if the counterparty sends a non-standard public view key, does this work?
```
dimalinux, line 193: It won't work. The names make it sound like you only need the spend key to spend, but in order to move funds around, you need both. I'm not sure on the technical reasons, but I think the wallet doesn't even know which transaction outputs belong to it without having the view key set correctly.

protocol/xmrmaker/swap_state.go
```diff
@@ -410,29 +410,37 @@ func (s *swapState) exit() error {
 	}
 }
 
-func (s *swapState) reclaimMonero(skA *mcrypto.PrivateSpendKey) (mcrypto.Address, error) {
+func (s *swapState) reclaimMonero(skA *mcrypto.PrivateSpendKey) error {
+	// write counterparty swap privkey to file in case something goes wrong
```
dimalinux, line 414: I would say "to disk" or "to the database", otherwise it sounds like an independent file.

protocol/xmrtaker/claim.go
```diff
@@ -83,82 +81,40 @@ func (s *swapState) claimMonero(skB *mcrypto.PrivateSpendKey) (mcrypto.Address,
...
-
-	// write keys to file in case something goes wrong
-	err := s.Backend.RecoveryDB().PutSharedSwapPrivateKey(s.ID(), kpAB.SpendKey())
-	if err != nil {
-		return "", err
-	}
-
-	conf := s.XMRClient().CreateWalletConf("xmrtaker-swap-wallet-claim")
-	abWalletCli, err := monero.CreateSpendWalletFromKeys(conf, kpAB, s.walletScanHeight)
+	// write counterparty swap privkey to file in case something goes wrong
```
dimalinux, line 84: Same as the maker comment above, "to disk" or "to the database".

protocol/xmrtaker/instance.go
```diff
@@ -173,6 +137,47 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
 	return nil
 }
 
+func (inst *Instance) completeSwap(s *swap.Info, skB *mcrypto.PrivateSpendKey) error {
```
dimalinux, line 140: It's not obvious from the name how this function is used, so a comment might be helpful. As I understand it, this function only completes recovered swaps from a previous run of `swapd`. It looks like it will use the current value of "transfer back" (verses whatever value was set when the swap was started). It will also only only recover to the primary wallet address, not whatever address was used when the swap was started.  Both of those features are painful to support in the corner cases. Assuming we continue support, we should shine as much light into their darker corners as possible.

protocol/xmrmaker/instance.go
```diff
@@ -127,21 +129,9 @@ func (inst *Instance) abortOngoingSwap(s *swap.Info) error {
...
-		primaryCli := inst.backend.XMRClient()
-		destAddr := primaryCli.PrimaryAddress()
-		conf := primaryCli.CreateWalletConf("xmrmaker-swap-wallet-refund-ongoing")
-		err = sweepRefund(inst.backend.Ctx(), inst.backend.Env(), s.ID, conf, s.MoneroStartHeight, kp, destAddr)
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return inst.completeSwap(s, skA)
```
dimalinux, line 134: Does the swap info database entry need to get updated to completed after this?

protocol/claim_monero.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		log.Errorf(
+			"failed to transfer XMR out of swap wallet, dest address %s is invalid: %s",
+			abAddr,
+			err,
+		)
+		return "", err
+	}
+
+	err = waitUntilBalanceUnlocks(ctx, abWalletCli)
```
dimalinux, line 74: This call isn't actually needed.  Waiting for the blocks to unlock is handled by the `SweepAll` method.

protocol/xmrmaker/instance.go
noot: yep I'm going to fix this in #304

protocol/claim_monero.go
noot: `SweepAll` wait only 2 blocks but this waits for the balance to fully unlock (10 blocks). I think waiting just the 2 blocks is fine, should I remove this func?

protocol/xmrmaker/instance.go
noot: good point, I'll fix this. and yep I'll add the `recoveryDB.DeleteSwap()` call

protocol/claim_monero.go
dimalinux: Sweep all waits for as many blocks as needed at the start if the SweepAll, but if new funds were received while waiting for those blocks, it won't continue waiting for the additional blocks.  I think that is good enough. I was trying to put all logic that would be used every time that `SweepAll` is called directly into our custom `SweepAll` method.

monero/wallet_client.go
```diff
@@ -76,6 +76,47 @@ type WalletClientConf struct {
 	LogPath             string               // optional, default is dir(WalletFilePath)/../monero-wallet-rpc.log
 }
 
+// Fill fills in the optional configuration values (Port, MonerodNodes, MoneroWalletRPCPath,
+// and LogPath) if they are not set.
+// If empty, MonerodNodes is set to the first validated node.
```
dimalinux, line 81: I'd remove the words "If empty". The code is picking the first validated node independent as to whether the nodes were defaulted or not.  I thought it was worth adding to the comments, as it is doing more than just filling.

monero/wallet_client.go
```diff
@@ -95,47 +136,22 @@ type walletClient struct {
 
 // NewWalletClient returns a WalletClient for a newly created monero-wallet-rpc process.
 func NewWalletClient(conf *WalletClientConf) (WalletClient, error) {
-	if conf.WalletFilePath == "" {
-		panic("WalletFilePath is a required conf field") // should have been caught before we were invoked
-	}
-
 	if path.Dir(conf.WalletFilePath) == "." {
 		return nil, errors.New("wallet file cannot be in the current working directory")
 	}
```
dimalinux, line 141: Does this check belong in `Fill()` or not? It's kind of an arbitrary decision on what Fill is supposed to accomplish.  Whatever you decide is fine, as long as it was considered.

protocol/xmrmaker/instance.go
```diff
@@ -198,6 +188,63 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
...
+		s.MoneroStartHeight,
+		kpAB,
+		inst.backend.XMRClient().PrimaryAddress(),
+		true, // always sweep back to our primary address
+	)
+	if err != nil {
+		return err
+	}
+
+	s.Status = types.CompletedSuccess
```
dimalinux, line 239: We are in the XMR Maker's code.  Should that be `CompletedRefund` above?

monero/wallet_client.go
noot: `Fill` doesn't do anything with `WalletFilePath`, so I left it here

protocol/xmrmaker/instance.go
noot: yes, updated

dimalinux (commented):
Looks good. Approved!  It won't let me select Approved because I created the PR.

noot (approved):

PR #303, opened 2023-02-08
fixes mock generation and attempts to fix TestDaemon_PersistOffers

dimalinux:
* Updates to get mock generation working
* Ensures mock generation is working in future PRs
* Attempts to fix `TestDaemon_PersistOffers` failures deleting its temporary data directory
* Updates some dependencies to their latest versions


go.mod
```diff
@@ -6,16 +6,16 @@ require (
 	filippo.io/edwards25519 v1.0.0
 	github.com/ChainSafe/chaindb v0.1.5-0.20221010190531-f900218c88f8
 	github.com/MarinX/monerorpc v1.0.5
-	github.com/Masterminds/semver/v3 v3.1.1
-	github.com/athanorlabs/go-dleq v0.0.0-20221228030413-8ac300febf46
+	github.com/Masterminds/semver/v3 v3.2.0
+	github.com/athanorlabs/go-dleq v0.0.0-20230113214619-d6fd7c03e213
```
dimalinux, line 10: We should start getting release tags onto all the anthorlabs repos.

cmd/swapd/main_test.go
```diff
@@ -137,6 +137,12 @@ func TestDaemon_PersistOffers(t *testing.T) {
 	startupTimeout := time.Millisecond * 100
 
 	dataDir := t.TempDir()
+	defer func() {
+		// CI has issues with the filesystem still being written to when it is
+		// recursively deleting dataDir. Can't be replicated outside of CI.
+		unix.Sync()
+		time.Sleep(500 * time.Millisecond)
+	}()
```
dimalinux, line 145: Line 192 below calls `d.stop()`, so I don't know what could be writing to the filesystem when the `dataDir` is recursively deleted:
```
testing.go:1097: TempDir RemoveAll cleanup: unlinkat /tmp/TestDaemon_PersistOffers1622151463/001: directory not empty
```
Given that I can't replicate it outside of CI, my theory is that the CI filesystem may just be crappy and still flushing stuff to disk. I tried running it a very large number of times outside of CI, so I could capture what was in the directory, but couldn't get a failure.

cmd/swapd/main.go
```diff
@@ -409,6 +409,7 @@ func (d *daemon) make(c *cli.Context) error { //nolint:gocyclo
 	if err != nil {
 		return err
 	}
+	defer ec.Close()
```
noot, line 412: remove this? it's passed into `newBackend` which is passed into `extethclient.NewEthClient`

dimalinux: The backend goes out of scope at the end of this function too. It is not part of the `daemon` type's fields.  The extended ethclient doesn't close the connection, because it didn't create the connection. The XMR client closer is also handled as a defer in this, somewhat oversized, `make` method.  I'm open to other ideas, but I think this is the logical place to close it at the current time.

noot: the backend goes out of scope, but it's passed into `getProtocolInstances` which uses the backend for both maker/taker. am I missing something? it seems like the ethclient is still used at that point

dimalinux: I agree, it is usually best to destroy from the outside to the inside (reverse order of creation). This function is already destroying the XMR client, which is analogous to the eth client, so it is not doing that already. I have some ideas on how to rewrite this code that I can put into a future PR. I removed the `ec.Close()` for now.

noot (approved):

PR #304, opened 2023-02-09
fix: update xmrmaker refund path wallet to use address from xmrtaker swap keys

noot:
- update xmrtaker to send private view key in key exchange; this allows the xmrmaker to generate the lock address using the counterparty private view key. this fixes the fact that there is no verification that the counterparty public view key (sent previously) actually will match the private view key that will eventually be derived from the private spend key.
- all monero address derivations on the xmrmaker side are now the same


protocol/xmrmaker/instance.go
```diff
@@ -198,6 +188,53 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
...
+
+	// we save the counterparty's public keys in case they send public keys derived
+	// in a non-standard way.
+	xmrtakerPublicKeys, err := inst.backend.RecoveryDB().GetXMRTakerSwapKeys(s.ID)
+	if err != nil {
+		return err
+	}
+
+	address := mcrypto.SumSpendAndViewKeys(
+		xmrtakerPublicKeys, kpAB.PublicKeyPair()).Address(inst.backend.Env())
```
dimalinux, line 223: I assume this address is the same as `kpAB.Address(inst.backend.Env())`.  If so, you can skip passing it down 2 more levels of code and just call `CreateSpendWalletFromKeys` instead of the new function `CreateSpendWalletFromKeysAndAddress`.

noot: it's not necessarily the same address, which is why I introduced the new function. the xmrtaker can send us public view keys that aren't derived the usual way, and there's no way to check if it is or not (unless we add some complex zk proof or something), in which case the address we lock the xmr in will not be the same as when we get the address from the summed private keys.

noot: the other solution is to get the xmrtaker to also send the xmrmaker its private view key, but I'd rather not make unnecessary protocol changes. it would also fix this issue though

dimalinux: Isn't this code path only creating spend wallets? For spend wallets, the address is always computed directly from the private keys.  For ViewOnly wallets, I created `CreateViewOnlyWalletFromKeys`, which does take the address as a parameter.

noot: yes it's creating the spend wallet, but the address still has to match that in which we locked the XMR in

dimalinux: Do we check that the address matches before doing the expensive operation of creating a new `monero-wallet-rpc` instance? On a relative scale, computing the address very cheap. As long as it is not in a loop, invoking the address function multiple times is not an issue.

`CreateSpendWalletFromKeys` and the new, `CreateSpendWalletFromKeysAndAddress`, are the exact same function. Let's only keep one of them.  I liked the original, because I thought it made the code more self documenting. It lets the reader better understand the difference between spend and view-only wallets at creation time.  That with spend wallets, the address is a derived value, while with view-only wallets, it is a mandatory, non-derived value critical to create the wallet. When creating a spend wallet, it is very well vetted at this point that the address will always match that which we calculate from the private keys. Our unit tests for this have never failed.

noot: @dimalinux the PR has been redone, pls re-review

monero/wallet_client_test.go
```diff
@@ -228,6 +228,44 @@ func TestCallGenerateFromKeys(t *testing.T) {
 	t.Logf("Address %s", addr.Address)
 }
 
+func TestCallGenerateFromKeys_UnusualAddress(t *testing.T) {
+	// this tests calling generateFromkeys passing an address derived in
+	// a non-standard manner (ie. the view keys aren't a hash of the spend
+	// key). the view key is actually a completely different key altogether.
```
dimalinux, line 234: I'd move this comment above the function declaration, so it is a description of the test itself, and doesn't look like a description of the first part of the test. Also describe what the goal of the test is. Are we testing that `monero-wallet-rpc` generates the same address that we do?  That the wallet works (might need a transfer for that test to be complete).  Are we testing that `monero-wallet-rpc` accepts view keys that are not derived from the spend key when using [generate_from_keys](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#generate_from_keys)?

dimalinux (approved):
Nice, elegant solution! I only had one request for clarification on a unit test.

monero/wallet_client_test.go
noot: yeah, we're testing that `monero-wallet-rpc` accepts an address that doesn't match the private view key provided. will update!

PR #305, opened 2023-02-14
JSON updates

dimalinux:
* Ensures that all p2p message types have JSON tags conforming to the same standard (camel case, first letter lower case)
* When available, marshaling uses our project types instead of string
* The `common` package was decoupled from the `message` package to allow non-string message fields
* Protocol version bumped

Follow up PR(s) will provide additional input validation on the p2p message types.


protocol/xmrtaker/swap_state_test.go
```diff
@@ -255,8 +259,12 @@ func TestSwapState_NotifyXMRLock(t *testing.T) {
...
 	require.NoError(t, err)
 
-	s.setXMRMakerKeys(xmrmakerKeysAndProof.PublicKeyPair.SpendKey(), xmrmakerKeysAndProof.PrivateKeyPair.ViewKey(),
-		xmrmakerKeysAndProof.Secp256k1PublicKey)
+	err = s.setXMRMakerKeys(
+		xmrmakerKeysAndProof.PublicKeyPair.SpendKey(),
+		xmrmakerKeysAndProof.PrivateKeyPair.ViewKey(),
+		xmrmakerKeysAndProof.Secp256k1PublicKey,
+	)
+	require.NoError(t, err)
```
dimalinux, line 267: These changes just add a check on the returned error value. My IDE was complaining.

crypto/monero/crypto.go
```diff
@@ -227,9 +194,9 @@ func NewPublicKeyFromBytes(b []byte) (*PublicKey, error) {
 	}, nil
 }
 
-// Hex returns the hex-encoded canonical byte representation of the PublicKey.
-func (k *PublicKey) Hex() string {
-	return hex.EncodeToString(k.key.Bytes())
+// String proves the key as hex
```
noot, line 197:
```suggestion
// String formats the key as a 0x-prefixed hex string
```

monero/wallet_client.go
```diff
@@ -492,14 +493,14 @@ func (c *walletClient) generateFromKeys(
 
 	spendKey := "" // not used when only generating a view key
 	if sk != nil {
-		spendKey = sk.Hex()
+		spendKey = hex.EncodeToString(sk.Bytes())
```
noot, line 496: I kinda preferred the `Hex()` method, what do you think about having `Hex()` return no 0x-prefix and `String` returns with the prefix?

protocol/txsender/external_sender.go
```diff
@@ -102,9 +102,15 @@ func (s *ExternalSender) Approve(spender ethcommon.Address,
 }
 
 // NewSwap prompts the external sender to sign a new_swap transaction
-func (s *ExternalSender) NewSwap(_pubKeyClaim [32]byte, _pubKeyRefund [32]byte,
-	_claimer ethcommon.Address, _timeoutDuration *big.Int, _nonce *big.Int, _ethAsset types.EthAsset,
-	value *big.Int) (ethcommon.Hash, *ethtypes.Receipt, error) {
+func (s *ExternalSender) NewSwap(
+	_pubKeyClaim [32]byte,
```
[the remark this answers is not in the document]
noot: yeah, you can remove the underscores

rpc/net.go
```diff
@@ -168,7 +168,7 @@ func (s *NetService) takeOffer(who peer.ID, offerID types.Hash, providesAmount *
 		return nil, fmt.Errorf("failed to initiate protocol: %w", err)
 	}
 
-	skm := swapState.SendKeysMessage()
+	skm := swapState.SendKeysMessage().(*message.SendKeysMessage)
```
noot, line 171:
```suggestion
	skm, ok := swapState.SendKeysMessage().(*message.SendKeysMessage)
	if !ok {
	    panic("expected SendKeysMessage")
	}
```
this shouldn't happen, but good to check again

noot (approved):
looks good! only minor comments

rpc/net.go
dimalinux: We'll get a panic with a line number in both cases. Is a slightly prettier panic worth additional code complexity?

PR #308, opened 2023-02-16
add timestamps to status updates in swapcli make/take

### LINKED ISSUE #307: swapcli make/take --subscribe should have timestamps
as title says, currently there are no timestamps so it's hard to gauge how far along we are. 

would also be good to add additional info other than status; eg estimated time for swap to complete based on block time, how many more blocks we're waiting for (although this second one might need its own issue)

noot:
closes #307 

will probably open another issue for additional info that would be useful to provide to the user that I briefly mentioned in the above.


cmd/swapcli/main.go
```diff
@@ -545,7 +546,7 @@ func runMake(ctx *cli.Context) error {
 		printOfferSummary(resp)
 
 		for stage := range statusCh {
-			fmt.Printf("> Stage updated: %s\n", stage)
+			fmt.Printf("%s > Stage updated: %s\n", time.Now().Format("2006-01-02T15:04:05"), stage)
```
dimalinux, line 549: We have a constant for that formatting option, `common.TimeFmtSecs`. Your call on whether to use it. My thought was that if we ever wanted to update how time is formatted in one place, the constant might help. It's not a guarantee that we always want to use the same format everywhere though.

dimalinux (approved):

cmd/swapcli/main.go
dimalinux: I had some extra time, so I just replaced the format strings without our constant. I didn't realize it initially, but the format you had was slightly different than the constant:

Our constant:
```
2023-02-18-16:41:17
```
Your initial version:
```
2023-02-18T16:41:17
```
If you like the `T` better than the `-`, let's update the constant so we use the same format everywhere.

PR #309, opened 2023-02-16
JSON validation

dimalinux:
Integrates the go-playground validator library to provide validations based on struct field annotations. Documentation here:
https://pkg.go.dev/github.com/go-playground/validator/v10

A wrapper library named `vjson` (validated JSON) was added to call the validator when we do marshal/unmarshal operations.


net/message/message.go
```diff
@@ -65,15 +65,15 @@ func DecodeMessage(b []byte) (common.Message, error) {
...
+	if err := vjson.UnmarshalStruct(msgJSON, msg); err != nil {
 		return nil, fmt.Errorf("failed to decode %s message: %w", TypeToString(msg.Type()), err)
 	}
 	return msg, nil
 }
 
 // QueryResponse ...
 type QueryResponse struct {
-	Offers []*types.Offer `json:"offers"`
+	Offers []*types.Offer `json:"offers" validate:"dive,required"`
```
dimalinux, line 76: The validate directives are processed in order. If you use, for example, `omitempty`, in most cases you need it at the front of the list. Above, the `dive` directive says that subsequent directives apply to elements of the slice, not the slice itself. Our goal here is to validate that no slice entry is `nil`.

net/host.go
```diff
@@ -24,7 +24,7 @@ const (
 	maxMessageSize = 1 << 17
 )
 
-var log = logging.Logger("host")
+var log = logging.Logger("net")
```
dimalinux, line 27: I lost a lot of time on this, because I wasn't seeing necessary logs in the integration tests. We should invest some time into switching to a different logger that is more automated or figuring out a less manual way of using the existing logger.

I also don't understand what version we should be using for our existing logger.  IPFS is putting out new releases under `v2` (`github.com/ipfs/go-log/v2`), but we are still using the original version and the README in their repo refers to the original version even in the v2 releases:
https://github.com/ipfs/go-log/tree/v2.5.1/

net/host_test.go
```diff
@@ -27,11 +27,11 @@ func (h *mockHandler) GetOffers() []*types.Offer {
 	return []*types.Offer{}
 }
 
-func (h *mockHandler) HandleInitiateMessage(_ *message.SendKeysMessage) (s SwapState, resp Message, err error) {
+func (h *mockHandler) HandleInitiateMessage(msg *message.SendKeysMessage) (s SwapState, resp Message, err error) {
 	if (h.id != types.Hash{}) {
-		return &mockSwapState{h.id}, &SendKeysMessage{}, nil
+		return &mockSwapState{h.id}, msg, nil
```
dimalinux, line 32: This is a little hacky to reuse the same `SenkKeysMessage` like this.  Or at least, it might be confusing for someone learning the code by looking at the tests. Suggestions on how to do it better welcome.

net/message/message.go
```diff
@@ -104,12 +104,12 @@ func (m *QueryResponse) Type() byte {
 // SendKeysMessage is sent by both parties to each other to initiate the protocol
 type SendKeysMessage struct {
 	OfferID            types.Hash              `json:"offerID"`
```
dimalinux, line 106: The `OfferID` above should definitely be tagged as required. The `required` validation will compare a variable to its default initialization value.  In this case, it would compare `OfferID` with an all-zero hash and fail it if it is all zeros. I'll lock down the other fields more in a follow-up PR, because these changes will require non-trivial updates to tests.

net/message/message.go
```diff
@@ -104,12 +104,12 @@ func (m *QueryResponse) Type() byte {
...
-	ProvidedAmount     *apd.Decimal            `json:"providedAmount"`
-	PublicSpendKey     *mcrypto.PublicKey      `json:"publicSpendKey"`
+	ProvidedAmount     *apd.Decimal            `json:"providedAmount" validate:"required"`
+	PublicSpendKey     *mcrypto.PublicKey      `json:"publicSpendKey" validate:"required"`
 	PublicViewKey      *mcrypto.PublicKey      `json:"publicViewKey"`
 	PrivateViewKey     *mcrypto.PrivateViewKey `json:"privateViewKey"`
```
dimalinux, line 110: @noot After your current work is complete, which fields will be optional by the maker and/or taker, but required by the other party in this structure?  If they don't line up, should we consider using 2 different types for better self-documenting code?  We already changed the top level functions to use an interface, so the impact might not be that bad.

rpc/ws.go
```diff
@@ -62,8 +62,8 @@ func (s *wsServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			break
 		}
 
-		var req *rpctypes.Request
-		err = json.Unmarshal(message, &req)
+		req := new(rpctypes.Request)
+		err = vjson.UnmarshalStruct(message, req)
```
dimalinux, line 66: The validation function expects a pointer to a struct as its top-level parameter, not a pointer to a pointer to a struct. We had discussed changing this previously, because most code examples you see don't push allocation into the Unmarshaller like this. They are all updated now.

net/initiate_test.go
```diff
@@ -1,14 +1,35 @@
...
+	require.NoError(t, err)
+	return &message.SendKeysMessage{
+		OfferID:            types.Hash{},
+		ProvidedAmount:     new(apd.Decimal),
+		PublicSpendKey:     keysAndProof.PublicKeyPair.SpendKey(),
+		PublicViewKey:      keysAndProof.PublicKeyPair.ViewKey(),
+		PrivateViewKey:     keysAndProof.PrivateKeyPair.ViewKey(),
+		DLEqProof:          hex.EncodeToString(keysAndProof.DLEqProof.Proof()),
+		Secp256k1PublicKey: keysAndProof.Secp256k1PublicKey,
+		EthAddress:         ethcommon.Address{},
```
dimalinux, line 29: Tests depend on the `OfferID` and the `ETHAddress` being all zeros.  I'll address that in a follow-up PR.

net/host.go
noot: yeah I haven't looked into loggers recently, just been using this for a while. might be worth exploring what's out there

net/host_test.go
noot: why does it need to do this? I guess it just needs to return a valid message? if so maybe making a valid test message might be more clear

noot: eg. return `createSendKeysMessage(t)`?

net/message/message.go
noot: `PublicViewKey` will be gone, only `PrivateViewKey` will be here, so the messages will actually be the same for both parties which is nicer

noot (approved):
looks good, this is a nice add!

net/host_test.go
dimalinux: That was what I wanted to do initially, but I didn't have access to `t`.  I added `t` to `mockHandler` so I could call it here.

PR #310, opened 2023-02-17
global solidity imports

dimalinux:
* Switches from v0.8.16 to v0.8.17
* Switches from global solidity symbol imports to selective symbol imports


ethereum/contracts/TestUtils.sol
```diff
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.5 .0;
 
-import "hardhat/console.sol";
+//import "hardhat/console.sol";
```
dimalinux, line 5: I don't see anything using this file. Can I just remove it?

noot: yeah sure

noot (approved):

Issue #288, opened 2023-01-25
swapd should immediately exit with good error message when a subcommand is passed


dimalinux:
It is easy to type `./bin/swapd` when you wanted to type `./bin/swapcli`.  You get something like this:
```
./bin/swapd get-offers
2023-01-25T03:41:18.100Z	ERROR	cmd	swapd/main.go:291	RPC/Websocket server exited: dial tcp 127.0.0.1:8545: connect: connection refused
```
If we have a non flag parameter on the `swapd` command line, we should just exit immediately with a clear error. We want the same behavior that the user already gets when they pass an unknown flag name.

PR #311, opened 2023-02-17
swapd error on subcommand

dimalinux:


ethereum/extethclient/eth_wallet_client.go
```diff
@@ -162,7 +162,7 @@ func (c *ethClient) SetGasPrice(gasPrice uint64) {
 		c.gasPrice = nil
 		return
 	}
-	c.gasPrice = big.NewInt(0).SetUint64(gasPrice)
+	c.gasPrice = new(big.Int).SetUint64(gasPrice)
```
dimalinux, line 165: I just did a global replace on `big.NewInt(0).` to `new(big.Int).`. If it didn't have an extra dot at the end, it didn't get replaced, as those are instances where the user needed a zero value `big.Int` verses an arbitrary `big.Int` object. (Although the standard library uses `new(big.Int)` for both.)

cmd/swapd/main.go
```diff
@@ -215,7 +215,6 @@ var (
 func main() {
 	if err := app.Run(os.Args); err != nil {
 		log.Fatal(err)
-		os.Exit(1)
```
dimalinux, line 218: I don't think there are any loggers for Go that don't exit in the Fatal* methods, but I made sure the IPFS logger wasn't an exception before committing this.

cmd/swapd/main.go
```diff
@@ -276,9 +275,10 @@ func setLogLevels(level string) {
 }
 
 func runDaemon(c *cli.Context) error {
-	ctx, cancel := context.WithCancel(c.Context)
-	defer cancel()
-	go signalHandler(ctx, cancel)
+	// Fail if any non-flag arguments were passed
+	if c.Args().Present() {
+		return fmt.Errorf("unknown command %q", c.Args().First())
+	}
```
dimalinux, line 281: I guess this is only an issue if the whole app runs off of the top level executable with no subcommands. `swapcli` automatically handles unknown subcommands.

cmd/swapd/main.go
noot: do they exit with a non zero exit code?

noot (approved):

noot:
does this close #288 ? if so can you update the PR description

cmd/swapd/main.go
dimalinux: Yes, the standard library `log.Fatal`, this IPFS `log.Fatal` as well as logrus' `Fatal` all exit with the status of `1`.

PR #312, opened 2023-02-18
harmonize offer deletion behavior

dimalinux:
Removes `GetOfferFromDB` and `DeleteOfferFromDB` methods. Ensures that critical errors are not ignored and that mutex locks are not bypassed.


protocol/xmrmaker/instance.go
```diff
@@ -134,7 +134,7 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
 		return inst.completeSwap(s, skA)
 	}
 
-	offer, err := inst.offerManager.GetOfferFromDB(s.ID)
```
dimalinux, line 137: This function was pulling from the cache first, contrary to its name. It didn't seem like we needed it, when we already have `GetOffer`. `GetOffer` returns an error if the offer isn't found, which should be the behavior that we want here.

protocol/xmrmaker/net.go
```diff
@@ -51,8 +51,10 @@ func (inst *Instance) initiate(
 		}
 	}
 
-	// checks passed, delete offer for now
-	inst.offerManager.DeleteOffer(offer.ID)
+	// checks passed, delete the offer for now
+	if err = inst.offerManager.DeleteOffer(offer.ID); err != nil {
+		return nil, err
+	}
```
dimalinux, line 57: `DeleteOffer` will only return serious errors. Deleting an offer that doesn't exist is error free. I think that behavior is much improved over not returning any errors, but I can see an argument even the new behavior is not homogeneous with the `GetOffer` behavior that returns an error if the offer does not exist. I'm okay with it, but I'm open to alternative suggestions.

protocol/xmrmaker/offers/offer_manager.go
```diff
@@ -127,24 +128,6 @@ func (m *Manager) TakeOffer(id types.Hash) (*types.Offer, *types.OfferExtra, err
 	return offer.offer, offer.extra, nil
 }
 
-// DeleteOfferFromDB deletes the offer from the database.
-func (m *Manager) DeleteOfferFromDB(id types.Hash) error {
-	return m.db.DeleteOffer(id)
-}
```
dimalinux, line 133: It seemed unsafe that this was an exported function bypassing our read-write lock. Callers now use `DeleteOffer` which now returns any serious errors.

protocol/xmrmaker/offers/offer_manager.go
```diff
@@ -127,24 +128,6 @@ func (m *Manager) TakeOffer(id types.Hash) (*types.Offer, *types.OfferExtra, err
...
-
-// GetOfferFromDB returns an offer from memory or the database, if it exists.
-func (m *Manager) GetOfferFromDB(id types.Hash) (*types.Offer, error) {
```
dimalinux, line 136: Callers now use `GetOffers` instead.

tests/integration_test.go
```diff
@@ -664,7 +664,7 @@ func (s *IntegrationTestSuite) testAbortXMRMakerCancels(asset types.EthAsset) {
 
 	peerIDs, err := c.Discover(string(coins.ProvidesXMR), defaultDiscoverTimeout)
 	require.NoError(s.T(), err)
-	require.Equal(s.T(), 1, len(peerIDs))
+	require.Equalf(s.T(), 1, len(peerIDs), "peer count mismatch")
```
dimalinux, line 667: It was easy to look at the test output and incorrectly assume that the offer count at the end did not match up, even though the test was actually failing here. I added messages to clarify.

As to why `TestAbort_ERC20_XMRMakerCancels` would fail, but not `TestAbort_XMRMakerCancels`, I did some research on testify's `Suite` and it runs the tests in alphabetical order. It is not because `Suite` intentionally alphabetizes the tests, but rather because the standard library reflect methods used to discover the functions return them in alphabetical order.

`TestAbort_ERC20_XMRMakerCancels` is always the very first test to run (of *all* the tests) and while the `swapd` daemons were fully started, peer discovery had not completed. I added more time in `run-integration-tests.sh` after starting Charlie and Bob's daemons.

protocol/xmrmaker/backend_offers.go
```diff
@@ -40,16 +40,9 @@ func (b *Instance) GetOffers() []*types.Offer {
...
 
 // ClearOffers clears all offers.
-// If the offer list is empty, it clears all offers.
 func (b *Instance) ClearOffers(offerIDs []types.Hash) error {
-	l := len(offerIDs)
-	if l == 0 {
-		err := b.offerManager.ClearAllOffers()
-		if err != nil {
-			return err
-		}
```
dimalinux, line 50: I don't think we intended to run `ClearOfferIDs` after already running `ClearAllOffers`.

protocol/xmrmaker/instance_test.go
```diff
@@ -157,6 +157,10 @@ func TestInstance_createOngoingSwap(t *testing.T) {
 	rate := coins.ToExchangeRate(apd.New(1, 0)) // 100% relayer commission
 	offer := types.NewOffer(coins.ProvidesXMR, one, one, rate, types.EthAssetETH)
 
+	offerDB.EXPECT().PutOffer(offer).Return(nil)
+	_, err := inst.offerManager.AddOffer(offer, "", nil)
+	require.NoError(t, err)
```
dimalinux, line 162: I'm still confused as to how this test previously worked. I don't see anyone adding the offer to the database or the cache that the offer manager is using.

noot (approved):
looks good! definitely cleaner now.

PR #313, opened 2023-02-18
added address QR code feature to swapcli

### LINKED ISSUE #301: swapcli subcommands for ETH and XMR account QR codes
Even on relatively dumb terminals, we can display QR codes for the monero and ethereum addresses of a `swapd` instance using unicode strings.

Feel free to suggest here what the default behavior and subcommands should be. Right now users have the `balances` command to wallet addresses in text only.  We could have these two commands return the respective wallet address both and text and QR-code format:
```
swapcli eth-address
swapcli xmr-address
```
Sample QR code generation as a multi-line UTF-8 string that displays well even in remote, mostly dumb, terminals:
```
package main

import (
	"fmt"
	"log"

	"github.com/skip2/go-qrcode"
)

func main() {
	const moneroAddress = "4BKjy1uVRTPiz4pHyaXXawb82XpzLiowSDd8rEQJGqvN6AD6kWosLQ6VJXW9sghopxXgQSh1RTd54JdvvCRsXiF41xvfeW5"
	code, err := qrcode.New(moneroAddress, qrcode.Medium)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(code.ToString(false))
}
```
Sample output:

dimalinux:
Adds sub-commands to `swapcli` to get the Ethereum and Monero addresses as QR codes. Closes #301.


cmd/swapcli/main.go
```diff
@@ -397,6 +414,36 @@ func runBalances(ctx *cli.Context) error {
 	return nil
 }
 
+func runETHAddress(ctx *cli.Context) error {
+	c := newRRPClient(ctx)
+	balances, err := c.Balances()
```
dimalinux, line 419: I don't think it is is great that these sub commands are generating traffic to query balances, but the alternative of creating a new API that just returns addresses has the downside of more code to maintain.

noot: I think adding an endpoint that returns just the addresses would be pretty small and cleaner potentially, but this is fine too for now

noot (approved):

PR #316, opened 2023-02-20
update relayer fee to be fixed at 0.009 ETH

### LINKED ISSUE #293: relayer fee should be higher than tx gas fee
The fee received by the relayer should always be higher than the gas cost to submit the tx. Gas cost is the product of gas (which is fixed at around 42k for the call to `claim()`) and gas price, which is set by the relayer but it also based on current network conditions.

Two solutions:
- before submitting a tx, check the current gas price we'll use and calculate the gas fee based on that, and ensure the fee received will be higher
- change the relayer commission to be a fixed fee instead of a percentage of the swap value. this fee should always be higher than the gas fee will be even if the gas price pumps.

Will probably go with the second, since there's the added benefit of users not doing a small swap first to get some ETH in a fresh wallet, then doing a larger swap from there.

noot:
- update relayer fee to be a fixed value instead of swap percentage
- update default fee to be 0.009 ETH 
- change all usages of `relayerCommission` -> `relayerFee`

closes #293

### DIFF
--- a/cmd/relayer/main.go
+++ b/cmd/relayer/main.go
@@ -29,6 +29,7 @@ import (
 	rrpc "github.com/athanorlabs/go-relayer/rpc"
 
 	"github.com/athanorlabs/atomic-swap/cliutil"
+	"github.com/athanorlabs/atomic-swap/coins"
 	"github.com/athanorlabs/atomic-swap/common"
 	swapnet "github.com/athanorlabs/atomic-swap/net"
 
@@ -45,11 +46,11 @@ const (
 	flagDeploy           = "deploy"
 	flagLog              = "log-level"
 	// TODO: do we need this, or can we assume all swap relayers will need to be on the p2p network?
-	flagWithNetwork       = "with-network"
-	flagLibp2pKey         = "libp2p-key"
-	flagLibp2pPort        = "libp2p-port"
-	flagBootnodes         = "bootnodes"
-	flagRelayerCommission = "relayer-commission"
+	flagWithNetwork = "with-network"
+	flagLibp2pKey   = "libp2p-key"
+	flagLibp2pPort  = "libp2p-port"
+	flagBootnodes   = "bootnodes"
+	flagRelayerFee  = "relayer-fee"
 
 	defaultLibp2pPort = 10900
 )
@@ -118,10 +119,10 @@ var (
 			EnvVars: []string{"SWAPD_BOOTNODES"},
 		},
 		&cli.StringFlag{
-			Name: flagRelayerCommission,
-			Usage: "Minimum commission percentage (of the swap value) to receive:" +
-				" eg. --relayer-commission=0.01 for 1% commission",
-			Value: common.DefaultRelayerCommission.Text('f'),
+			Name: flagRelayerFee,
+			Usage: "Minimum commission fee to receive in ETH:" +
+				" eg. --relayer-commission=0.01 for a fee of 0.01 ETH",
+			Value: common.DefaultRelayerFee.Text('f'),
 		},
 	}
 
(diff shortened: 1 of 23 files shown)


cmd/swapcli/errors.go
```diff
@@ -4,11 +4,10 @@ import (
 	"fmt"
 )
 
-//nolint:lll
 var (
-	errNoDuration                         = fmt.Errorf("must provide non-zero --duration")
-	errCannotHaveGreaterThan100Commission = fmt.Errorf("%s must be less than 1", flagRelayerCommission)
-	errMustSetRelayerEndpoint             = fmt.Errorf("%s must be set if %s is set", flagRelayerEndpoint, flagRelayerCommission)
+	errNoDuration                  = fmt.Errorf("must provide non-zero --duration")
+	errCannotHaveGreaterThan100Fee = fmt.Errorf("%s must be less than 1", flagRelayerFee)
```
dimalinux, line 9: Can we make the error message, the check, and the name of the error agree?  The message says that it must be strictly less than, but the check is less than or equal to 100%.  You could call the error `errFeeNotLessThan100Percent` or `errFeeGreaterThan100Percent` depending on the check. Or just `errRelayerFeeTooHigh`.

cmd/relayer/main.go
```diff
@@ -118,10 +119,10 @@ var (
 			EnvVars: []string{"SWAPD_BOOTNODES"},
 		},
 		&cli.StringFlag{
-			Name: flagRelayerCommission,
-			Usage: "Minimum commission percentage (of the swap value) to receive:" +
-				" eg. --relayer-commission=0.01 for 1% commission",
-			Value: common.DefaultRelayerCommission.Text('f'),
+			Name: flagRelayerFee,
+			Usage: "Minimum commission fee to receive in ETH:" +
+				" eg. --relayer-fee=0.01 for a fee of 0.01 ETH",
```
dimalinux, line 124: I think we need slightly different wording so the user understands that this is a filter, under which the relayer rejects the request. The word "receive" sounds passive. Maybe something like this?
```
Min fee to accept a request: eg. --relayer-fee=0.01 rejects requests under 0.01 ETH
```

cmd/relayer/validate.go
```diff
@@ -70,18 +68,18 @@ var (
 )
 
 type validator struct {
-	ctx               context.Context
-	ec                *ethclient.Client
-	relayerCommission *apd.Decimal
-	forwarderAddress  ethcommon.Address
+	ctx              context.Context
+	ec               *ethclient.Client
+	relayerFee       *big.Int
```
dimalinux, line 73: Maybe `minFee` or `minRelayerFee` to clarify its purpose?

cmd/relayer/main.go
```diff
@@ -45,11 +46,11 @@ const (
...
-	flagWithNetwork       = "with-network"
-	flagLibp2pKey         = "libp2p-key"
-	flagLibp2pPort        = "libp2p-port"
-	flagBootnodes         = "bootnodes"
-	flagRelayerCommission = "relayer-commission"
+	flagWithNetwork = "with-network"
+	flagLibp2pKey   = "libp2p-key"
+	flagLibp2pPort  = "libp2p-port"
+	flagBootnodes   = "bootnodes"
+	flagRelayerFee  = "relayer-fee"
```
dimalinux, line 53: What if we just called this `--min-fee`? It is the `relayer` executable, so the word `relayer` probably isn't adding any context for the end user.

docs/rpc.md
```diff
@@ -181,7 +181,7 @@ Parameters:
   zero address for regular ETH. default: regular ETH
 - `relayerEndpoint`: (optional) RPC endpoint of the relayer to use for submitting claim
   transactions.
-- `relayerCommission`: (optional) Commission in percentage that the relayer receives for
+- `relayerFee`: (optional) Fee in ETH that the relayer receives for
```
dimalinux, line 184: We might want to state that a the default fee is used if not specified, to prevent any user from thinking the fee itself is optional.  :)

Does we give a clear error message if the client passes a fee that is smaller than all registered relayers are willing to accept?

noot: updated!

> Does we give a clear error message if the client passes a fee that is smaller than all registered relayers are willing to accept?

this will require additional features on the relayer p2p network as discussed before, will leave it for later

dimalinux (approved):

PR #317, opened 2023-02-20
feat: add start/end times to `swap.Info` and swap RPC returns

### LINKED ISSUE #222: add start/completion time to swaps to allow for time-based ordering
add start and/or end times to `swap.Info` to allow for swaps to be sorted based on start/end time when they're retrieved via `GetPastIDs` for example.

noot:
-  add start/end times to `swap.Info`
- update `swap_getPastIDs` to return start/end times also and sort by start time
- update other swap RPC getters to return swap start times
- also update `swap_getOngoing` to return all ongoing swaps if no ID is provided

closes #222

### DIFF
--- a/cmd/swapcli/main.go
+++ b/cmd/swapcli/main.go
@@ -662,6 +662,7 @@ func runGetOngoingSwap(ctx *cli.Context) error {
 			receivedCoin = "XMR"
 		}
 
+		fmt.Printf("ID: %s\n", info.ID)
 		fmt.Printf("Start time: %s\n", info.StartTime)
 		fmt.Printf("Provided: %s %s\n", info.ProvidedAmount.Text('f'), info.Provided)
 		fmt.Printf("Receiving: %s %s\n", info.ExpectedAmount, receivedCoin)
--- a/cmd/swapcli/main.go
+++ b/cmd/swapcli/main.go
@@ -292,9 +292,9 @@ var (
 				},
 			},
 			{
-				Name:   "get-stage",
-				Usage:  "Get the stage of a current swap.",
-				Action: runGetStage,
+				Name:   "get-status",
+				Usage:  "Get the status of a current swap.",
+				Action: runGetStatus,
 				Flags: []cli.Flag{
 					&cli.StringFlag{
 						Name:     flagOfferID,
@@ -677,8 +677,8 @@ func runGetPastSwapIDs(ctx *cli.Context) error {
 		}
 
 		fmt.Printf("ID: %s\n", id.ID)
-		fmt.Printf("Start time: %s\n", id.StartTime)
-		fmt.Printf("End time: %s\n", id.EndTime)
+		fmt.Printf("Start time: %s\n", id.StartTime.Format(common.TimeFmtSecs))
+		fmt.Printf("End time: %s\n", id.EndTime.Format(common.TimeFmtSecs))
 	}
 
 	return nil
@@ -710,8 +710,8 @@ func runGetOngoingSwap(ctx *cli.Context) error {
 		}
 
 		fmt.Printf("ID: %s\n", info.ID)
-		fmt.Printf("Start time: %s\n", info.StartTime)
-		fmt.Printf("Provided: %s %s\n", info.ProvidedAmount.Text('f'), info.Provided)
+		fmt.Printf("Start time: %s\n", info.StartTime.Format(common.TimeFmtSecs))
+		fmt.Printf("Provided: %s %s\n", info.ProvidedAmount, info.Provided)
 		fmt.Printf("Receiving: %s %s\n", info.ExpectedAmount, receivedCoin)
 		fmt.Printf("Exchange Rate: %s ETH/XMR\n", info.ExchangeRate)
 		fmt.Printf("Status: %s\n", info.Status)
@@ -734,9 +734,9 @@ func runGetPastSwap(ctx *cli.Context) error {
 		receivedCoin = "XMR"
 	}
 
-	fmt.Printf("Start time: %s\n", info.StartTime)
-	fmt.Printf("End time: %s\n", info.EndTime)
-	fmt.Printf("Provided: %s %s\n", info.ProvidedAmount.Text('f'), info.Provided)
+	fmt.Printf("Start time: %s\n", info.StartTime.Format(common.TimeFmtSecs))
+	fmt.Printf("End time: %s\n", info.EndTime.Format(common.TimeFmtSecs))
+	fmt.Printf("Provided: %s %s\n", info.ProvidedAmount, info.Provided)
 	fmt.Printf("Receiving: %s %s\n", info.ExpectedAmount, receivedCoin)
 	fmt.Printf("Exchange Rate: %s ETH/XMR\n", info.ExchangeRate)
 	fmt.Printf("Status: %s\n", info.Status)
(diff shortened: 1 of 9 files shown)


protocol/swap/types_test.go
```diff
@@ -27,8 +27,12 @@ func Test_InfoMarshal(t *testing.T) {
 		200,
 		make(chan types.Status),
 	)
+	err := info.StartTime.UnmarshalJSON([]byte("\"2023-02-20T17:29:43.471020297-05:00\""))
```
dimalinux, line 30: You'd need 2 lines, but this is probably more elegant:
```
	var err error
	info.StartTime, err = time.Parse(time.RFC3339, "2023-02-20T17:29:43.471020297-05:00")
```
See here:
https://cs.opensource.google/go/go/+/refs/tags/go1.19.6:src/time/time.go;l=1360

cmd/swapcli/main.go
```diff
@@ -620,33 +619,56 @@ func runGetPastSwapIDs(ctx *cli.Context) error {
...
+			fmt.Printf("---\n")
+		}
+
+		receivedCoin := "ETH"
+		if info.Provided == coins.ProvidesETH {
+			receivedCoin = "XMR"
+		}
+
+		fmt.Printf("ID: %s\n", info.ID)
+		fmt.Printf("Start time: %s\n", info.StartTime)
```
dimalinux, line 666: That is going to be a pretty ugly string. Do we want to use `.Format(common.TimeFmtSecs)` or something similar?

cmd/swapcli/main.go
```diff
@@ -620,33 +619,56 @@ func runGetPastSwapIDs(ctx *cli.Context) error {
...
+		fmt.Printf("Provided: %s %s\n", info.ProvidedAmount.Text('f'), info.Provided)
+		fmt.Printf("Receiving: %s %s\n", info.ExpectedAmount, receivedCoin)
```
dimalinux, line 668: `.Text('f')` if you want to guarantee that exponent notation never gets used. The exchange rate below doesn't need it, as it is our own type.

dimalinux: Same thing for the identical line in `runGetPastSwap`.

cmd/swapcli/main.go
dimalinux: Whatever you use, use them on the other places we print start/end times above and below in this file.

docs/rpc.md
```diff
@@ -384,9 +385,14 @@ curl -s -X POST http://127.0.0.1:5000 -H 'Content-Type: application/json' -d \
...
 Parameters:
-- `offerID`: the swap's ID.
+- `offerID`: optional: the swap's ID.
 
 Returns:
+- `swaps`: a list of ongoing swaps. If an offerID is provided, this returns only the swap with that ID, if it exists.
+
+Each items in `swaps` contains:
+- `id`: the swap ID.
+- `startTime`: the start time of the swap.
```
dimalinux, line 395: Should we mention that it is in RFC 3339 format? That's what I did in `swap_suggestedExchangeRate`. I'm not crazy about the way I used parentheses at the end. Feel free to update those entries with whatever style you use here.

rpc/swap.go
```diff
@@ -44,22 +46,44 @@ func NewSwapService(
 	}
 }
 
+// PastSwapInfo contains a swap ID and its start and end time.
+type PastSwapInfo struct {
+	ID        string    `json:"id"`
```
dimalinux, line 51: I'd just leave this as `types.Hash` and remove the call to `.String()` below.

rpc/swap.go
```diff
@@ -44,22 +46,44 @@ func NewSwapService(
...
+	StartTime time.Time `json:"startTime"`
+	EndTime   time.Time `json:"endTime"`
```
dimalinux, line 53: It was on my TODO list to start adding validation tags on these methods. They are a lower priority, because they are not peer-to-peer messages. Would it be reasonable to add `validate:"required"` to all 3 of these fields now? It provides code documentation. If I see the tag on `startTime` and not on `endTime`, as an example, I'll know instantly that the `endTime` field might not always be present.

rpc/swap.go
```diff
@@ -44,22 +46,44 @@ func NewSwapService(
...
+}
+
 // GetPastIDsResponse ...
 type GetPastIDsResponse struct {
-	IDs []string `json:"ids"`
+	IDs []*PastSwapInfo `json:"ids"`
```
dimalinux, line 58: `PastSwaps` for the field name?  These are not IDs anymore.

It would be helpful if the request struct, response struct, RPC method name, and Go method names lined up.

I don't know the answer, but this is probably a good time to ask: Do we want this function at the current time? The implementation is pulling up the entire swap info data structure from disk already to create this list. Given that the reply is a local host transfer, we are not saving a measurable amount of resources by reducing the data before sending it. At some point, especially if we have memory constrained clients, we'll probably want some solution, but now might be too early to optimize. The desired client solution might just be a paging system layered on top of a method that returns the full data structures.

rpc/swap.go
```diff
@@ -174,8 +235,9 @@ type GetStageRequest struct {
 
 // GetStageResponse ...
 type GetStageResponse struct {
-	Stage string `json:"stage"`
-	Info  string `json:"info"`
+	Stage     string    `json:"stage"`
```
dimalinux, line 238: Can we update the type to `types.Status`? I incorrectly assumed this was the contract stage until I looked down at where it is filled in.

We really need some renames if it is not too far out of scope for this PR. Based on the name, I would expect the next field to come from `swap.Info`, but it is actually a textual `description` of the Status.

dimalinux: If we need `MarshalText` / `UnmarshalText` methods for the Status type, I can add them to this PR.

rpc/swap.go
noot: yeah, it would be reasonable to mark these all as required, do you want me to do that here or will you do it in another PR?

rpc/swap.go
noot: do you mean to return all past swap info in this method, instead of just the IDs and start/end time? I think this is useful for quickly seeing your past swaps and checking on their status individually based on time, you might not want to see every swap's full info.

rpc/swap.go
noot: updated. I'd rather do other changes in another PR since I'd like to get this in asap as I've been using it for testing

rpc/swap.go
dimalinux: Let's start doing it anywhere we can now, especially if we are modifying a serialized structure or creating a new one. It will make for less work and code review when I add it to the others.

rpc/swap.go
dimalinux: Yes, but the client (or `jq` if you are using curl) can be used to filter the fields that you want to see. I was just thinking if we wanted to start simple and only implement one of the variations, returning the full version is probably more useful, and given the current implementation, it has a negligible performance difference. With either version, we may eventually need a paging or time based filter in the request, but that is something that can easily be added later, and adding it can even be done without breaking existing clients.

rpc/swap.go
dimalinux: I added the Status marshalling support here:
https://github.com/AthanorLabs/atomic-swap/pull/323

I tried to do it in a way that would create minimal merge conflicts.

cmd/swapcli/main.go
dimalinux: It looks like you removed the .Text('f') call on the `ProvidedAmount` instead of adding it to the `ExpectedAmount`. I was suggesting that we add it to both. Were you thinking that exponent notation is fine?

I'm playing around to see when `apd.Decimal` adds the exponent notation. At least in this example, it switches over the exponent notation at 7 decimals:

```
func TestExample(t *testing.T) {
	amt := coins.StrToDecimal("0.000001")
	require.Equal(t, "0.000001", amt.String())

	amt = coins.StrToDecimal("0.0000001")
	require.Equal(t, "1E-7", amt.String())
}
```

docs/rpc.md
```diff
@@ -384,9 +385,14 @@ curl -s -X POST http://127.0.0.1:5000 -H 'Content-Type: application/json' -d \
 Gets information for the specified ongoing swap.
 
 Parameters:
-- `offerID`: the swap's ID.
+- `offerID`: optional: the swap's ID.
```
dimalinux, line 388: Can you use `(optional)` to match your syntax on the other examples?

dimalinux (approved):
Approved, but see unresolved comments for minor fix suggestions and consider merging in
https://github.com/AthanorLabs/atomic-swap/pull/323 first.

PR #318, opened 2023-02-20
feat: add `--check-code` flag to toggle code check on startup

noot:
- add `--check-code` flag to toggle code check on startup; defaults to true
- I added this because I was stuck in the case where the daemon starts up with updated contract code, but there was an ongoing swap that needed to be completed, but it couldn't because the contract code didn't match:
```
2023-02-20T21:08:35.469Z	ERROR	cmd	swapd/main.go:293	RPC/Websocket server exited: failed to create new swap state for ongoing swap, id 0xc89fbbfc5ccb11d6ba3336da554bebcb6e6d981a9e7f79d7b682905741c98bdb: cannot recover from swap where contract address is not the one loaded at start-up; please restart with --contract-address=0xaEc2Db3B24fab372AE93bA56f55F01B522C1Ee13
```

- but then starting the daemon with the above `--contract-address` fails because of the code check on startup.
- the easiest way to fix this was to add a flag to disable the code check

let me know if this is an adequate solution, or if there's another method you'd prefer.

### DIFF
--- a/cmd/swapd/contract.go
+++ b/cmd/swapd/contract.go
@@ -57,10 +57,10 @@ func getOrDeploySwapFactory(
 		}
 		log.Infof("loaded SwapFactory.sol from address %s", address)
 
-		_, err = contracts.CheckSwapFactoryContractCode(ctx, ec, address)
-		if err != nil {
-			return nil, ethcommon.Address{}, err
-		}
+		// _, err = contracts.CheckSwapFactoryContractCode(ctx, ec, address)
+		// if err != nil {
+		// 	return nil, ethcommon.Address{}, err
+		// }
 	}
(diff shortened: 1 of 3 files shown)


dimalinux:
What if we just handled this with release versioning? Maybe protocol version and contract changes trigger a new major release. And we make it clear in the documentation that users should complete all existing swaps before upgrading to a new major version.

My concern about this solution is that it will be difficult to document. We'd have to explain to users when and why they would use the flag and also explain what risks they would be exposing themselves to by using it. If we keep the flag, we'd probably want to prevent any new swaps from being initiated.

dimalinux:
Decided not to implement

PR #319, opened 2023-02-20
fix: ensure stage is changed if funds are locked, even in error case

noot:
I encountered a bug when testing where if the counterparty exits the swap while the xmrmaker is locking funds, the protocol aborts on the xmrmaker side when it shouldn't (because funds were already locked). This was because the next expected event was being set *after* the handling of the event, so the handling errored and returned, but the next expected event wasn't updated to be `EventContractReady`.

This PR updates the behaviour and fixes that bug. This is also more correct since the stage should definitely be updated once locking occurs.

### DIFF
--- a/protocol/xmrmaker/event.go
+++ b/protocol/xmrmaker/event.go
@@ -190,15 +190,17 @@ func (s *swapState) handleEvent(event Event) {
 			return
 		}
 
-		err := s.setNextExpectedEvent(EventContractReadyType)
-		if err != nil {
-			e.errCh <- fmt.Errorf("failed to set next expected event to EventContractReadyType: %w", err)
-			return
-		}
-
-		err = s.handleEventETHLocked(e)
+		err := s.handleEventETHLocked(e)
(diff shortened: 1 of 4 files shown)


protocol/xmrmaker/event.go
```diff
@@ -193,7 +193,9 @@ func (s *swapState) handleEvent(event Event) {
 		err := s.handleEventETHLocked(e)
 		if err != nil {
 			e.errCh <- fmt.Errorf("failed to handle EventETHLocked: %w", err)
-			return
+			if !s.fundsLocked {
```
dimalinux, line 196: Is this variable only modified by a single go process?

dimalinux (approved):
I had one question, but I think the answer to it is "yes", in which case this looks good.

protocol/xmrmaker/event.go
noot: yep, all the swapState functions are accessed by one go process at a time.

PR #322, opened 2023-02-21
fix xmrmaker initiation + recovery, don't delete offer from db

noot:
Previously, xmrmaker recovery of an ongoing swap would fail, because it attempts to get the offer from the db (which it needs), but it was already deleted on swap initiation, causing the daemon to be unable to start up. Now the offer is just deleted from memory (so no one else can take it), but not from the db, so we can still get it on recovery.

### DIFF
--- a/protocol/xmrmaker/net.go
+++ b/protocol/xmrmaker/net.go
@@ -51,8 +51,9 @@ func (inst *Instance) initiate(
 		}
 	}
 
-	// checks passed, delete the offer for now
-	if err = inst.offerManager.DeleteOffer(offer.ID); err != nil {
+	// checks passed, delete the offer from memory for now
+	_, _, err = inst.offerManager.TakeOffer(offer.ID)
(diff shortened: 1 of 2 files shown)


protocol/xmrmaker/net.go
```diff
@@ -51,8 +51,9 @@ func (inst *Instance) initiate(
 		}
 	}
 
-	// checks passed, delete the offer for now
-	if err = inst.offerManager.DeleteOffer(offer.ID); err != nil {
+	// checks passed, delete the offer from memory for now
+	_, _, err = inst.offerManager.TakeOffer(offer.ID)
```
dimalinux, line 55: It looks like `TakeOffer` doesn't document the subtle nuance that it removes the entry from the cache while leaving it in the database.  Can we also add that directly on the `TakeOffer` method?

dimalinux (approved):
Looks good. One comment.

PR #327, opened 2023-02-22
dedicated monero address type (verses string)

### LINKED ISSUE #136: validate externally submitted xmr addresses based on network

dimalinux:
Adds custom type for monero addresses with better validations. Everything but environment validation is handled early, during JSON serialization and deserialization. Handles subaddresses. Closes #136.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,7 @@ linters-settings:
   maligned:
     # print struct with more effective memory layout or not, false by default
     suggest-new: true
+
   dupl:
     # tokens count to trigger issue, 150 by default
     threshold: 100
@@ -29,12 +30,19 @@ linters-settings:
     # Correct spellings using locale preferences for US or UK.
     # Default is to use a neutral variety of English.
     # Setting locale to US will correct the British spelling of 'colour' to 'color'.
+
+  tagliatelle:
+    # Check the struct tag name case.
+    case:
+      rules:
+        # See https://github.com/ldez/tagliatelle for values and examples:
+        json: goCamel
+
   gocritic:
     # Enable multiple checks by tags, run `GL_DEBUG=gocritic golangci-lint` run to see all tags and checks.
     # Empty list by default. See https://github.com/go-critic/go-critic#usage -> section "Tags".
     enabled-tags:
       - performance
-
     settings: # settings passed to gocritic
       captLocal: # must be valid enabled check name
         paramsOnly: true
@@ -64,6 +72,7 @@ linters:
     - predeclared
     - revive
     - staticcheck
+    - tagliatelle
     - typecheck
     - unconvert
     - unparam
--- a/crypto/monero/address.go
+++ b/crypto/monero/address.go
@@ -41,6 +41,7 @@ const (
 )
 
 var (
+	errAddressNotInitialized    = errors.New("monero address is not initialized")
 	errChecksumMismatch         = errors.New("invalid address checksum")
 	errInvalidAddressLength     = errors.New("invalid monero address length")
 	errInvalidAddressEncoding   = errors.New("invalid monero address encoding")
@@ -135,6 +136,10 @@ func (a *Address) Equal(b *Address) bool {
 // This validation can't be performed when decoding JSON, as the environment is
 // not known at that time.
 func (a *Address) ValidateEnv(env common.Environment) error {
+	if a == nil || a.decoded == new(Address).decoded {
+		return errAddressNotInitialized
+	}
+
 	switch a.Network() {
 	case Mainnet:
 		if env != common.Mainnet && env != common.Development {
--- a/crypto/monero/address.go
+++ b/crypto/monero/address.go
@@ -54,25 +54,27 @@ var (
 )
 
 // Address represents a base58-encoded string
-type Address [AddressBytesLen]byte
+type Address struct {
+	decoded [AddressBytesLen]byte
+}
 
 // NewAddress converts a string to a monero Address with validation.
-func NewAddress(addrStr string, env common.Environment) (Address, error) {
-	b58Bytes, err := moneroAddrBase58ToBytes(addrStr)
-	if err != nil {
-		return Address{}, err
+func NewAddress(addrStr string, env common.Environment) (*Address, error) {
+	addr := &Address{}
+	if err := addr.UnmarshalText([]byte(addrStr)); err != nil {
+		return nil, err
 	}
-	addr := Address(*(*[AddressBytesLen]byte)(b58Bytes))
+
 	return addr, addr.Validate(env)
 }
 
-func (a Address) String() string {
-	return moneroAddrBytesToBase58(a[:])
+func (a *Address) String() string {
+	return moneroAddrBytesToBase58(a.decoded[:])
 }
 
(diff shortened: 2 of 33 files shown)


.golangci.yml
```diff
@@ -29,12 +30,19 @@ linters-settings:
     # Correct spellings using locale preferences for US or UK.
     # Default is to use a neutral variety of English.
     # Setting locale to US will correct the British spelling of 'colour' to 'color'.
+
+  tagliatelle:
+    # Check the struct tag name case.
+    case:
+      rules:
+        # See https://github.com/ldez/tagliatelle for values and examples:
+        json: goCamel
```
dimalinux, line 39: Not in theme with the rest of the PR, but I found one structure that was still using snake case in the JSON tags. This will prevent it from happening in the future.

crypto/monero/address.go
```diff
@@ -20,83 +46,141 @@ var (
...
+		return Testnet
+	default:
+		// Our methods to deserialize and create Address values all verify
+		// that the address byte is valid
+		panic("address has invalid network prefix")
+	}
+}
+
+// Type returns the Address type
+func (a *Address) Type() AddressType {
```
dimalinux, line 91: I'm going to leave this here, but it doesn't actually appear to be needed at the current time. I thought it might be needed when the XMR Taker receives the NotifyXMRLock message. When that happens, I thought we might need to validate that the passed address is standard/primary, but that code is already comparing the received address with a value it generated from keys, which already guarantees that the address is standard/primary.

rpc/mocks_test.go
```diff
@@ -46,19 +46,19 @@ func (*mockNet) ConnectedPeers() []string {
 	panic("not implemented")
 }
 
-func (*mockNet) Discover(provides string, searchTime time.Duration) ([]peer.ID, error) {
+func (*mockNet) Discover(_ string, _ time.Duration) ([]peer.ID, error) {
```
dimalinux, line 49: I had to edit this file for the PR, and it was bothering me that all 3 different ways of dealing with unused method parameters were being mixed in the same file (sometimes even on the same method):
* giving the parameter a name
* omitting the parameter name
* using an underscore for the parameter name

In some places, multiple different ways were being use on the same function declaration.

protocol/xmrtaker/event_test.go
```diff
@@ -80,7 +80,7 @@ func TestSwapState_handleEvent_EventETHClaimed(t *testing.T) {
 
 	// test transferBack functionality
 	s.transferBack = true
-	s.Backend.SetBaseXMRDepositAddress(s.Backend.XMRClient().PrimaryAddress())
+	s.Backend.SetXMRDepositAddress(s.Backend.XMRClient().PrimaryAddress(), s.ID())
```
dimalinux, line 83: @noot Make sure this is correct. This fixed an error where monero was being claimed with a `transferBack` value of `true`, and a deposit address that was nil. In `newTestSwapStateAndNet` above, the swap is initially being setup with `transferBack` as `false` before we flip it to `true` on the line above.

crypto/monero/address.go
```diff
@@ -9,94 +9,183 @@ import (
...
+			return errInvalidPrefixGotMainnet
+		}
+	case Stagenet:
+		if env != common.Stagenet {
+			return errInvalidPrefixGotStagenet
+		}
+	case Testnet:
+		return errInvalidPrefixGotTestnet
+	default:
+		panic("unhandled network")
```
noot, line 155: don't think this should panic - I think a user can trigger this case?

crypto/monero/address_marshal.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil, err
+	}
+	return []byte(addrBytesToBase58(a.decoded[:])), nil
+}
+
+// UnmarshalText converts a base58 encoded monero address to our Address type.
+// The encoding, length and checksum are all validated, but not the network, as
+// it is unknown by the JSON parser. Empty strings are not allowed. Use an
+// address pointer in your serialized types if the Address is optional.
+func (a *Address) UnmarshalText(base58Input []byte) error {
```
noot, line 15: can you add a test for unmarshalling an address with the wrong length?

net/message/message.go
```diff
@@ -176,8 +176,8 @@ func (m *NotifyETHLocked) Type() byte {
 
 // NotifyXMRLock is sent by XMRMaker to XMRTaker after locking his XMR.
 type NotifyXMRLock struct {
-	Address string     `json:"address" validate:"required"` // address the monero was sent to
-	TxID    types.Hash `json:"txID"`                        // Monero transaction ID (transaction hash in hex)
+	Address *mcrypto.Address `json:"address" validate:"required"` // address the monero was sent to
+	TxID    types.Hash       `json:"txID"`                        // Monero transaction ID (transaction hash in hex)
```
noot, line 180: this should be required

protocol/xmrtaker/event_test.go
noot: this seems fine, did the test start failing or something? was the base XMR deposit address not set?

noot (commented):
overall looks good! just some minor comments.

protocol/xmrtaker/event_test.go
dimalinux: Yeah, it was failing in my PR. It looks like it worked in master, because the claim was pulling a deposit address using an all zero hash, but in my code the ID was not all zeros. I'm putting together a PR that greatly simplifies this deposit address code.

crypto/monero/address.go
dimalinux: They don't have a way to trigger it. Users outside of this package only have 3 ways to create an `Address`:
1. `new(Address)` - handled at the top of this function
2. `UnmarshalText` - verifies the network prefix when it calls `validateDecoded`
3. `NewAddress` -verifies the network prefix when it calls `UnmarshalText`

crypto/monero/address_marshal.go
dimalinux: There is a test:
```
func TestNewAddress_fail(t *testing.T) {
	_, err := NewAddress("fake", common.Mainnet)
	require.ErrorIs(t, err, errInvalidAddressLength)
}
```
We have full test coverage on the address specific code. The only lines that don't have coverage are unreachable code where I put in panics.

net/message/message.go
dimalinux: I'll add that in a follow-up PR this week. I want to add required everywhere possible, but adding it often requires unit test fixes, so I didn't go too crazy in this PR.

protocol/xmrtaker/event_test.go
dimalinux: Here's the add-on PR to simplify deposit addresses:
https://github.com/AthanorLabs/atomic-swap/pull/328

net/message/message.go
dimalinux: Here is the PR where the required tag will be added: https://github.com/AthanorLabs/atomic-swap/pull/330

noot (approved):

PR #330, opened 2023-02-24
added more required tags to serialized structs

dimalinux:
This was branched from `dimalinux/monero-address-type`. It will be retargeted to `master` after the PR for that branch is merged.

This PR adds required tags on all fields that shouldn't be empty on our serialized data structures.

### DIFF
--- a/net/initiate.go
+++ b/net/initiate.go
@@ -107,8 +107,10 @@ func (h *Host) handleProtocolStream(stream libp2pnetwork.Stream) {
 	}
 
 	if err := p2pnet.WriteStreamMessage(stream, resp, stream.Conn().RemotePeer()); err != nil {
-		log.Warnf("failed to send response to peer: err=%s", err)
-		_ = s.Exit()
+		log.Warnf("failed to send response to peer: %s", err)
+		if err = s.Exit(); err != nil {
+			log.Warnf("Swap exit failure: %s", err)
+		}
 		_ = stream.Close()
 		return
 	}
--- a/net/message/message.go
+++ b/net/message/message.go
@@ -103,13 +103,13 @@ func (m *QueryResponse) Type() byte {
 
 // SendKeysMessage is sent by both parties to each other to initiate the protocol
 type SendKeysMessage struct {
-	OfferID            types.Hash              `json:"offerID" validate:"required"`
+	OfferID            types.Hash              `json:"offerID"` // Not set by XMR Maker
 	ProvidedAmount     *apd.Decimal            `json:"providedAmount" validate:"required"`
 	PublicSpendKey     *mcrypto.PublicKey      `json:"publicSpendKey" validate:"required"`
 	PrivateViewKey     *mcrypto.PrivateViewKey `json:"privateViewKey" validate:"required"`
 	DLEqProof          string                  `json:"dleqProof" validate:"required"`
 	Secp256k1PublicKey *secp256k1.PublicKey    `json:"secp256k1PublicKey" validate:"required"`
-	EthAddress         ethcommon.Address       `json:"ethAddress" validate:"required"`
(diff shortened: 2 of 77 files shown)


net/message/message.go
```diff
@@ -103,13 +103,13 @@ func (m *QueryResponse) Type() byte {
...
 type SendKeysMessage struct {
-	OfferID            types.Hash              `json:"offerID"`
+	OfferID            types.Hash              `json:"offerID"` // Not set by XMR Maker
 	ProvidedAmount     *apd.Decimal            `json:"providedAmount" validate:"required"`
 	PublicSpendKey     *mcrypto.PublicKey      `json:"publicSpendKey" validate:"required"`
 	PrivateViewKey     *mcrypto.PrivateViewKey `json:"privateViewKey" validate:"required"`
 	DLEqProof          string                  `json:"dleqProof" validate:"required"`
 	Secp256k1PublicKey *secp256k1.PublicKey    `json:"secp256k1PublicKey" validate:"required"`
-	EthAddress         ethcommon.Address       `json:"ethAddress"`
+	EthAddress         ethcommon.Address       `json:"ethAddress"` // not set by XMR Taker
```
dimalinux, line 112: @noot The `offerID` and `ETHAddress` fields still don't line up. That might be an argument to make these 2 separate data structures (one for the taker and one for the maker).

net/initiate.go
```diff
@@ -107,8 +107,10 @@ func (h *Host) handleProtocolStream(stream libp2pnetwork.Stream) {
 	}
 
 	if err := p2pnet.WriteStreamMessage(stream, resp, stream.Conn().RemotePeer()); err != nil {
-		log.Warnf("failed to send response to peer: err=%s", err)
-		_ = s.Exit()
+		log.Warnf("failed to send response to peer: %s", err)
+		if err = s.Exit(); err != nil {
+			log.Warnf("Swap exit failure: %s", err)
+		}
```
dimalinux, line 113: For the most part, the PR stays on theme, but I added logging to the 3 places that were swallowing the bubbled up error from calls to the swap `Exit()` method. Without inspecting the error, we can't judge how critical the error is. There were places in previous PRs, where it was decided not to log critical errors because they were getting bubbled up. This means we need to ensure that we never silently swallow a bubbled-up error.

net/message/message.go
```diff
@@ -103,13 +103,13 @@ func (m *QueryResponse) Type() byte {
 
 // SendKeysMessage is sent by both parties to each other to initiate the protocol
 type SendKeysMessage struct {
-	OfferID            types.Hash              `json:"offerID"`
+	OfferID            types.Hash              `json:"offerID"` // Not set by XMR Maker
 	ProvidedAmount     *apd.Decimal            `json:"providedAmount" validate:"required"`
 	PublicSpendKey     *mcrypto.PublicKey      `json:"publicSpendKey" validate:"required"`
 	PrivateViewKey     *mcrypto.PrivateViewKey `json:"privateViewKey" validate:"required"`
-	DLEqProof          string                  `json:"dleqProof" validate:"required"`
+	DLEqProof          []byte                  `json:"dleqProof" validate:"required"`
```
dimalinux, line 110: This is what we discussed before. The proof is fairly large and sending it in base64 (automatic when using a byte slice) instead of base16, reduces the size by 4x.

protocol/txsender/external_sender.go
```diff
@@ -28,8 +29,8 @@ var (
 // Transaction represents a transaction to be signed by the front-end
 type Transaction struct {
 	To    ethcommon.Address
-	Data  string
-	Value string
+	Data  []byte
+	Value *apd.Decimal // ETH (or ETH asset), not WEI
```
noot, line 33: isn't this wei? `coins.NewWeiAmount` assumes the input is in wei

noot (approved):
looks good!

PR #331, opened 2023-02-25
add timeouts to `swap.Info`

### LINKED ISSUE #320: add t0/t1 (refund/claim times) to `swap.Info`
`swap.Info` should cache the swap t0/t1 times and return them in swap RPC calls. This provides more info to the user as to what's happening in their swap.

noot:
- add timeouts to `swap.Info`
- change `swap.Info.EndTime` to a pointer
closes #320

### DIFF
--- a/protocol/swap/types.go
+++ b/protocol/swap/types.go
@@ -37,7 +37,9 @@ type Info struct {
 	// MoneroStartHeight is the Monero block number when the swap begins.
 	MoneroStartHeight uint64            `json:"moneroStartHeight"`
 	StartTime         time.Time         `json:"startTime"`
-	EndTime           time.Time         `json:"endTime"`
+	EndTime           *time.Time        `json:"endTime,omitempty"`
+	Timeout0          *time.Time        `json:"timeout0,omitempty"`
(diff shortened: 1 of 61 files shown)


protocol/swap/types.go
```diff
@@ -37,7 +37,9 @@ type Info struct {
 	// MoneroStartHeight is the Monero block number when the swap begins.
 	MoneroStartHeight uint64            `json:"moneroStartHeight"`
 	StartTime         time.Time         `json:"startTime"`
-	EndTime           time.Time         `json:"endTime"`
+	EndTime           *time.Time        `json:"endTime,omitempty"`
+	Timeout0          *time.Time        `json:"timeout0,omitempty"`
+	Timeout1          *time.Time        `json:"timeout1,omitempty"`
```
dimalinux, line 42: Reader's are going to look at these 4 values together and assume that they are connected.  They'll incorrectly assume that `StartTime` is the block timestamp of the `newSwap` transaction and `EndTime` is the block timestamp of the `claim` transaction.  Can we add comments to clarify what each of the 4 timestamps represent?

dimalinux (approved):
Looks good, but let's document the timestamps before merging.

PR #334, opened 2023-02-27
switch to codecov-action@v3 for uploading code coverage

### LINKED ISSUE #321: Codecov updates
In our code coverage emails, we have this message:
> This organization is not using Codecov’s [GitHub App Integration](https://github.com/apps/codecov). We recommend you install it so Codecov can continue to function properly for your repositories.

We should also figure out how to add exclusions. It would be nice to exclude generated files, like the solidity Go bindings, from coverage.

dimalinux:
Closes #321

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -65,4 +65,3 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           verbose: true
-          fail_ci_if_error: true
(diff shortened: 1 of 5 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -61,5 +61,8 @@ jobs:
           ETH_MAINNET_ENDPOINT: ${{ secrets.ETH_MAINNET_ENDPOINT }}
         run: ./scripts/run-unit-tests.sh
 
-      - name: Run coverage
-        run: bash <(curl -s https://codecov.io/bash)
+      - name: Upload code coverage
+        uses: codecov/codecov-action@v3
+        with:
+          verbose: true
+          fail_ci_if_error: true
```
dimalinux, line 68: If this line becomes problematic, we can remove it.

noot (approved):

PR #341, opened 2023-03-06
remove `NotifyXMRLocked` message, have xmrtaker check for lock independently

### LINKED ISSUE #335: remove address field from `NotifyXMRLock`
we can probably remove the `address` field, as each party can calculate the xmr lock address independently after the `SendKeysMessage` exchange.

noot:
as title says, we don't need to worry about the balance unlock time anymore either, as the xmrtaker only notifies itself of the event `EventXMRLocked` once the unlocked balance is >= the expected amount to receive.

closes #335

### DIFF
--- a/protocol/xmrmaker/event.go
+++ b/protocol/xmrmaker/event.go
@@ -251,12 +251,7 @@ func (s *swapState) handleEvent(event Event) {
 }
 
 func (s *swapState) handleEventETHLocked(e *EventETHLocked) error {
-	resp, err := s.handleNotifyETHLocked(e.message)
-	if err != nil {
-		return err
-	}
-
-	return s.SendSwapMessage(resp, s.ID())
+	return s.handleNotifyETHLocked(e.message)
 }
 
 func (s *swapState) handleEventContractReady() error {
--- a/protocol/xmrmaker/event.go
+++ b/protocol/xmrmaker/event.go
@@ -190,7 +190,7 @@ func (s *swapState) handleEvent(event Event) {
 			return
 		}
 
-		err := s.handleEventETHLocked(e)
+		err := s.handleNotifyETHLocked(e.message)
 		if err != nil {
 			e.errCh <- fmt.Errorf("failed to handle EventETHLocked: %w", err)
 			if !s.fundsLocked {
@@ -248,10 +248,6 @@ func (s *swapState) handleEvent(event Event) {
 	default:
 		panic("unhandled event type")
 	}
-}
-
-func (s *swapState) handleEventETHLocked(e *EventETHLocked) error {
-	return s.handleNotifyETHLocked(e.message)
 }
 
 func (s *swapState) handleEventContractReady() error {
--- a/protocol/xmrtaker/event.go
+++ b/protocol/xmrtaker/event.go
@@ -246,7 +246,7 @@ func (s *swapState) handleEvent(event Event) {
 			return
 		}
 
-		err := s.handleEventXMRLocked()
+		err := s.handleNotifyXMRLock()
 		if err != nil {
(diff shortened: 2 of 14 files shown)


protocol/xmrmaker/event.go
```diff
@@ -251,12 +251,7 @@ func (s *swapState) handleEvent(event Event) {
 }
 
 func (s *swapState) handleEventETHLocked(e *EventETHLocked) error {
-	resp, err := s.handleNotifyETHLocked(e.message)
-	if err != nil {
-		return err
-	}
-
-	return s.SendSwapMessage(resp, s.ID())
+	return s.handleNotifyETHLocked(e.message)
```
dimalinux, line 254: Should the caller just invoke `handleNotifyETHLocked` directly? I'm looking at the calling code, and it seems counterintuitive that the error channel is passed here, but any returned error is only pushed on to the channel by the calling code.

protocol/xmrtaker/message_handler.go
```diff
@@ -148,6 +143,52 @@ func (s *swapState) handleSendKeysMessage(msg *message.SendKeysMessage) (common.
 	return out, nil
 }
 
+func (s *swapState) checkForXMRLock() {
+	// monero block time is >1 minute, so this should be fine
+	const checkForXMRLockInterval = time.Minute
```
dimalinux, line 148: It might be nice if we could lower this for testing.  We could have a `BlockInterval` variable defined to 2 minutes in the `monero` package.  Then in `TestBackgroundMineBlocks`, it could get lowered to 1 second.  Then this constant here could be half the block interval.  (Just throwing out ideas ... you may have a better one.) What I described above, is similar to what we do with the `blockSleepDuration` variable.

Another idea: You could call `WaitForBlocks` passing 1 for the number of blocks. That method handles the context being cancelled.

protocol/xmrtaker/message_handler.go
```diff
@@ -148,6 +143,52 @@ func (s *swapState) handleSendKeysMessage(msg *message.SendKeysMessage) (common.
...
+	abViewCli, err := monero.CreateViewOnlyWalletFromKeys(conf, vk, lockedAddr, s.walletScanHeight)
+	if err != nil {
+		log.Errorf("failed to generate view-only wallet to verify locked XMR: %s", err)
+		return
+	}
+	defer abViewCli.CloseAndRemoveWallet()
+
+	log.Debugf("generated view-only wallet to check funds: %s", abViewCli.WalletName())
+
+	timer := time.NewTimer(checkForXMRLockInterval)
```
dimalinux, line 163: We have a `SleepWithContext` method that might be simpler.

protocol/xmrmaker/event.go
noot: updated

protocol/xmrtaker/event.go
```diff
@@ -248,7 +246,7 @@ func (s *swapState) handleEvent(event Event) {
 			return
 		}
 
-		err := s.handleEventXMRLocked(e)
+		err := s.handleEventXMRLocked()
```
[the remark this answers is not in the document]
noot: updated

protocol/xmrtaker/message_handler.go
noot: I updated this function to just set the interval to 1s if we're on development. I decided against `WaitForBlocks` since it returns errors in other cases than just context cancel which isn't necessarily what we want here

protocol/xmrtaker/message_handler.go
noot: I updated this to use a ticker, left it as is since I think it's a bit more readable and explicit

dimalinux (approved):

Issue #324, opened 2023-02-21
feature to list all relayers and their requested fee


dimalinux:
This issue is to implement all the needed backend support for swapcli to be able to list relayers with their request fee.

dimalinux:
With https://github.com/AthanorLabs/atomic-swap/pull/342 this can be done using:
```
./bin/swapcli discover --provides relayer
```
We could just update the documentation on the `--provides` flag to mention it, unless we want something more sophisticated.

noot:
going to close this for now, as all relayers are using the same fee (and can be discovered w the above)

PR #342, opened 2023-03-06
integrated relayer into swapd

### LINKED ISSUE #329: TestSwapState_ClaimRelayer_ETH test is flaky
This test is failing with some frequency even outside of CI:
```
=== RUN   TestSwapState_ClaimRelayer_ETH
    claim_test.go:147: gas cost to deploy Forwarder.sol: 1998934
    claim_test.go:153: gas cost to call RegisterDomainSeparator: 50391
    claim_test.go:163: gas cost to deploy SwapFactory.sol: 2075776
    claim_test.go:188: gas cost to call new_swap: 53007
    claim_test.go:219: gas cost to call SetReady: 33660
    claim_test.go:239: 
        >-------Error Trace:>---/home/dh/dev/atomic-swap/protocol/xmrmaker/claim_test.go:239
        >-------            >--->------->------->-------/home/dh/dev/atomic-swap/protocol/xmrmaker/claim_test.go:115
        >-------Error:      >---Received unexpected error:
        >-------            >---failed to submit transaction to relayer: message=read tcp 127.0.0.1:42754->127.0.0.1:8545: i/o timeout; code=-32000; data=map[]
        >-------Test:       >---TestSwapState_ClaimRelayer_ETH
--- FAIL: TestSwapState_ClaimRelayer_ETH (401.07s)
```

### LINKED ISSUE #337: relayer data dir issues
### How {ENV} for the data-dir is determined (flag vs chain-ID)

The relayer's default value for `--data-dir`, as shown in the help, is `{HOME}/.atomicswap/{ENV}`. This left me confused, as I couldn't find the `--env` flag for the relayer. I had to dig through the code to determine that the value of ENV is derived from the chain ID of the ethereum endpoint.

Options to make our tools consistent:
(1) Add an `--env` flag to the relayer and validate that the chain ID of the ethereum endpoint matches the environment. (Current behavior of `swapd`.)
(2) Remove the `--env` flag from `swapd` and derive it from the chain ID of the ethereum endpoint. (Current behavior of the `realyer`.)

### Avoiding `relayer`/`swapd` conflicts

After determining what `{ENV}` was for the `relayer`, I was then confused as to the ETH private key location.  The help says it is in `{DATA_DIR}/eth.key`, which converts to `{HOME}/.atomicswap/{ENV}/eth.key` if you didn't pass a `--data-dir` flag. This is the same location as `swapd`'s key, which had me concerned, as swapd and the relayer can't share the same key for nonce synchronization reasons.  Digging through the code, I realized that the help lied, and that the default data-dir for the relayer is actually `{HOME}/.atomicswap/{ENV}/relayer` making this another inconsistency between our tools.

Here are some possible pairs that would make our data-dir locations consistent between programs:
1. `{HOME}/.atomicswap/{ENV}/relayer`
`{HOME}/.atomicswap/{ENV}/swapd`
1. `{HOME}/.atomicswap/relayer/{ENV}`
`{HOME}/.atomicswap/swapd/{ENV}`
1. `{HOME}/.atomic/{ENV}/swapd`
`{HOME}/.atomic/{ENV}/relayer`
1. `{HOME}/.atomic/swapd/{ENV}`
`{HOME}/.atomic/relayer/{ENV}`

### LINKED ISSUE #338: relayer should give error on non-flag arguments
I had a script that accidentally omitted the dashes on an argument. The expected behavior would be a failure with a message about the non-flag argument, but it was silently ignored. It was crazy confusing to debug, because flag arguments after a non-flag argument are ignored, so the error message was about not being able to connect to the wrong ETH endpoint even though the script error was on the `--log-level` flag.

We just need to apply the same fix to the relayer that was done for `swapd` in this PR:
https://github.com/AthanorLabs/atomic-swap/pull/311/files

### LINKED ISSUE #339: relayer not using stagenet bootnodes
I thought the plan was that relayer instances for goerli and mainnet would use our swapd instances for those networks as bootnodes, but there doesn't appear to be any wiring in place for that to happen.

### LINKED ISSUE #340: remove uneeded flags from the relayer
# --deploy flag

Can we get rid of this? I don't see the use-case on the relayer given that `swapd` deploys the forwarder, and the forwarder address needs to be embedded in the SwapFactory contract. Providing multiple ways of doing the exact same thing leads to end user confusion about possible differences, as well as increased documentation, testing and maintenance costs.

# --with-network

I think we decided that relayers are only going to be accessed remotely using libp2p, so I'd like to remove this flag. If we kept it, I'd ask to rename it to `--enable-libp2p` to clarify it's purpose.

dimalinux:
This PR removes the separate relayer binary in favor of tight relay integration with swapd. The tight integration has many advantages:

* Lowers barriers to becoming a relayer. We want a large percentage of the network participating, not a small, centralized set of relayers. In this PR, explicit opt-in is still required, but long-term we should strongly consider making it opt-in by default (and opt-out explicit). Not in this PR, but in follow up work, Alice should always be willing to relay for her swap with Bob if she can pay for the gas, even if she explicitly opted out of relaying for others. 
* Less code and maintenance for us. Examples: No need to write a balance API, we already have it. No need write a new client tool to adjust minimum relay fees, it can be done using `swapcli`.
* Doesn't require a separate ethereum wallet to be a relayer, we can prevent nonce reuse when integrated.
* We were already going to use the swapd bootnodes for relayers and swapd was already deploying the forwarder contract. Having the rest of the relay functionality integrated is, arguably, more logically consistent than having it separated.

Addresses or obsoletes the need for:
* resolves #337
* resolves #338 
* resolves #339 
* resolves #340
* resolves #329
* resolves #342

Part of this PR are `go-p2p-net` changes that will need to be merged in tandem: https://github.com/AthanorLabs/go-p2p-net/pull/4.  Right now this PR's branch points to the head of the `dimalinux/callback-for-namespaces` branch in the `go-p2p-net` repo.


cmd/swapcli/main.go
```diff
@@ -175,14 +178,11 @@ var (
...
 						Usage: "Ethereum ERC-20 token address to receive, or the zero address for regular ETH",
 					},
-					&cli.StringFlag{
-						Name:  flagRelayerEndpoint,
-						Usage: "HTTP RPC endpoint of relayer to use for claiming funds. No relayer is used if this is not set",
-					},
 					&cli.StringFlag{
 						Name: flagRelayerFee,
-						Usage: "Fee to pay the relayer in ETH:" +
+						Usage: "Fee to pay the relayer in ETH if you have insufficient funds to claim:" +
```
dimalinux, line 183: When I wrote that phrase, I thought were not relaying if Bob already has enough ether in the account being claimed into. I need to look into it more. If we do it that way, which I don't think we are, it will make our integration testing hard, as you'll need a swapd instance with no ETH.

common/consts.go
```diff
@@ -38,7 +36,3 @@ const (
 	GanacheChainID = 1337
 	HardhatChainID = 31337
 )
-
-// DefaultRelayerFee is the default fee for swap relayers.
-// It's set to 0.009 ETH.
-var DefaultRelayerFee = apd.New(9, -3)
```
dimalinux, line 44: I moved it to the `relay` package.

common/rpctypes/types.go
```diff
@@ -78,12 +78,11 @@ type TakeOfferRequest struct {
 
 // MakeOfferRequest ...
 type MakeOfferRequest struct {
-	MinAmount       *apd.Decimal        `json:"minAmount" validate:"required"`
-	MaxAmount       *apd.Decimal        `json:"maxAmount" validate:"required"`
-	ExchangeRate    *coins.ExchangeRate `json:"exchangeRate" validate:"required"`
-	EthAsset        types.EthAsset      `json:"ethAsset,omitempty"`
-	RelayerEndpoint string              `json:"relayerEndpoint,omitempty"`
```
dimalinux, line 85: I removed support for non-p2p relayers. In this PR, Bob just uses relayers in the order of discovery, but it might make sense to first request from Alice.

net/host.go
```diff
@@ -19,9 +19,11 @@ import (
...
 
 const (
-	// ProtocolID is the base atomic swap network protocol ID.
-	ProtocolID     = "/atomic-swap/0.2"
-	maxMessageSize = 1 << 17
+	// ProtocolID is the base atomic swap network protocol ID prefix. The full ID
+	// includes the chain ID at the end.
+	ProtocolID          = "/atomic-swap/0.2"
+	maxMessageSize      = 1 << 17            // TODO: implement separate sizes for query/relay protocols
+	maxRelayMessageSize = maxMessageSize / 2 // TODO: optimize
```
dimalinux, line 26: Reminder to self: Look at relayer message sizes to figure out what a good value above would be.

net/host.go
```diff
@@ -61,16 +66,17 @@ type Host struct {
 // NewHost returns a new Host.
 // The host implemented in this package is swap-specific; ie. it supports swap-specific
 // messages (initiate and query).
-func NewHost(cfg *p2pnet.Config) (*Host, error) {
+func NewHost(cfg *p2pnet.Config, isRelayer bool) (*Host, error) {
```
dimalinux, line 69: This was a little awkward. The config passed in is for the "base class" from another repo, so I couldn't add a new field.

net/host.go
```diff
@@ -79,25 +85,37 @@ func (h *Host) P2pHost() P2pHost {
...
+	h.makerHandler = makerHandler
+	h.takerHandler = takerHandler
+
 	fn := func() bool {
-		return len(handler.GetOffers()) == 0
+		return len(makerHandler.GetOffers()) > 0 || h.isRelayer
 	}
-
-	h.handler = handler
 	h.h.SetShouldAdvertiseFunc(fn)
```
dimalinux, line 96: I think we got the API on this wrong. Advertising always happens with the empty ("") namespace if nothing else. Instead of having the function ask if we should advertise, it should be asking *what* we should be advertising. That will have to be done in follow up work that updates the other repo though.

net/host.go
```diff
@@ -79,25 +85,37 @@ func (h *Host) P2pHost() P2pHost {
...
 	h.h.SetStreamHandler(swapID, h.handleProtocolStream)
-	return h.h.Start()
+
+	// Note: Start() is non-blocking
+	if err := h.h.Start(); err != nil {
+		return err
+	}
+
+	if h.isRelayer {
+		h.Advertise(nil)
```
dimalinux, line 115: If we don't call it here, we won't advertise that we are a relayer until the first XMR offer is made. Proposed fix for this was mentioned earlier.

net/host.go
```diff
@@ -133,6 +151,9 @@ func (h *Host) CloseProtocolStream(id types.Hash) {
 
 // Advertise advertises in the DHT.
 func (h *Host) Advertise(strs []string) {
+	if h.isRelayer {
+		strs = append(strs, relayerProvidesStr)
```
dimalinux, line 155: We don't want the offer manager overwriting our desire to advertise that we are a relayer, and it doesn't seem like the offer manager should need to know that we are a relayer, so I had to put this hack in. We'll have to give more thought to these APIs.

cmd/swapcli/main.go
```diff
@@ -19,29 +19,32 @@ import (
...
-	flagMinAmount       = "min-amount"
-	flagMaxAmount       = "max-amount"
-	flagPeerID          = "peer-id"
-	flagOfferID         = "offer-id"
-	flagOfferIDs        = "offer-ids"
-	flagExchangeRate    = "exchange-rate"
-	flagProvides        = "provides"
-	flagProvidesAmount  = "provides-amount"
-	flagRelayerFee      = "relayer-fee"
-	flagRelayerEndpoint = "relayer-endpoint"
```
dimalinux, line 39: `relayer-endpoint` was removed, the rest are the same.

cmd/swapcli/main.go
```diff
@@ -19,29 +19,32 @@ import (
...
+	flagSwapdPort      = "swapd-port"
+	flagMinAmount      = "min-amount"
+	flagMaxAmount      = "max-amount"
+	flagPeerID         = "peer-id"
+	flagOfferID        = "offer-id"
+	flagOfferIDs       = "offer-ids"
+	flagExchangeRate   = "exchange-rate"
+	flagProvides       = "provides"
+	flagProvidesAmount = "provides-amount"
+	flagRelayerFee     = "relayer-fee"
```
noot, line 39: if there's only one relayer-fee flag, how do we differentiate between the fee we want to pay for a relayed tx vs the fee we want to receive as a relayer?

cmd/swapcli/main.go
noot: yeah maybe revert it back for now? what's the current logic for relayer/vs no relayer?

noot: is it basically if the relayerFee is set, or if balance is zero?

cmd/swapcli/main.go
```diff
@@ -549,17 +549,17 @@ func runMake(ctx *cli.Context) error {
...
+		relayerFee, err = cliutil.ReadUnsignedDecimalFlag(ctx, flagRelayerFee)
+		if err != nil {
 			return err
 		}
-	} else if ctx.IsSet(flagRelayerFee) {
-		return errMustSetRelayerEndpoint
-	}
-	if relayerFee.Cmp(apd.New(1, 0)) > 0 {
-		return errRelayerFeeTooHigh
+		if relayerFee.Cmp(minRelayerFee) < 0 || relayerFee.Cmp(maxRelayerFee) > 0 {
```
noot, line 560: why do we need the `rrelayerFee >= minRelayerFee` check? a relayer could still manually set a lower fee, I think we should warn users potentially but not reject

noot: also related to #324

noot: ah I see you removed any way for the relayer to configure their fee.

cmd/swapd/main.go
```diff
@@ -76,6 +75,7 @@ const (
 	flagGasPrice             = "gas-price"
 	flagGasLimit             = "gas-limit"
 	flagUseExternalSigner    = "external-signer"
+	flagClaimRelayer         = "claim-relayer"
```
noot, line 78: not really a fan of this flag, I'd just call it `relayer` or `is-relayer` or something?

cmd/swapd/main.go
```diff
@@ -203,6 +203,11 @@ var (
 				Name:  flagUseExternalSigner,
 				Usage: "Use external signer, for usage with the swap UI",
 			},
+			&cli.BoolFlag{
+				Name:  flagClaimRelayer,
+				Usage: "Relay claims for other XMR makers and earn a modest fee per relayed transaction",
```
noot, line 208:
```suggestion
				Usage: "Relay transactions for other XMR makers and earn a fee (0.009 ETH) per relayed transaction",
```

if there is no way for the relayer to set their fee, I'd just put it in the description

net/host.go
```diff
@@ -31,27 +35,30 @@ type P2pHost interface {
 	Start() error
 	Stop() error
 
-	Advertise([]string)
+	RefreshNamespaces()
```
noot, line 38: I mentioned this in your other PR but I think this should be named `Advertise`

net/host.go
noot: you could make a config for this `Host` that accepts the `cfg` and `isRelayer`?

net/host.go
```diff
@@ -131,9 +157,10 @@ func (h *Host) CloseProtocolStream(id types.Hash) {
 	_ = swap.stream.Close()
 }
 
-// Advertise advertises in the DHT.
-func (h *Host) Advertise(strs []string) {
-	h.h.Advertise(strs)
+// RefreshNamespaces forces an immediate update to the advertised namespaces in
+// the DHT
+func (h *Host) RefreshNamespaces() {
+	h.h.RefreshNamespaces()
```
noot, line 163: not necessarily, it only adds new "namespaces" immediately, but doesn't purge ones if you decide to not readvertise them. I'd call this `Advertise`

net/message/message.go
```diff
@@ -19,7 +19,10 @@ import (
 // Identifiers for our p2p message types. The first byte of a message has the
 // identifier below telling us which type to decode the JSON message as.
 const (
+	Unknown                = iota // occupies the uninitialized value
 	QueryResponseType byte = iota
```
noot, line 23: remove the iota from this line?

net/types.go
```diff
@@ -12,17 +12,25 @@ type SwapState = common.SwapStateNet //nolint:revive
...
+type MakerHandler interface {
 	GetOffers() []*types.Offer
-	HandleInitiateMessage(msg *SendKeysMessage) (s SwapState, resp Message, err error)
+	HandleInitiateMessage(msg *SendKeysMessage) (SwapState, Message, error)
+}
+
+// TakerHandler handles relay claim requests. It is implemented by
+// *xmrtaker.xmrtaker.
+type TakerHandler interface {
+	HandleRelayClaimRequest(msg *RelayClaimRequest) (*RelayClaimResponse, error)
```
noot, line 33: why is the relayer part of the taker? it shouldn't be, as it's not actually related logically; it should be its own instance

protocol/backend/backend.go
```diff
@@ -28,13 +26,8 @@ import (
 type MessageSender interface {
```
noot, line 26: I feel like this struct name/comment can be updated

protocol/xmrmaker/backend_offers.go
```diff
@@ -24,12 +24,21 @@ func (b *Instance) MakeOffer(
...
 	}
 
-	extra, err := b.offerManager.AddOffer(o, relayerEndpoint, relayerFee)
+	if relayerFee != nil {
+		if o.EthAsset != types.EthAssetETH {
+			return nil, errRelayingWithNonEthAsset
+		}
+		if relayerFee.Cmp(relayer.MinRelayerFeeEth) < 0 {
+			return nil, errRelayerFeeTooLow{relayer.MinRelayerFeeEth, relayerFee}
+		}
```
noot, line 33: if we're going to force the relayer fee to be fixed, we should probably just remove the ability for someone to set the fee they want to pay.

protocol/xmrtaker/instance.go
```diff
@@ -245,3 +249,22 @@ func (inst *Instance) ExternalSender(offerID types.Hash) (*txsender.ExternalSend
 
 	return es, nil
 }
+
+// HandleRelayClaimRequest validates and sends the transaction for a relay claim request
+func (inst *Instance) HandleRelayClaimRequest(request *message.RelayClaimRequest) (*message.RelayClaimResponse, error) {
```
noot, line 254: mentioned in another comment, but this isn't related to taker logic, so seems weird that it's here.

relayer/claim_request.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/athanorlabs/atomic-swap/coins"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+	"github.com/athanorlabs/atomic-swap/net/message"
+)
+
+// DefaultRelayerFeeWei is the default fee for swap relayers.
+// It's set to 0.009 ETH. Currently, the minimum and default
+// values are identical.
+var (
+	DefaultRelayerFeeWei = big.NewInt(9e15)
```
noot, line 22: I didn't know go had this syntax lol, that's cool!!

protocol/xmrmaker/backend_offers.go
dimalinux: In the name of simplicity, at least in the short term, I was thinking the same thing.  I can make that change. I wasn't ruling out the possibility of providing an API to set minimum relayer fees in the future. We'd need to design any complexity in carefully though. In the short term, we need to maximize the change that Bob can claim. Any design needs to include how we recover when Bob can't claim immediately.

relayer/forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+	calldata, err := getClaimRelayerTxCalldata(relayerFeeWei, swap, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	req := &gsnforwarder.IForwarderForwardRequest{
+		From:           swap.Claimer,
+		To:             swapFactoryAddress,
+		Value:          big.NewInt(0),
+		Gas:            big.NewInt(200000), // TODO: fetch from ethclient?
```
noot, line 86: can you set this to 52000 actually? a claim shouldn't be more than that; this will prevent the relayer from maybe doing something with our excess gas.

noot: a claim is usually just under 40000 actually, unless it's the very first claim in a contract

net/types.go
dimalinux: We could do that, but here's my reasoning and what I thought we would do in followup work that might be better done in the taker:

Bob should always go directly to Alice first for help with claiming and that, at least with our software, Alice always helps even if she doesn't want to help relay claims from other swaps. Also, even if we use a separate instance, the 2 instances would be sharing Alice's wallet and the ability to lock it while transactions are being created/mined.

cmd/swapd/main.go
dimalinux: I'll change it to `relayer`

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// The size of request.Secret was vetted when it was deserialized
+	secret := (*[32]byte)(request.Secret)
+
+	callData, err := getClaimRelayerTxCalldata(request.RelayerFeeWei, request.Swap, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	transSender, err := rrelayer.NewRelayer(&rrelayer.Config{
```
noot, line 49:
```suggestion
	txRelayer, err := rrelayer.NewRelayer(&rrelayer.Config{
```
perhaps?

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	forwarder, domainSeparator, err := getForwarderAndDomainSeparator(ctx, ec.Raw(), forwarderAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	nonce, err := forwarder.GetNonce(&bind.CallOpts{Context: ctx}, request.Swap.Claimer)
```
noot, line 36: it would be nice to validate the nonce before we send the tx; ie check that the nonce signed was the same as here, otherwise the tx has already been invalidated. the only drawback if we don't is the chance the tx will revert

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Lock the wallet's nonce until we get a receipt
+	ec.Lock()
+	defer ec.Unlock()
+
+	resp, err := transSender.SubmitTransaction(&rcommon.SubmitTransactionRequest{
+		From:            request.Swap.Claimer,
+		To:              request.SFContractAddress,
+		Value:           big.NewInt(0),
+		Gas:             big.NewInt(200000), // TODO: fetch from ethclient?
```
noot, line 71: turn this into a global? related to my previous comment about setting it to 52000

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, err
+	}
+
+	// Lock the wallet's nonce until we get a receipt
+	ec.Lock()
+	defer ec.Unlock()
+
+	resp, err := transSender.SubmitTransaction(&rcommon.SubmitTransactionRequest{
+		From:            request.Swap.Claimer,
+		To:              request.SFContractAddress,
```
noot, line 69: I don't see us validating the contract anywhere; we definitely need to check the bytecode

relayer/validate.go
```diff
@@ -0,0 +1,127 @@
...
+	ctx context.Context,
+	req *message.RelayClaimRequest,
+	ec *ethclient.Client,
+	expectedForwarderAddress ethcommon.Address,
+	minFee *big.Int,
+) error {
+	// Validate that the deployed SwapFactory contract has the same bytecode as
+	// the one we use. There is a good chance that it has the same exact address
+	// as ours, but we check for binary compatibility regardless.
+	requestedForwarderAddr, err := contracts.CheckSwapFactoryContractCode(ctx, ec, req.SFContractAddress)
```
noot, line 51: if it's the same address as our own, we don't need to validate this

relayer/validate.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	// The forwarder used must have the same exact address as ours, so we don't
+	// need to check the forwarder contract bytecode.
+	if requestedForwarderAddr != expectedForwarderAddress {
+		return fmt.Errorf("claim request had expected forwarder address: got %s, expected %s",
```
noot, line 59:
```suggestion
		return fmt.Errorf("claim request had unexpected forwarder address: got %s, expected %s",
```

noot: I would argue that using forwarder that aren't the same address as ours is fine, as long as the bytecode matches

relayer/forwarder.go
```diff
@@ -0,0 +1,123 @@
...
+	relayerFeeWei *big.Int,
+	swapFactoryAddress ethcommon.Address,
+	forwarderAddress ethcommon.Address,
+	swap *contracts.SwapFactorySwap,
+	secret *[32]byte,
+) ([]byte, error) {
+
+	if swap.Claimer != ethcrypto.PubkeyToAddress(claimerEthKey.PublicKey) {
+		return nil, fmt.Errorf("signing key does not match claimer %s", swap.Claimer)
+	}
```
noot, line 32: this is interesting; I think there's an attack vector if someone else signs the `claimRelayer` function for someone else;
ie.
1. I want to steal someone's swap funds so I become an (evil) relayer and wait for someone to send me a tx to relay
2. The tx they send me has their swap secret.  I sign the `claimRelayer` transaction with `fee == swap.value` and the swap secret
3. I submit this to the contract via the forwarder contract
4. I claim all the swap funds

This can easily be mitigated by adding a check `_msgSender() == swap.claimer` inside `claimRelayer()`. There's no check for who actually signed the message inside `claimRelayer()` currently, which opens up this attack vector.

Normal non-relayer `claim()` doesn't have this issue unless someone gives out their swap secret, which they obviously shouldn't.

I will add this in :)

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,90 @@
...
+		SuffixData:      nil,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = block.WaitForReceipt(ctx, ec.Raw(), resp.TxHash)
+	if err != nil {
+		return nil, err
+	}
```
noot, line 87: don't think we need this; the recipient will check for the receipt, so if the tx fails, it should be handled on the recipient side anyways

scripts/run-integration-tests.sh
```diff
@@ -133,7 +104,8 @@ start-daemons() {
 		--libp2p-port=9955 \
 		--rpc-port 5002 \
 		"--bootnodes=${ALICE_MULTIADDR}" \
-		"--contract-address=${SWAP_FACTORY_ADDR}"
+		"--contract-address=${SWAP_FACTORY_ADDR}" \
+		"--claim-relayer"
```
noot, line 108: wait so bob always uses a relayer now? don't we want to still test non-relayer txs?

noot (commented):
overall, I don't have major comments, in general this seems to be an improvement maintenance-wise. thanks for taking this on!

net/host.go
dimalinux: The comment was updated too.

net/message/message.go
dimalinux: Done

relayer/validate.go
dimalinux: That seems reasonable. Requiring the exact same forwarder address and always verifying the swap factory contract, even if it has the same address as our own, was what the existing code was doing.

relayer/forwarder.go
dimalinux: Fix merged into master by https://github.com/AthanorLabs/atomic-swap/pull/348.

scripts/run-integration-tests.sh
dimalinux: That flag `--claim-relayer` here is on Charlie's `swapd` instance.  It is not Bob or Alice's `swapd`. So Charlie is the instance providing relaying services in the integration tests.

Bob will only use the relayer in these cases, which I don't think have changed (yet) in this PR:
* No ETH to pay for the claim
*  He uses `--relayer-fee` from `swapcli` or, equivalently, passes a non-nil value in the `RelayerFee` field of `MakeOfferRequest`.

What I'd like to change based on other PR feedback:
* Remove the `RelayerFee` entirely from the `MakeOffer` request and replace it, at least temporarily, with a boolean `--force-relayer` (or `--use-relayer`) flag. This will allow our integration tests to work without a major overhaul. It will be nice even for manually testing, as you won't have to clear out your ETH balance to exercise the relayer.

My vision is that Alice will be the relay over 90% of the time. If we come up with ideas to reliably make Alice the relayer 100% of the time when Bob needs it, it will probably be solid gold for reducing code/protocol/end-user complexity. Bob's life is easier when he only needs to negotiate with one party instead of two.  :)

cmd/swapd/main.go
dimalinux: Yeah, I didn't add a way for the relayer to customize their fee in this PR. It would be follow-up work if we decided that we need it. I'll use `Sprintf` when using your suggestion above so any changes to the relayer fee can be in one place.

cmd/swapcli/main.go
dimalinux: Yeah, if we wanted to make the relayer fee configurable, we'd need to do it in conjunction with https://github.com/AthanorLabs/atomic-swap/issues/324 at the same time. Otherwise, it is an anti-feature in the sense that would increase failed swaps.

cmd/swapcli/main.go
dimalinux: Yes, the relayer is used if this flag is explicitly set or if the maker has no eth.  See 
https://github.com/AthanorLabs/atomic-swap/pull/342#discussion_r1133194128 for more details.

scripts/run-integration-tests.sh
noot: I don't think we can ever guarantee someone will be able to find a relayer unfortunately, people might choose not to relay for whatever reason. 

I think a `--use-relayer` flag instead of `--relayer-fee` is fine

relayer/send_relayed_transaction.go
dimalinux: We need to wait for the receipt while we still have the wallet lock to keep the nonce synced.  If we don't do it, and we are asked to perform another transaction before the outstanding transaction is mined, we'll reuse the nonce and one of the two transactions will fail. This was one of several reasons we turned off instamine in ganache (to catch code not waiting for the receipt).

relayer/validate.go
dimalinux: The code is updated and no longer requires that the forwarder address be exactly the same as ours. `CheckSwapFactoryContractCode` calls `CheckForwarderContractCode`, so the forwarder doesn't need to have its byte code checked separate from the swap factory contract.

relayer/forwarder.go
dimalinux: Constant in place

cmd/swapd/main.go
dimalinux: Renamed

net/host.go
dimalinux: You can only inherit methods from an anonymous struct field. Access to nested fields still require the type name prefix, so composing wasn't a good option. I just cloned the field's in `p2pnet`'s `Config` into our net's `Config`. More code to maintain another `Config`, but if our `net` is wrapping `p2pnet`'s `Host`, it should be wrapping the `Config` too in order to avoid transitive import requirements.

cmd/swapd/main.go
```diff
@@ -203,6 +203,14 @@ var (
 				Name:  flagUseExternalSigner,
 				Usage: "Use external signer, for usage with the swap UI",
 			},
+			&cli.BoolFlag{
+				Name: flagClaimRelayer,
+				Usage: fmt.Sprintf(
+					"Relay transactions for other XMR makers and earn %s ETH per relayed transaction",
```
noot, line 209:
```suggestion
					"Relay transactions for other XMR makers and earn %s ETH (minus gas fees) per relayed transaction",
```

net/host.go
```diff
@@ -79,25 +107,43 @@ func (h *Host) P2pHost() P2pHost {
...
+}
+
+// SetHandlers sets the maker and taker instances used by the host.
+func (h *Host) SetHandlers(makerHandler MakerHandler, takerHandler TakerHandler) {
+	h.makerHandler = makerHandler
+	h.takerHandler = takerHandler
+	h.h.SetAdvertisedNamespacesFunc(h.advertisedNamespaces)
+
+	h.h.SetStreamHandler(queryProtocolID, h.handleQueryStream)
+	h.h.SetStreamHandler(relayProtocolID, h.handleRelayStream)
```
noot, line 131:
```suggestion
if h.isRelayer {
	h.h.SetStreamHandler(relayProtocolID, h.handleRelayStream)
}
```

we should only handle incoming relay streams if we're a relayer

net/message/relay_message.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/athanorlabs/atomic-swap/common/vjson"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+)
+
+// RelayClaimRequest implements common.Message for our p2p relay claim requests
+type RelayClaimRequest struct {
+	SFContractAddress ethcommon.Address          `json:"sfContractAddress" validate:"required"`
+	Swap              *contracts.SwapFactorySwap `json:"swap" validate:"required"`
+	Secret            []byte                     `json:"secret" validate:"required,len=32"`
```
noot, line 16: what's the difference between this and having the type be `[32]byte`?

net/message/relay_message.go
```diff
@@ -0,0 +1,65 @@
...
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+
+	"github.com/athanorlabs/atomic-swap/common/vjson"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+)
+
+// RelayClaimRequest implements common.Message for our p2p relay claim requests
+type RelayClaimRequest struct {
+	SFContractAddress ethcommon.Address          `json:"sfContractAddress" validate:"required"`
```
noot, line 14:
```suggestion
	SwapContractAddress ethcommon.Address          `json:"swapContractAddress" validate:"required"`
```

potentially? in case of contract rename

net/relay.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := p2pnet.WriteStreamMessage(stream, request, relayerID); err != nil {
+		log.Warnf("failed to send RelayClaimRequest to peer: err=%s", err)
+		return nil, err
+	}
+
+	return receiveRelayClaimResponse(stream)
+}
+
+func receiveRelayClaimResponse(stream libp2pnetwork.Stream) (*RelayClaimResponse, error) {
+	msg, err := readStreamMessage(stream, maxMessageSize)
```
noot, line 89:
```suggestion
	msg, err := readStreamMessage(stream, maxRelayMessageSize)
```

net/relay.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading relay SubmitTransactionResponse: %w", err)
```
noot, line 91:
```suggestion
		return nil, fmt.Errorf("failed to read RelayClaimResponse: %w", err)
```

net/types.go
```diff
@@ -12,17 +12,25 @@ type SwapState = common.SwapStateNet //nolint:revive
...
+	SendKeysMessage    = message.SendKeysMessage
+	RelayClaimRequest  = message.RelayClaimRequest
+	RelayClaimResponse = message.RelayClaimResponse
 )
 
-// Handler handles swap initiation messages.
-// It is implemented by *xmrmaker.xmrmaker
-type Handler interface {
+// MakerHandler handles swap initiation messages and offer queries. It is
+// implemented by *xmrmaker.xmrmaker.
```
noot, line 24:
```suggestion
// implemented by *xmrmaker.Instance.
```

net/types.go
noot: I think it would make more sense to have a `relayer.Instance` or such that also contains the reference to the eth wallet, and either the database or the swap manager to tell if the relay request is from a current swap or not. relaying is logically separate from the taker, and thus shouldn't live in the same package imo

noot: also, the ethereum wallet (extendedethclient) is currently in the backend, so it's already used by both maker and taker anyways.

protocol/xmrmaker/offers/offer_manager.go
```diff
@@ -100,10 +98,12 @@ func (m *Manager) AddOffer(
 		return nil, err
 	}
 
+	// TODO: Is it a problem that we don't save the relayerFee to the database?
+	//       Should we place the relayer fee directly in the offer?
```
noot, line 102:
```suggestion
	// TODO: Is it a problem that we don't save `useRelayer` to the database?
	//       Should we place `useRelayer` directly in the offer?
```

I'm wondering if it makes more sense to have swapd take a `--use-relayer` flag rather than each individual offer. what do you think? either way doesn't need to be addressed in this PR

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,84 @@
+package relayer
```
noot, line 1: I'd name this file `submit_transaction.go` or something like that - the transaction hasn't been relayed yet, so `send_relayed_transaction` is a bit weird lol

relayer/validate.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// validateClaimValues validates the non-signature aspects of the claim request:
+//  1. the claim request swap factory contract is byte compatible with ours
+//  2. the forwarder in the claim request swap factory contract has an identical
+//     address with our forwarder
+//  3. the relayer fee is equal to or greater than the passed minFee
+//  4. the swap is for ETH and not an ERC20 token
+//  5. the swap value is strictly greater than the relayer fee
+//  6. TODO: Validate that the swap exists and is in a claimable state?
```
noot, line 39: this is a good idea, I'd add it in

relayer/validate.go
```diff
@@ -0,0 +1,127 @@
...
+	err := validateClaimValues(ctx, request, ec, ourSFContractAddr)
+	if err != nil {
+		return err
+	}
+
+	return validateClaimSignature(ctx, ec, request)
+}
+
+// validateClaimValues validates the non-signature aspects of the claim request:
+//  1. the claim request swap factory contract is byte compatible with ours
```
noot, line 33:
```suggestion
//  1. the claim request swap factory contract bytecode matches ours
```

relayer/validate.go
```diff
@@ -0,0 +1,127 @@
...
+//  2. the forwarder in the claim request swap factory contract has an identical
```
noot, line 34: it doesn't check this anymore - just checks the bytecode right?

relayer/validate.go
```diff
@@ -0,0 +1,127 @@
...
+//     address with our forwarder
+//  3. the relayer fee is equal to or greater than the passed minFee
```
noot, line 36: this isn't checked anymore

relayer/send_relayed_transaction.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// The size of request.Secret was vetted when it was deserialized
+	secret := (*[32]byte)(req.Secret)
+
+	// Lock the wallet's nonce until we get a receipt
+	ec.Lock()
+	defer ec.Unlock()
```
noot, line 54: move this to above the `ec.TxOpts(ctx)` line?

noot (commented):
looks mostly good, smaller comments mainly! about the xmrtaker vs relayer thing - I'd rather the code be in the relayer package, but feel free to do it in a follow up if you're planning on working on the "alice should be the priority relayer for bob" issue after this.

protocol/xmrmaker/offers/offer_manager.go
dimalinux: I'll remove this TODO. With the latest version of the PR, the only reason to use the `--use-relayer` flag is for testing. It will force the use of the relayer, even when Bob has sufficient ETH to pay for the claim. But that use case doesn't make financial sense, nor does it provide additional privacy. I think it is fine if `swapd` forgets that the flag was given when the original offer was made on restart. I'll also tag the flag as hidden so end users are not tempted to use it and pay additional fees by mistake.

 Using it doesn't make financial sense or provide additional privacy if you have enough ETH in the account being claimed into, and it is used automatically if Bob doesn't have enough ETH.

net/message/relay_message.go
dimalinux: With `[32]byte`, go will use the annoyingly verbose `{1, 128, 240, 10, ...}` JSON encoding of arrays. If we created a custom type for `[32]byte`, I could marshal it as base64 or base16, but since we didn't have a custom type, using a byte slice with size validation seemed like an easy solution to get base64 encoding.

tests/integration_test.go
```diff
@@ -141,10 +142,10 @@ func (s *IntegrationTestSuite) TestXMRTaker_Query() {
 func (s *IntegrationTestSuite) testXMRTakerQuery(asset types.EthAsset) {
 	ctx := context.Background()
 	bc := rpcclient.NewClient(ctx, defaultXMRMakerSwapdEndpoint)
-	offerResp, err := bc.MakeOffer(xmrmakerProvideAmount, xmrmakerProvideAmount, exchangeRate, asset, "", nil)
+	offerResp, err := bc.MakeOffer(xmrmakerProvideAmount, xmrmakerProvideAmount, exchangeRate, asset, false)
 	require.NoError(s.T(), err)
 
-	common.SleepWithContext(ctx, time.Second) // Give offer advertisement time to propagate
+	_ = common.SleepWithContext(ctx, time.Second) // Give offer advertisement time to propagate
```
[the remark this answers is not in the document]
dimalinux: Done

relayer/send_relayed_transaction.go
dimalinux: I moved the `ec.Lock()` call to after `createForwarderRequest(...)` which didn't need the lock's protection. That may have been what you were asking for. I left `ec.TxOpts(ctx)` inside the lock, because it's a fast/trivial operation and leaving it inside the lock unburdens the reader from needing to know that it leaves the `Nonce` field of the returned object as `nil`. (Otherwise, it would need to be inside the lock.)

relayer/validate.go
dimalinux: I removed (3). The fee is no longer part of the claim request, and if we change the fee, we'll need to update `relayProtocolID`. If, for some reason, the 2 sides are not in agreement on the fee, the relayer won't send the transaction, as the signature will not validate.

relayer/validate.go
dimalinux: Correct. I should have updated that. I combined (1) and (2) into just:
```text
1. the claim request's swap factory and forwarder contract bytecode matches ours
```

relayer/validate.go
dimalinux: Done. I also combined (1) and (2) into the same bullet. (See below.)

relayer/validate.go
dimalinux: Yeah, I'd like to see what happens now when we send a request, if the swap is not in a claimable state. In theory, the ethereum client simulates the transaction using the data (e.g. block timestamp) from the last block received. If the transaction reverts during simulation, it won't be sent to the mempool and no gas is lost. If this is the case, then the additional validation is probably a low priority.

We have seen cases where the transaction failed after it went out to the mempool. It would happen because the mined block's timestamp was further into the future than the last block triggering a revert. We had to write extra code to extract out the revert message when that happens.

net/types.go
dimalinux: Let's push this work into the follow-up PR that will ask Alice to relay Bob's claim before reaching out to the open market of relayers. We'll be modifying this code anyways and it will be easier to view in a smaller PR.

protocol/backend/backend.go
dimalinux: The updated name is `NetSender`. It's the interface for someone sending messages via the `net` package's p2p Host, so that name seemed a little more helpful. Feel free to suggest better.

cmd/swapd/main.go
dimalinux: The help message now looks like this:
```text
--relayer    Relay claims for XMR makers and earn 0.009 ETH (minus gas fees) per transaction (default: false)
```

cmd/swapd/main.go
```diff
@@ -76,6 +75,7 @@ const (
 	flagGasPrice             = "gas-price"
 	flagGasLimit             = "gas-limit"
 	flagUseExternalSigner    = "external-signer"
+	flagClaimRelayer         = "relayer"
```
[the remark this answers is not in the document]
dimalinux: Done.

net/host.go
dimalinux: We might need to remove it in the follow up PR, but I put it in as suggested.

net/message/relay_message.go
dimalinux: I changed it to `SwapFactoryAddress` to make it consistent with our naming elsewhere. That will help with the contract rename, as it will get changed in any global string replacement of `swapFactory` and `SwapFactory` into whatever the new name becomes.

net/relay.go
dimalinux: Good catch. Implemented verbatim, so I'll resolve this one.

net/relay.go
dimalinux: Done (verbatim fix).

net/types.go
dimalinux: Done

relayer/send_relayed_transaction.go
dimalinux: The new name is `submit_transaction.go`, so I'll resolve this one.

net/host.go
```diff
@@ -79,25 +107,45 @@ func (h *Host) P2pHost() P2pHost {
...
+func (h *Host) SetHandlers(makerHandler MakerHandler, takerHandler TakerHandler) {
+	h.makerHandler = makerHandler
+	h.takerHandler = takerHandler
+	h.h.SetAdvertisedNamespacesFunc(h.advertisedNamespaces)
+
+	h.h.SetStreamHandler(queryProtocolID, h.handleQueryStream)
+	if h.isRelayer {
+		h.h.SetStreamHandler(relayProtocolID, h.handleRelayStream)
+	}
+	h.h.SetStreamHandler(swapID, h.handleProtocolStream)
```
noot, line 134: curious why the stream handlers were moved here instead of left in `Start()`? it doesn't make a huge difference, but this function was meant to be called specifically to set the maker/taker handlers. if you leave it here, update the comment

noot (approved):
looks good! excited to test this out

net/host.go
dimalinux: As a general principle, methods named `Start` should avoid initialization that could have been done earlier during a constructor or other setup methods. This minimizes the time delta between invoking start and starting.  While there are no errors to handle in this particular case, pushing initialization code earlier, to when a program is likely to be in a single thread/process context, also leads to cleaner error handling. The method is called `SetHandlers`, and these are handlers, so it seemed like a good place to move the code. I'll update the comment.

PR #348, opened 2023-03-11
fix contract `claimRelayer` vulnerability

noot:
see https://github.com/AthanorLabs/atomic-swap/pull/342#discussion_r1133154000


dimalinux (approved):
I made one commit to the branch to update the goerli contract address.

noot:
@dimalinux good point, we should probably make a checklist of things to update if the contract is updated

PR #350, opened 2023-03-18
update `swap_getPast` to match `swap_getOngoing`; update docs

### LINKED ISSUE #349: update rpc get-past-swaps API to match get-ongoing-swaps
the APIs for past and ongoing swaps are different; this issue is to make them consistent (ie. remove the get-past-swap-ids method and change it to get-past-swaps)

noot:
- update `swap_getPast` to match `swap_getOngoing`'s behaviour; returns all past swaps if no ID is provided
- remove `swap_getPastIDs`
- update `swapcli get-past-swaps` -> `swapcli past` and `swapcli get-ongoing-swaps` -> `swapcli ongoing` (I think this is clearer and less to type out)
- update docs

closes #349

### DIFF
--- a/rpc/swap.go
+++ b/rpc/swap.go
@@ -46,54 +46,9 @@ func NewSwapService(
 	}
 }
 
-// PastSwapInfo contains a swap ID and its start and end time.
-type PastSwapInfo struct {
-	ID        types.Hash `json:"id" validate:"required"`
-	StartTime time.Time  `json:"startTime" validate:"required"`
-	EndTime   *time.Time `json:"endTime"`
-}
-
-// GetPastIDsResponse ...
-type GetPastIDsResponse struct {
-	Swaps []*PastSwapInfo `json:"swaps" validate:"dive,required"`
-}
-
-// GetPastIDs returns all past swap IDs and their start and end times.
-// It sorts them in order from oldest to newest.
-func (s *SwapService) GetPastIDs(_ *http.Request, _ *interface{}, resp *GetPastIDsResponse) error {
-	ids, err := s.sm.GetPastIDs()
-	if err != nil {
-		return err
-	}
-
-	resp.Swaps = make([]*PastSwapInfo, len(ids))
-	for i, id := range ids {
-		info, err := s.sm.GetPastSwap(id)
-		if err != nil {
-			return fmt.Errorf("failed to get past swap %s: %w", id, err)
-		}
-
-		resp.Swaps[i] = &PastSwapInfo{
-			ID:        id,
-			StartTime: info.StartTime,
-			EndTime:   info.EndTime,
-		}
-	}
-
-	sort.Slice(resp.Swaps, func(i, j int) bool {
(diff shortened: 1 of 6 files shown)


rpc/swap.go
```diff
@@ -46,37 +46,68 @@ func NewSwapService(
...
+	ExpectedAmount *apd.Decimal        `json:"expectedAmount" validate:"required"`
+	ExchangeRate   *coins.ExchangeRate `json:"exchangeRate" validate:"required"`
+	Status         types.Status        `json:"status" validate:"required"`
+	StartTime      time.Time           `json:"startTime" validate:"required"`
+	EndTime        *time.Time          `json:"endTime"`
+}
+
+// GetPastRequest ...
+type GetPastRequest struct {
+	OfferID types.Hash `json:"offerID"`
```
dimalinux, line 63: You can put `validate:"required"` on this one to make the check for an all-zero hash automatic.

As a side note, on something I learned recently: It works here, because `Hash` is an array. The validator can check any field that resolves to a built-in type for the uninitialized value, but it can't check to see if a structure type has all of its fields uninitialized (which would potentially be a recursive operation). This is because the validator recurses into structures and looks for annotations on the nested fields. Because both `big.Int` and `apd.Decimal` are implemented as structures (and don't have annotations on their nested fields), we can't use the default validation annotations to see if they are zero. We can check that their pointer types are initialized/required, but we can't do a `dive,required` validation on their pointer types.

rpc/swap.go
```diff
@@ -46,37 +46,68 @@ func NewSwapService(
...
 }
 
-// GetPastIDsResponse ...
-type GetPastIDsResponse struct {
-	Swaps []*PastSwapInfo `json:"swaps" validate:"dive,required"`
+// GetPastResponse ...
+type GetPastResponse struct {
+	Swaps []*PastSwap `json:"swaps" validate:"required"`
```
dimalinux, line 68: This validation is saying that the slice must be non-empty. Can it be empty? I suspect that you wanted `dive,required` to validate that each entry in the slice is non-nil. If you wanted to validate that both the slice is non-empty and that each pointer in the slice is non-nil, you would use `required,dive,required`.

rpc/swap.go
noot: I want to allow empty hashes though, wouldn't adding required defeat the purpose there?

rpc/swap.go
noot: updated

rpc/swap.go
dimalinux: Given that it is an optional parameter, I'd use this declaration:
```
OfferID *types.Hash `json:"offerID,omitempty"`
```
You get several benefits:
* the field becomes self documenting that it is optional
* the JSON on the wire, which people do look at, becomes just `{}`
* checking for nil is arguably cleaner than checking for a zero hash

dimalinux (approved):
Looks good. I added one new comment that might be worth implementing.

PR #351, opened 2023-03-19
add estimated time for swap to complete for `swap_getOngoing`

### LINKED ISSUE #314: swapcli: add additional details to ongoing swaps
It would also be nice to add additional info other than status to ongoing swaps, for example:
- estimated time for swap to complete based on network block time
- how many more blocks we're waiting for until the next stage

Right now, it's a bit painful when waiting for a swap to complete, as there isn't a lot of detail (especially from the taker side) as to how long things will take.

noot:
- add estimated time for swap to complete based on network block times; this is in the success case. in the refund case (>t1), it's impossible to estimate a time unfortunately.

closes #314


rpc/swap.go
```diff
@@ -120,15 +120,17 @@ func (s *SwapService) GetPast(_ *http.Request, req *GetPastRequest, resp *GetPas
...
-	StartTime      time.Time           `json:"startTime" validate:"required"`
-	Timeout0       *time.Time          `json:"timeout0"`
-	Timeout1       *time.Time          `json:"timeout1"`
+	ID                        types.Hash          `json:"id" validate:"required"`
+	Provided                  coins.ProvidesCoin  `json:"provided" validate:"required"`
+	ProvidedAmount            *apd.Decimal        `json:"providedAmount" validate:"required"`
+	ExpectedAmount            *apd.Decimal        `json:"expectedAmount" validate:"required"`
+	ExchangeRate              *coins.ExchangeRate `json:"exchangeRate" validate:"required"`
+	Status                    types.Status        `json:"status" validate:"required"`
+	LastStatusUpdateTime      time.Time           `json:"lastStatusUpdateTime" validate:"required"`
```
dimalinux, line 129: This is my fault, and doesn't cause any real harm to leave it, but the `required` validation here doesn't do anything. The reason is because `time.Time` is a struct, so the validation library recurses into it and looks for validation tags on the internal fields (which, in this case, have no tags).

rpc/swap.go
```diff
@@ -120,15 +120,17 @@ func (s *SwapService) GetPast(_ *http.Request, req *GetPastRequest, resp *GetPas
...
+	StartTime                 time.Time           `json:"startTime" validate:"required"`
+	Timeout0                  *time.Time          `json:"timeout0"`
+	Timeout1                  *time.Time          `json:"timeout1"`
+	EstimatedTimeToCompletion time.Duration       `json:"estimatedTimeToCompletion" validate:"required"`
```
dimalinux, line 133: If you are looking to trim a few characters you could use `EstimatedTimeRemaining`.  (No need to make that change unless you like it, it only save 3 symbols.)

cmd/swapcli/main.go
```diff
@@ -667,10 +667,12 @@ func runGetOngoingSwap(ctx *cli.Context) error {
 		fmt.Printf("Receiving: %s %s\n", info.ExpectedAmount.Text('f'), receivedCoin)
 		fmt.Printf("Exchange Rate: %s ETH/XMR\n", info.ExchangeRate)
 		fmt.Printf("Status: %s\n", info.Status)
+		fmt.Printf("Time status was last updated: %s\n", info.LastStatusUpdateTime.Format(common.TimeFmtSecs))
 		if info.Timeout0 != nil && info.Timeout1 != nil {
 			fmt.Printf("First timeout: %s\n", info.Timeout0.Format(common.TimeFmtSecs))
 			fmt.Printf("Second timeout: %s\n", info.Timeout1.Format(common.TimeFmtSecs))
 		}
+		fmt.Printf("Estimated time to completion: %s\n", info.EstimatedTimeToCompletion)
```
dimalinux, line 675: The output seems a little overly exact:
```
Estimated time to completion: 20m45.316266949s
```
I didn't test this, but you might be able to use the duration `Round` function passing it 1 minute or one second in the place that you are creating it.

dimalinux (approved):
Do we have plans to add any kind of time estimates to the subscribe channel? (Just a question, the answer might be that we don't need it.)

I left one comment about rounding the estimates that would be worth implementing before merging.

rpc/swap.go
noot: ah okay, so should I remove it for all `time.Time` types?

cmd/swapcli/main.go
noot: good point lol, I rounded it to the nearest second.

rpc/swap.go
dimalinux: If it is a pointer to a `time.Time` type, it will make sure that the pointer is not `nil`, so definitely don't remove those. We just can't use it to validate that the time is not `0001-01-01 00:00:00 +0000 UTC`.  If it is a non-pointer type, or we have a `dive,required` on a pointer to `time.Time`, we should remove the `required` tag, because it misleads the reader into thinking that we prevent the all-zero time value.

PR #354, opened 2023-03-21
swapd daemon setup/start decoupled from CLI parsing

dimalinux:
The primary driver for decoupling swapd from the CLI parsing code is to make it easy to unit tests running multiple swapd instances simultaneously. Hopefully the end result is slightly easier to maintain as well. If we ever have the desire to switch to a different CLI parser like Cobra, it will be an easier task.


cmd/swapd/contract.go
```diff
@@ -31,43 +36,29 @@ func getOrDeploySwapFactory(
...
 	} else {
 		// otherwise, load the contract from the given address
 		// and check that its bytecode is valid (ie. matches the
 		// bytecode of this repo's swap contract)
-		sf, err = getSwapFactory(ec, address)
+		_, err = contracts.CheckSwapFactoryContractCode(ctx, ec.Raw(), address)
 		if err != nil {
-			return nil, ethcommon.Address{}, err
-		}
-		log.Infof("loaded SwapFactory.sol from address %s", address)
```
dimalinux, line 58: `sf` wasn't being used by any of the callers, so I removed it from the returned values. The print probably doesn't matter, since the user passed the address on the command line.

common/config.go
```diff
@@ -32,76 +30,86 @@ type MoneroNode struct {
...
 
 // MainnetConfig is the mainnet ethereum and monero configuration
-var MainnetConfig = Config{
-	DataDir: path.Join(baseDir, "mainnet"),
-	MoneroNodes: []*MoneroNode{
-		{
-			Host: "node.sethforprivacy.com",
-			Port: 18089,
+func MainnetConfig() *Config {
+	return &Config{
```
dimalinux, line 43: There are a bunch of places in the code where I removed global and package level variables that we modify, so multiple swapd instances (simultaneous or successive) do not modify each other's state.

common/config.go
```diff
@@ -32,76 +30,86 @@ type MoneroNode struct {
 
 // Config contains constants that are defaults for various environments
 type Config struct {
+	Env                      Environment
```
dimalinux, line 33: I added this, so code can just pass the config around without passing both the config and the env. We no longer modify the configs globally, so you have to pass the config around now.

common/utils.go
```diff
@@ -73,3 +74,17 @@ func SleepWithContext(ctx context.Context, d time.Duration) error {
 		return nil
 	}
 }
+
+// GetFreeTCPPort returns an OS allocated and immediately freed port. There is
+// nothing preventing something else on the system from using the port before
+// the caller has a chance, but OS allocated ports are randomised to make the
+// risk negligible.
+func GetFreeTCPPort() (uint, error) {
```
dimalinux, line 82: Previously this was a private function in the monero package (for picking a free monero-wallet-rpc port). Moved it here and made it public so others can use it.

protocol/xmrtaker/swap_state_test.go
```diff
@@ -129,7 +124,7 @@ func newTestSwapStateAndNet(t *testing.T) (*swapState, *mockNet) {
 	providedAmt := coins.EtherToWei(coins.StrToDecimal("1"))
 	expectedAmt := coins.MoneroToPiconero(coins.StrToDecimal("1"))
 	exchangeRate := coins.ToExchangeRate(coins.StrToDecimal("1.0")) // 100%
-	swapState, err := newSwapStateFromStart(b, types.Hash{}, false,
+	swapState, err := newSwapStateFromStart(b, types.Hash{}, true,
```
dimalinux, line 127: The swapped boolean here is `noTransferBack`. Some of the XMR taker tests depend on this behavior.

daemon/swap_daemon.go
```diff
@@ -0,0 +1,165 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"path"
+
+	"github.com/ChainSafe/chaindb"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/hashicorp/go-multierror"
```
dimalinux, line 14: Apparently, multierror is so popular that Go 1.20 will have a version of it integrated into the language.

cmd/swapd/main.go
```diff
@@ -81,14 +73,14 @@ const (
 	flagDevXMRMaker      = "dev-xmrmaker"
 	flagDeploy           = "deploy"
 	flagForwarderAddress = "forwarder-address"
-	flagTransferBack     = "transfer-back"
+	flagNoTransferBack   = "no-transfer-back"
```
dimalinux, line 76: Ideally, boolean flags and boolean fields in Go structs use false for their default value. When we first changed the default behavior on this, it seemed like too much work to update all the code. In this PR, we were adding new code (`SwapdConfig`), in which I used `NoTransferBack` (so the uninitialized value was the was common behavior) and I decided to make everything consistent.

cmd/swapd/main.go
```diff
@@ -148,7 +140,7 @@ var (
 			},
 			&cli.StringFlag{
 				Name:  flagEthereumEndpoint,
-				Usage: "Ethereum client endpoint",
+				Usage: "Ethereum websockets client endpoint",
```
dimalinux, line 143: With this PR we now fail quickly (instead of slow and painfully), if the user passes an http/https endpoint.

net/host.go
```diff
@@ -81,25 +80,29 @@ type Config struct {
...
+	var err error
+	h.h, err = p2pnet.NewHost(&p2pnet.Config{
+		Ctx:                      cfg.Ctx,
+		DataDir:                  cfg.DataDir,
+		Port:                     cfg.Port,
+		KeyFile:                  cfg.KeyFile,
+		Bootnodes:                cfg.Bootnodes,
+		ProtocolID:               cfg.ProtocolID,
+		ListenIP:                 cfg.ListenIP,
+		AdvertisedNamespacesFunc: h.advertisedNamespaces,
```
dimalinux, line 99: The self reference on this line is why I had to reverse the initialization order.

protocol/xmrmaker/instance.go
```diff
@@ -35,8 +35,6 @@ type Instance struct {
 
 	net Host
 
-	walletFile, walletPassword string
```
dimalinux, line 38: These fields were not longer used even before this PR.

cmd/swapd/main.go
```diff
@@ -1,35 +1,27 @@
-// Package main provides the entrypoint of swapd, a daemon that manages atomic swaps
-// between monero and ethereum assets.
+// Package main provides the entrypoint of swapd executable, a daemon that
```
noot, line 1:
```suggestion
// Package main provides the entrypoint of the swapd executable, a daemon that
```

cmd/swapd/main.go
noot: why was it failing before?

cmd/swapd/main.go
```diff
@@ -301,375 +276,153 @@ func runDaemon(c *cli.Context) error {
...
+// validateOrDeployContracts validates or deploys the swap factory. The SwapFactoryAddress field
+// of envConf should be all zeros if deploying and its value will be replaced by the new deployed
+// contract.
+func validateOrDeployContracts(c *cli.Context, envConf *common.Config, ec extethclient.EthClient) error {
+	deploy := c.Bool(flagDeploy)
+	if deploy {
+		if envConf.SwapFactoryAddress != (ethcommon.Address{}) {
+			panic("contract address should have been zeroed when envConf was initialized")
+		}
+	}
```
noot, line 388:
```suggestion
	if deploy && envConf.SwapFactoryAddress != (ethcommon.Address{}) {
		panic("contract address should have been zeroed when envConf was initialized")
	}
```

cmd/swapd/main.go
```diff
@@ -301,375 +276,153 @@ func runDaemon(c *cli.Context) error {
...
-		ok := ethcommon.IsHexAddress(forwarderAddressStr)
-		if !ok {
-			return nil, errors.New("forwarder-address is invalid")
+		if !ethcommon.IsHexAddress(forwarderAddressStr) {
+			return errors.New("forwarder-address is invalid")
 		}
 
 		forwarderAddress = ethcommon.HexToAddress(forwarderAddressStr)
 	} else if !deploy && forwarderAddressStr != "" {
 		log.Warnf("forwarder-address is unused")
```
noot, line 401: should this return an error?

daemon/swap_daemon.go
```diff
@@ -0,0 +1,165 @@
...
+	})
+
+	log.Infof("starting swapd with data-dir %s", conf.EnvConf.DataDir)
+	err = rpcServer.Start() // blocks until server is shutdown or context is cancelled
+	if err != nil && !errors.Is(err, http.ErrServerClosed) {
+		return err
+	}
+
+	// err can get set in defer blocks, so don't return nil below
+	return err
```
noot, line 164: I don't think this works, defer statements get evaluated after the return. see: https://go.dev/play/p/SbbM1nQj9ks

protocol/xmrmaker/claim.go
```diff
@@ -125,9 +127,10 @@ func (s *swapState) discoverRelayersAndClaim() (ethcommon.Hash, error) {
 	}
 
 	for _, relayer := range relayers {
+		log.Debugf("submitting claim to relayer peer ID %s", relayer)
```
noot, line 130:
```suggestion
		log.Debugf("submitting claim to relayer with peer ID %s", relayer)
```

protocol/xmrmaker/claim.go
```diff
@@ -201,11 +204,24 @@ func waitForClaimReceipt(
...
+			txHash, receipt.BlockNumber)
 	}
 
-	return checkClaimedLog(receipt.Logs[0], contractAddr, contractSwapID, secret)
+	if err = checkClaimedLog(receipt.Logs[0], contractAddr, contractSwapID, secret); err != nil {
+		return fmt.Errorf("relayer's claim had logs error (tx=%s block=%d): %w",
+			txHash, receipt.BlockNumber, err)
+	}
+
+	log.Debugf("relayer's claim transaction validated, gas used: %d", receipt.GasUsed)
```
noot, line 223:
```suggestion
	log.Debugf("relayer's claim transaction included, gas used: %d", receipt.GasUsed)
```

noot: didn't look like the tx hash was being logged anywhere on success

tests/ganache.go
```diff
@@ -1,3 +1,4 @@
+// Package tests provides
```
noot, line 1: complete comment?

noot (commented):
this is looking really good! only minor comments, this will help a lot with testing :)

tests/ganache.go
dimalinux: Ops. I should have deleted that. I was experimenting with moving this file to a different package, but decided against it. The package comments are already in integration_test.go.

daemon/swap_daemon.go
dimalinux: I guess it makes sense, but I didn't realize that there is a subtle difference between named and unnamed return values in this regard.  From [the Go standard](https://go.dev/ref/spec):
> if the deferred function is a [function literal](https://go.dev/ref/spec#Function_literals) and the surrounding function has [named result parameters](https://go.dev/ref/spec#Function_types) that are in scope within the literal, the deferred function may access and modify the result parameters before they are returned.

I'll fix the code. Here's an alternative version of that example: https://go.dev/play/p/ajSRoozvuhK?v=goprev

cmd/swapd/main.go
dimalinux: I'll make it return an error. I actually thought it should return an error, but didn't know if it was written this way intentionally.

cmd/swapd/main.go
dimalinux: Resolving. Implemented as suggested.

daemon/swap_daemon.go
dimalinux: Done

protocol/xmrmaker/claim.go
dimalinux: Done

cmd/swapd/main.go
noot: @dimalinux I've been running the node with http endpoints always and never had any issues?

cmd/swapd/main.go
dimalinux: It returns an error now and I added a unit test.  I'll resolve this one.

cmd/swapd/main.go
dimalinux: I reverted it back and created a ticket to remove the dependency on websocket URLs:
https://github.com/AthanorLabs/atomic-swap/issues/356

noot (approved):

PR #358, opened 2023-03-26
fix: remove ws eth endpoint requirement, update taker exit

### LINKED ISSUE #356: Allow non-websocket ETH endpoint URLs
Right now failures when using an http(s) are subtle and don't happen immediately:
```
xmrtaker/message_handler.go:276	Failure waiting for T1 timeout: err=notifications not supported
```
The problem is probably limited to `WaitForEthBlockAfterTimestamp`'s use of `ec.SubscribeNewHead` which we can replace with polling.

noot:
- closes #356 
- update `recoveryDB.DeleteSwap` to not do anything (moved the logic to an unused func) since it was being called when it shouldn't and it's better to just not delete to be safe
- add more logs for debugging

there's a few bug fixes in here, let me know if you want me to split them out.


rpc/swap.go
```diff
@@ -303,6 +303,10 @@ func (s *SwapService) Cancel(_ *http.Request, req *CancelRequest, resp *CancelRe
 		ss = s.xmrmaker.GetOngoingSwapState(req.OfferID)
 	}
 
+	if ss == nil {
+		return fmt.Errorf("failed to find swap state with ID %s", req.OfferID)
+	}
```
noot, line 308: this was an attempt to fix a panic in `cancel()` when it's called twice successively:
```
2023-03-26T00:40:48.371Z	INFO	xmrmaker	xmrmaker/event.go:241	EventExit
2023-03-26T00:40:48.371Z	DEBUG	xmrmaker	xmrmaker/swap_state.go:338	attempting to exit swap: nextExpectedEvent=EventETHLockedType
2023-03-26T00:40:48.371Z	DEBUG	rpc	rpc/ws.go:61	failed to read websockets message: websocket: close 1006 (abnormal closure): unexpected EOF
2023-03-26T00:40:48.372Z	INFO	xmrmaker	xmrmaker/swap_state.go:347	exit status Aborted
2023-03-26T00:40:48.373Z	DEBUG	xmrmaker	xmrmaker/swap_state.go:356	re-added offer 0xadc8490398ecf78797aebd055ca13f698652340faf8de05367f806bdb26e634b
2023-03-26T00:40:48.377Z	INFO	xmrmaker	xmrmaker/swap_state.go:383	**swap aborted: id=0xadc8490398ecf78797aebd055ca13f698652340faf8de05367f806bdb26e634b**
2023-03-26T00:40:48.377Z	DEBUG	net	net/host.go:179	closing stream: peer=12D3KooWDqCzbjexHEa8Rut7bzxHFpRMZyDRW1L6TGkL1KY24JH5 protocol=/atomic-swap/0.2/5/swap/0
2023-03-26T00:40:48.377Z	DEBUG	net	net/initiate.go:149	Failed to read message from peer, id=12D3KooWDq-55-86 protocol=/atomic-swap/0.2/5/swap/0: stream reset
2023-03-26T00:40:48.377Z	DEBUG	net	net/initiate.go:131	closing stream: peer=12D3KooWDqCzbjexHEa8Rut7bzxHFpRMZyDRW1L6TGkL1KY24JH5 protocol=/atomic-swap/0.2/5/swap/0
2023-03-26T00:40:48.377Z	DEBUG	net	net/initiate.go:134	exiting swap...
2023/03/26 00:42:34 http: panic serving 127.0.0.1:38580: runtime error: invalid memory address or nil pointer dereference
goroutine 3013 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1850 +0xbf
panic({0x122e320, 0x2114cd0})
	/usr/local/go/src/runtime/panic.go:890 +0x262
github.com/athanorlabs/atomic-swap/protocol/xmrmaker.(*swapState).Exit(0x0)
	/root/atomic-swap/protocol/xmrmaker/swap_state.go:332 +0x77
github.com/athanorlabs/atomic-swap/rpc.(*SwapService).Cancel(0xc0077cf6e0, 0xc0073abc68?, 0xc0071dac80, 0xc000417d38)
	/root/atomic-swap/rpc/swap.go:306 +0x262
reflect.Value.call({0xc0077d1a80?, 0xc0076b9c20?, 0xc000881590?}, {0x13f5267, 0x4}, {0xc000881740, 0x4, 0xc0008815c0?})
	/usr/local/go/src/reflect/value.go:584 +0x8c5
reflect.Value.Call({0xc0077d1a80?, 0xc0076b9c20?, 0xc0071dac80?}, {0xc000881740?, 0x203000?, 0x203000?})
	/usr/local/go/src/reflect/value.go:368 +0xbc
github.com/gorilla/rpc/v2.(*Server).ServeHTTP(0xc0077eb180, {0x1872750, 0xc0082bfce0}, 0xc0006a2500)
	/root/go/pkg/mod/github.com/gorilla/rpc@v1.2.0/v2/server.go:220 +0x7f7
github.com/gorilla/mux.(*Router).ServeHTTP(0xc00773ad80, {0x1872750, 0xc0082bfce0}, 0xc0006a2300)
	/root/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210 +0x1cf
github.com/gorilla/handlers.(*cors).ServeHTTP(0xc0077990e0, {0x1872750, 0xc0082bfce0}, 0xc0006a2300)
	/root/go/pkg/mod/github.com/gorilla/handlers@v1.5.1/cors.go:54 +0x370
net/http.serverHandler.ServeHTTP({0x186e910?}, {0x1872750, 0xc0082bfce0}, 0xc0006a2300)
	/usr/local/go/src/net/http/server.go:2947 +0x30c
net/http.(*conn).serve(0xc0072086e0, {0x1873be0, 0xc0077fe660})
	/usr/local/go/src/net/http/server.go:1991 +0x607
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:3102 +0x4db
```

I think a lock somewhere could also fix this.

protocol/xmrmaker/checks.go
```diff
@@ -94,7 +94,7 @@ func checkContractSwapID(msg *message.NotifyETHLocked) error {
 func (s *swapState) checkContract(txHash ethcommon.Hash) error {
 	tx, _, err := s.ETHClient().Raw().TransactionByHash(s.ctx, txHash)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get transaction by hash: %w", err)
```
dimalinux, line 97: Would it be correct to write `failed to get newSwap transaction by hash`? Knowing which transaction wasn't found would be helpful. The function comments should probably state which txHash is being passed as well (or it could be placed into the parameter name).

protocol/xmrmaker/instance.go
```diff
@@ -142,7 +142,7 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
 
 	ethSwapInfo, err := inst.backend.RecoveryDB().GetContractSwapInfo(s.ID)
 	if err != nil {
-		return fmt.Errorf("failed to get info for ongoing swap, id %s: %s", s.ID, err)
+		return fmt.Errorf("failed to get contract info for ongoing swap, id %s: %s", s.ID, err)
```
dimalinux, line 145: Should we change that second `%s` to `%w`? I'm not sure adding the word `contract` in isolation clarifies things. It makes it sound like it was an ethereum related error.  Maybe this?
`failed to load contract info for ongoing swap from db, swapID %s: %w`

protocol/xmrtaker/instance.go
```diff
@@ -116,7 +117,7 @@ func (inst *Instance) createOngoingSwap(s *swap.Info) error {
 
 	ethSwapInfo, err := inst.backend.RecoveryDB().GetContractSwapInfo(s.ID)
 	if err != nil {
-		return fmt.Errorf("failed to get info for ongoing swap, id %s: %s", s.ID, err)
+		return fmt.Errorf("failed to get contract info for ongoing swap, id %s: %s", s.ID, err)
```
dimalinux, line 120: Whatever change you make on the other one, match the change here.

protocol/xmrtaker/message_handler.go
```diff
@@ -120,6 +120,8 @@ func (s *swapState) handleSendKeysMessage(msg *message.SendKeysMessage) (common.
 		return nil, fmt.Errorf("failed to set xmrmaker keys: %w", err)
 	}
 
+	log.Debugf("set and stored XMR maker's keys, going to lock ETH")
```
dimalinux, line 123: "Set" is implied by "stored". This could be "XMR maker keys stored, ...".

protocol/xmrtaker/swap_state.go
```diff
@@ -390,16 +390,21 @@ func (s *swapState) exit() error {
 		txHash, err := s.tryRefund()
 		if err != nil {
 			if strings.Contains(err.Error(), revertSwapCompleted) {
-				return s.tryClaim()
+				// note: this should NOT ever error; it could if the ethclient
+				// or monero clients crash during the course of the claim,
+				// but that would be very bad.
+				err = s.tryClaim()
+				if err != nil {
+					panic(fmt.Sprintf("failed to claim even though swap was completed on-chain: %s", err))
```
dimalinux, line 398: panic may be too extreme here. It looks like this could happen if, for example, the endpoints are overloaded ... or a user went past their API limit on an endpoint. It wouldn't take a full crash of the client.

protocol/xmrtaker/swap_state.go
```diff
@@ -484,17 +487,34 @@ func (s *swapState) tryRefund() (ethcommon.Hash, error) {
 	// it won't handle those events while this function is executing.)
 	log.Infof("waiting until time %s to refund", s.t1)
 
-	event := <-s.eventCh
-	log.Debugf("got event %s while waiting for T1", event)
-	switch event.(type) {
-	case *EventShouldRefund:
+	waitCtx, waitCtxCancel := context.WithCancel(context.Background())
```
dimalinux, line 490: Should this be using `s.ctx` instead of `context.Background()`?

protocol/xmrtaker/swap_state.go
```diff
@@ -484,17 +487,34 @@ func (s *swapState) tryRefund() (ethcommon.Hash, error) {
...
+		case *EventETHClaimed:
+			// we should claim; returning this error
+			// causes the calling function to claim
+			return ethcommon.Hash{}, fmt.Errorf(revertSwapCompleted)
+		default:
+			panic(fmt.Sprintf("got unexpected event while waiting for Claimed/T1: %s", event))
+		}
+	case err = <-waitCh:
+		if err != nil {
+			panic(fmt.Sprintf("failed to wait for T1: %s", err))
```
dimalinux, line 514: Same thing here. This seems like something that can legitimately happen without any bugs in our code.

rpc/swap.go
dimalinux: What happens if two `Cancel(...)` requests come in at the same time? They will both succeed in getting the same `ss` variable and then go on to call `Exit` on that same object simultaneously. Is that code re-entryancy safe? If it is, just add some code comments to that effect. If the code is not re-entrant, we'll need locking.

dimalinux: I see one more problem.  `GetOngoingSwapState` in the `xmrtaker` isn't grabbing the lock where it should. In `xmrmaker` it is.

protocol/xmrtaker/swap_state.go
noot: yeah, I removed the panics. I guess my thought process was that if this case is hit, it's really bad because the swap is unable to complete. if this happens it's probably due to bad endpoint or some other reason though, and the user would need to restart anyways

rpc/swap.go
noot: `Exit()` is safe for concurrent calls, since it puts an exit event into the eventCh, and events are handled sequentially

however the `Refund` and `Cancel` methods could probably be tested more. we should consider removing `Refund` in favour of `Cancel`

dimalinux: Can you create that ticket? I could create it, but I'm not sure I can describe the finer nuances. If Cancel can already do everything that Refund does, we should definitely remove Refund and just adjust the documentation.

dimalinux (approved):

PR #361, opened 2023-03-28
xmrmaker: check if swap was completed already when restarting swap

noot:
as title says, I came across an issue when testing where the maker claimed ETH, but then shutdown/didn't get the result and mark the swap complete due to a bad endpoint. Then on restart, it would be waiting for a ready/refunded event, but neither of these would ever happen b/c the swap was already done.

this PR updates the xmrmaker to check if the swap was already completed and the ETH was claimed already (by them) on restart, and if so marks the swap successfully complete.


protocol/xmrmaker/errors.go
```diff
@@ -30,6 +30,7 @@ var (
 	errRelayingWithNonEthAsset       = errors.New("relayers with ERC20 token swaps are not currently supported")
 
 	// protocol initiation errors
+	errSwapDoesNotExist          = errors.New("swap with given ID in contract does not exist")
```
dimalinux, line 33: I'm in favor of standardized naming since we have several ID values:
OfferID (or Offer ID)
SwapID (or Swap ID)
PeerID (or Peer ID)

In this case, you can extract that it is the SwapID, because we don't have Offer and Peer IDs in the contract, but with standardized naming, the reader has to think less.

protocol/xmrmaker/swap_state.go
```diff
@@ -165,6 +167,98 @@ func newSwapStateFromStart(
 	return s, nil
 }
 
+func checkIfCompletedSuccessfully(
```
dimalinux, line 170: Can this 90 line function be broken up at all? The name says "check", implying that is is read-only, but if you scroll down enough, it is also updating the state of the swap.

noot: yeah, I broke it into two, one that just checks if the swap is already completed (funds claimed) and one that updates the state

protocol/xmrmaker/errors.go
noot: you mean naming the contract swap ID `SwapID`? I'd rather name it something more explicit like `OnChainID` or `ContractSwapID` (this is usually used in places anyways)

I think it's a bit out of scope for this PR but we can open an issue for it!

dimalinux: No, I was proposing that we be more rigorous in the order of the words so readers don't have to carefully parse sentences and so they can search the code base for keywords and word sequences. While I do like the phrase `OnChainID`, the code base is currently using the phrase "Swap ID" and sometimes "Contract Swap ID" to refer to the on-chain ID.  The error message can be rewritten from `swap with given ID in contract does not exist` to `contract swap ID does not exist`.

protocol/xmrmaker/instance_test.go
```diff
@@ -101,12 +108,12 @@ func newBackendAndNet(t *testing.T) (backend.Backend, *mockNet) {
 	rdb.EXPECT().PutCounterpartySwapKeys(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
 	rdb.EXPECT().DeleteSwap(gomock.Any()).Return(nil).AnyTimes()
 
-	extendedEC, err := extethclient.NewEthClient(context.Background(), env, common.DefaultEthEndpoint, pk)
+	extendedEC, err := extethclient.NewEthClient(ctx, env, common.DefaultEthEndpoint, pk)
```
dimalinux, line 111: Not related to this PR, but should we have changed `common.DefaultEthEndpoint` to use the `http` endpoint after the recent pull request removing the websockets dependency?

protocol/xmrmaker/instance_test.go
```diff
@@ -204,6 +223,8 @@ func TestInstance_createOngoingSwap(t *testing.T) {
 }
 
 func TestInstance_CompleteSwap(t *testing.T) {
+	_ = logging.SetLogLevel("protocol", "debug")
+	_ = logging.SetLogLevel("xmrmaker", "debug")
```
dimalinux, line 227: These might be more helpful at the test package level. Your call. I've been putting them in `init` methods at the test package level.  I'm slightly tempted to put the function that sets them all in `common`, and then have `main` and the tests use the same function to enable logging levels.

protocol/xmrmaker/swap_state.go
```diff
@@ -165,6 +168,104 @@ func newSwapStateFromStart(
...
+
+	// let's see if we have logs
+	logs, err := b.ETHClient().Raw().FilterLogs(b.Ctx(), filterQuery)
+	if err != nil {
+		return false, fmt.Errorf("failed to filter logs for topic %s: %s", claimedTopic, err)
+	}
+
+	log.Debugf("filtered for logs from block %s to head", filterQuery.FromBlock)
+
+	var foundClaimed *ethtypes.Log
```
dimalinux, line 216: This is declared as `*ethtypes.Log`, but the code is only using it as a boolean.  You could declare `claimFound := false` here and just set it to `true` when you find it, and then you won't need the nil check at the end.

dimalinux (approved):
Approved, but make sure to look at the comments and maybe take action on some of them before merging.

protocol/xmrmaker/swap_state.go
```diff
@@ -165,6 +168,104 @@ func newSwapStateFromStart(
...
+
+	if foundClaimed == nil {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// completeSwap marks the swap as completed and deletes it from the db.
+// note: it does not return an error if it fails to delete the swap from the db.
```
dimalinux, line 245: The note explains "what it does", but I suspect that there is a reason "why" which might also be helpful for the reader.

protocol/xmrmaker/errors.go
noot: ah, i see. okay i'll update the message!

protocol/xmrmaker/instance_test.go
noot: yeah we can, either one should still work though

protocol/xmrmaker/instance_test.go
noot: ah I just put these here to debug, will remove

PR #363, opened 2023-03-29
move stagenet from goerli to sepolia

### LINKED ISSUE #359: move to sepolia for ethereum test network
currently we're using goerli as our ethereum test network, but we should probably be using sepolia, as goerli will be deprecated, it's harder to get testnet eth, and it's supposed to be used for testing validators. sepolia is meant for dapp testing and it's easier to find testnet eth for it.

noot:
with this PR, we'll be using Sepolia for testnet/stagenet instead of Goerli. It's a lot easier to get SepETH and the public endpoints are more reliable. I've personally been using the PoW faucet.

closes #359

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -64,7 +64,7 @@ func MainnetConfig() *Config {
 	}
 }
 
-// StagenetConfig is the monero stagenet and ethereum Gorli configuration
+// StagenetConfig is the monero stagenet and ethereum Sepolia configuration
 func StagenetConfig() *Config {
 	return &Config{
 		Env:     Stagenet,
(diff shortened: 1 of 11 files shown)


common/config.go
```diff
@@ -83,8 +83,8 @@ func StagenetConfig() *Config {
 				Port: 38081,
 			},
 		},
-		SwapFactoryAddress:       ethcommon.HexToAddress("0x3d561C6f938aDBc45239772cc6A39e1Db7192154"),
-		ForwarderContractAddress: ethcommon.HexToAddress("0x4a707181842Ef084daFC90DeF367a1825eCcBCab"),
+		SwapFactoryAddress:       ethcommon.HexToAddress("0xE5656d3549478BdE248e5D7aE22ff8DCC5Cd11a3"),
+		ForwarderContractAddress: ethcommon.HexToAddress("0x171f294A0Ca7085Ce5F73DE1BE28b9721e007B94"),
```
dimalinux, line 87: Side note, not necessarily for this pull request: I'm thinking we can just get rid of the `ForwarderContractAddress` field just load it from the SwapFactory contract. Then we'd have one less thing to keep updated in the code.

docs/stagenet.md
```diff
@@ -31,21 +31,18 @@ For Linux 64-bit, you can do:
...
+   named `{DATA_DIR}/eth.key`. If you skip this step, a new wallet will be created for you
+   that you can transfer Sepolia ether to or fund directly in the next step.
 
-6. Fund your Goerli account using a faucet: 
-- https://goerli-faucet.pk910.de/
-- https://goerlifaucet.com/
-- https://goerli-faucet.mudit.blog/
+6. Fund your Sepolia account using a faucet: 
+- https://sepolia-faucet.pk910.de/
+- https://sepolia.dev/
```
dimalinux, line 43: https://sepoliafaucet.com/ is the Alchemy maintained faucet that gives out 0.5 SepoliaETH every 24 hours. I think it is the best option if you already have an Alchemy account. The sepolia.dev link is sending me to https://faucet.sepolia.dev/ which is giving a "bad gateway" error for me.

dimalinux (approved):

common/config.go
noot: will open an issue for this!

Issue #357, opened 2023-03-25
Reduce T0/T1 timeouts


dimalinux:
Part of this work will be removing the restriction that the delta between `T-1` and `T0` be identical to the delta beween `T0` and `T1`.

PR #364, opened 2023-03-29
update contract to allow setting t0/t1 separately

noot:
related to #357

will open for review after #363 is merged so I can redeploy the contract onto sepolia


ethereum/contracts/SwapFactory.sol
```diff
@@ -61,11 +60,14 @@ contract SwapFactory is ERC2771Context, Secp256k1 {
...
     // it returns the swap's ID.
+    // _timeoutDuration0: duration between the current timestamp and timeout0
+    // _timeoutDuration1: duration between timeout0 and timeout1
     function newSwap(
         bytes32 _pubKeyClaim,
         bytes32 _pubKeyRefund,
         address payable _claimer,
-        uint256 _timeoutDuration,
+        uint256 _timeoutDuration0,
+        uint256 _timeoutDuration1,
```
dimalinux, line 70: At some point we discussed setting these timeout durations at contract creation time instead of doing it on every single new swap. Then `newSwap` transactions would use slightly less gas, but we'd still be able to optimize each contract deployment for the block interval of the deployed chain. Would there be any downside in doing this?

noot: I can check the gas savings if we do that though, might be worth it if it's significant

dimalinux (commented):
Is updating `SwapTimeoutFromEnv` (split it into 2 functions?) to handle T0/T1 separately and minor updates to the XMR maker's `checkAndSetTimeouts` follow up work?

ethereum/contracts/SwapFactory.sol
dimalinux: I tried it out and it only saved 1387 gas units (went from 52594 to 51207). The test I ran moved the timeout values to the constructor and set them on immutable variables, then I removed them from the `newSwap` arguments and removed them from the emitted logs, since you can calculate their values based off of the block timestamp that the newSwap transaction was mined into.

If the timeout values change, we'd have to update the protocol version to prevent failed swaps, and if the protocol version is being changed, redeploying might not be that big of a deal. But I agree, if it is not saving an amount of gas that anyone cares about, flexibility is better.

dimalinux (approved):
Approved, but see my last comment. I'd be glad to re-review if we put more of the needed work into this PR.

noot:
@dimalinux yes, I'm not totally sure if we should change the timeouts yet, especially considering when relayers are used.

PR #366, opened 2023-03-31
XMR taker is relayer of last resort for the maker

### LINKED ISSUE #299: record peer IDs in swap info
We should probably start recording peer IDs in the swap info database entries. A changed peer ID on either end after a swap is accepted up until the XMR locked notification will prohibit the swap from working.  Takers, who have potential gas to lose, probably want that history information as well. They can use that history information to give reliable makers repeat business.

### LINKED ISSUE #367: Alice is relayer of last resort, but willing to relay claim for her own swap
Bob only seeks help from a relayer if he does not have sufficient ETH in the receiving account to perform his own claim. He will first try all advertised-in-the-DHT relayers that are not Alice for help relaying the claim. If, for some reason, that option fails, he'll ask Alice for help. At least if Alice is running our software, she will always be willing to help if she has sufficient ETH to cover the relayed claim when the asker is Bob, the counterparty to her own swap.

dimalinux:
When Bob does not have sufficient ETH in the receiving account to perform his own claim, he will first try all advertised-in-the-DHT relayers that are not Alice. If he doesn't succeed with any of them, he'll then ask Alice for help. At least if Alice is running our software, she will always be willing to help if she has sufficient ETH to cover the relayed claim. 

Closes #367
Closes #299


common/interfaces.go
```diff
@@ -21,13 +21,13 @@ type SwapState interface {
 // It is implemented by *xmrtaker.swapState and *xmrmaker.swapState
 type SwapStateNet interface {
 	HandleProtocolMessage(msg Message) error
-	ID() types.Hash
+	OfferID() types.Hash
```
dimalinux, line 24: This PR adds more use of the swap ID and the peer ID to our code base. Renaming adds clarity when it is called in places that are working with multiple ID types.

daemon/swap_daemon_test.go
```diff
@@ -69,6 +78,69 @@ func getSwapFactoryAddress(t *testing.T, ec *ethclient.Client) ethcommon.Address
...
+	//   refund:      46692
+	//   relayClaim: 130507
+	//
+	const (
+		aliceGasRation = 150000 // roughly 10% more than newSwap+setRead+refund
+	)
+	// We give Alice enough gas money to refund if needed, but not enough to
+	// relay a claim:
+	//    150000 - (53787 + 34452) = 61761
+	//
```
dimalinux, line 130: This didn't quite work on my first attempt. I had to add a minimum balance check to the relaying code, on top of this setup, to force the relaying over to a DHT relayer.  Otherwise the minimally funded Alice would try to relay using a low-ball gas price, and that would succeed in Ganache, but probably not on mainnet.

ethereum/block/wait_for_receipt.go
```diff
@@ -42,16 +42,11 @@ func WaitForReceipt(ctx context.Context, ec *ethclient.Client, txHash ethcommon.
 			continue
 		}
 		if receipt.Status != ethtypes.ReceiptStatusSuccessful {
-			err = fmt.Errorf("transaction failed (gas-lost=%d tx=%s block=%d), %w",
-				receipt.GasUsed, txHash, receipt.BlockNumber, ErrorFromBlock(ctx, ec, receipt))
+			err = fmt.Errorf("mined transaction failed (%s): %w",
```
dimalinux, line 45: Hopefully, the verb "mined" is still used even though Ethereum is now PoS.

ethereum/erc20_mock_test.go
```diff
@@ -13,6 +13,11 @@ import (
 )
 
 func TestSwapFactory_NewSwap_ERC20(t *testing.T) {
+	// the ETH version is working fine, but my changes to testNewSwap
+	// are making this test fail when it calls contract.NewSwap(...)
+	// with the error: "revert ERC20: insufficient allowance"
+	t.Skip("TODO: FIX ME")
```
dimalinux, line 19: I'll get this fixed before we merge. Let me know if you have any ideas on what the issue is.

ethereum/swap_factory_test.go
```diff
@@ -44,18 +46,58 @@ func testNewSwap(t *testing.T, asset ethcommon.Address) {
...
+		PubKeyRefund: pubKeyRefund,
+		Timeout0:     t0,
+		Timeout1:     t1,
+		Asset:        ethcommon.Address(types.EthAssetETH),
+		Value:        auth.Value,
+		Nonce:        nonce,
+	}
+
+	// validate our off-net calculation of the SwapID
+	require.Equal(t, types.Hash(swapID).Hex(), swap.SwapID().Hex())
```
dimalinux, line 100: I moved the off-net swap ID calculation (`SwapID` method above) into this package and testing it here allows us to compare the on-net/off-net calculation. This is also why I didn't want the swap to consist of mostly zero values.

ethereum/utils.go
```diff
@@ -49,6 +52,75 @@ func StageToString(stage byte) string {
 	}
 }
 
+// SwapID calculates and returns the same hashed swap identifier that newSwap
+// emits and that is used to index the on-net stage of a swap.
+func (sfs *SwapFactorySwap) SwapID() types.Hash {
```
dimalinux, line 57: This code was largely moved as-is from the `xmrmaker` package.

net/relay_test.go
```diff
@@ -16,7 +17,7 @@ import (
 func twoHostRelayerSetup(t *testing.T) (*Host, *Host) {
 	// ha is not a relayer
 	haCfg := basicTestConfig(t)
-	haCfg.IsRelayer = true
+	haCfg.IsRelayer = false
```
dimalinux, line 20: Code didn't match the comment

ethereum/block/wait_for_receipt.go
noot: I'd change this to "transaction was included but failed" or something like that

ethereum/erc20_mock_test.go
noot: check that the erc20 is approved to be spent by the swap contract by the caller of `newSwap` before the call to `newSwap` and that the addresses are correct

ethereum/utils.go
```diff
@@ -49,6 +52,75 @@ func StageToString(stage byte) string {
 	}
 }
 
+// SwapID calculates and returns the same hashed swap identifier that newSwap
+// emits and that is used to index the on-net stage of a swap.
```
noot, line 56:
```suggestion
// emits and that is used to track the on-chain stage of a swap.
```

nit :p

net/host.go
```diff
@@ -170,8 +163,10 @@ func (h *Host) SendSwapMessage(msg Message, id types.Hash) error {
 }
 
 // CloseProtocolStream closes the current swap protocol stream.
-func (h *Host) CloseProtocolStream(id types.Hash) {
-	swap, has := h.swaps[id]
+func (h *Host) CloseProtocolStream(offerID types.Hash) {
+	h.swapMu.Lock()
+	swap, has := h.swaps[offerID]
+	h.swapMu.Unlock()
```
noot, line 169: minor, but change to `RWMutex`?

net/relay.go
```diff
@@ -31,27 +31,51 @@ func (h *Host) DiscoverRelayers() ([]peer.ID, error) {
...
+	// While HandleRelayClaimRequest(...) will do lower level validation on the
+	// claim request, there are 2 validations best handled here:
+	// (1) If the network layer is not advertising that we are a relayer to the
+	//     DHT, we should not be getting claim requests targeted for open
+	//     relayers (i.e. requests that do not have the OfferID set).
+	// (2) If the request is purportedly from a maker to a taker of a current
+	//     swap, then:
+	//     (a) The swap should exist in our swaps map
+	//     (b) The peerID who sent us the request much match the peerID with
+	//         whom we are performing the swap.
```
noot, line 61: do we validate that the offerID and contractSwapID in the relay request match? (more of a note to self while reviewing)

net/relay.go
```diff
@@ -31,27 +31,51 @@ func (h *Host) DiscoverRelayers() ([]peer.ID, error) {
...
+	if req.OfferID == nil && !h.isRelayer {
 		return
+	} else if req.OfferID != nil {
+		h.swapMu.Lock()
+		swap, ok := h.swaps[*req.OfferID]
+		h.swapMu.Unlock()
+
+		if !ok || curPeer != swap.stream.Conn().RemotePeer() {
```
noot, line 69: why was the peer ID added to the *swap.Info struct when we can do the peer ID check here?

protocol/backend/backend.go
```diff
@@ -228,3 +232,29 @@ func (b *backend) ClearXMRDepositAddress(offerID types.Hash) {
...
+// HandleRelayClaimRequest validates and sends the transaction for a relay claim request
+func (b *backend) HandleRelayClaimRequest(request *message.RelayClaimRequest) (*message.RelayClaimResponse, error) {
+	// In the taker relay scenario, the net layer has already validated that we
+	// have an ongoing swap with the requesting peer that uses the passed
+	// offerID, but we have not verified that the claim in the swap matches the
+	// offerID. The backend, with its access to the recovery DB, is in the best
+	// position to perform this check. The remaining validations will be in the
+	// relayer library.
+	if request.OfferID != nil {
+		swapInfo, err := b.recoveryDB.GetContractSwapInfo(*request.OfferID)
```
noot, line 245: check that we're the xmrtaker for this also as a sanity check? I assume it would be caught in later validations but still

protocol/xmrmaker/claim.go
```diff
@@ -147,19 +162,55 @@ func (s *swapState) discoverRelayersAndClaim() (ethcommon.Hash, error) {
 			continue
 		}
 
-		return resp.TxHash, nil
+		log.Infof("DHT relayer's claim validated block=%d gas-used=%d tx=%s",
```
noot, line 165:
```suggestion
		log.Infof("relayer's claim included in block=%d gas-used=%d tx=%s",
```

protocol/xmrmaker/claim.go
```diff
@@ -147,19 +162,55 @@ func (s *swapState) discoverRelayersAndClaim() (ethcommon.Hash, error) {
...
+		s.contractSwap,
+		&secret,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	receipt, err := s.relayClaimWithXMRTaker(request)
+	if err != nil {
+		log.Warnf("failed to relay claim with XMR taker: %s", err)
```
noot, line 200:
```suggestion
		log.Warnf("failed to relay claim with counterparty: %w", err)
```

relayer/claim_request.go
```diff
@@ -8,14 +8,16 @@ import (
...
 
 	"github.com/athanorlabs/atomic-swap/coins"
 	contracts "github.com/athanorlabs/atomic-swap/ethereum"
 	"github.com/athanorlabs/atomic-swap/net/message"
 )
 
 const (
-	relayedClaimGas = 70000
+	relayedClaimGas   = 70000  // worst case gas usage for the claimRelayer swapFactory call
+	forwarderClaimGas = 195000 // worst case gas usage when using forwarder to claim
```
noot, line 20: how did you get this? unit tests?

net/relay.go
noot: upon reviewing I don't think this is checked unless I missed it, pls add

noot: okay it's in `HandleRelayClaimRequest` I believe

noot (commented):
I'm not sure why peer ID was added to `swap.Info`? I don't see it being used anywhere, since the peer ID check is done in the network package

net/relay.go
dimalinux: The problem isn't the peer ID check. Since we are using a separate stream for the claim, the maker needs to know the peer ID that he is sending the claim request to.

net/relay.go
dimalinux: Yes. I did it there, because the backend has access to the database. The checks here are only network/peer related checks.

ethereum/swap_factory_test.go
```diff
@@ -44,18 +46,71 @@ func testNewSwap(t *testing.T, asset ethcommon.Address) {
...
+	require.NoError(t, err)
+	value, err := rand.Prime(rand.Reader, 53) // (2^54 - 1) / 10^18 ~= .018 ETH
+	require.NoError(t, err)
+
+	isEthAsset := asset == ethAssetAddress
+
+	if isEthAsset {
+		auth.Value = value
+	} else {
+		value = big.NewInt(0)
```
dimalinux, line 73: The original test was always calling `newSwap` with a value of zero. To get this test, which is now testing our off-net calculation of the SwapID, working with ERC20 tokens, just setting the value back to zero, the way it previously was, removed the transaction revert. I didn't dig though the solidity, but I'm going to guess it failed because the message sender didn't have the tokens to lock.

ethereum/swap_factory_test.go
```diff
@@ -44,18 +46,71 @@ func testNewSwap(t *testing.T, asset ethcommon.Address) {
...
+		Timeout1:     t1,
+		Asset:        ethcommon.Address(types.EthAssetETH),
+		Value:        value,
+		Nonce:        nonce,
+	}
+
+	// validate our off-net calculation of the SwapID
+	if !isEthAsset {
+		t.Skip("off-net calculation of swapID doesn't match")
+	}
```
dimalinux, line 112: I'll have to look at this again tomorrow with fresh eyes. Any idea what field in `swap` isn't matching up to generate the correct hash when this test is running with ERC20 tokens?

protocol/backend/backend.go
dimalinux: The net layer, in the hash of swap connections by offerID, now tracks who initiated the swap connection. In order for us to accept a  taker-specific relay request, the code now checks that the entry exists in the hash, and that the swap entry was created because we (the taker) initiated the connection. I added that today when converting the lock on the net-layer's hash to a RW lock.

ethereum/block/wait_for_receipt.go
dimalinux: Done.

net/host.go
dimalinux: Yeah, I was thinking that too. It's a RW mutex now.

protocol/xmrmaker/claim.go
dimalinux: We have a separate log depending on whether the Maker got the receipt from a DHT relayer or from the Taker. I wanted the logged message to be unique, so the log reader can tell which type was used. In this case the word "validated" does not mean "mined" or "included in block", it means that we did additional validation in `waitForClaimReceipt` which checks the receipt status and the logs. Do you have a suggestion to improve the message with that additional context in mind?

protocol/xmrmaker/claim.go
```diff
@@ -91,50 +95,65 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
+	request.OfferID = &s.offer.ID
+	defer func() { request.OfferID = nil }()
 
-	if len(relayers) == 0 {
-		return ethcommon.Hash{}, errors.New("no relayers found to submit claim to")
-	}
-	log.Debugf("Found %d relayers to submit claim to", len(relayers))
-
-	forwarderAddress, err := s.Contract().TrustedForwarder(&bind.CallOpts{Context: s.ctx})
+	response, err := s.Backend.SubmitClaimToRelayer(s.info.PeerID, request)
```
dimalinux, line 108: Here is one of the two places that the PeerID in the info is used.

protocol/xmrmaker/claim.go
```diff
@@ -91,50 +95,65 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(relayers) == 0 {
+		return nil, errors.New("no relayers found to submit claim to")
+	}
+	log.Debugf("Found %d relayers to submit claim to", len(relayers))
+	for _, relayerPeerID := range relayers {
+		if relayerPeerID == s.info.PeerID {
```
dimalinux, line 145: Here is the 2nd place that the PeerID in the info is used.

protocol/xmrmaker/claim.go
dimalinux: The code shifted around when I made the Taker the relayer of last resort. Since we know which counterparty was relaying, isn't the more specific message better? It matches the function names and other logged messages with the more specific wording.

relayer/claim_request.go
dimalinux: Yes, it is roughly 50% more than what I was sampling in unit tests:
```text
130507 * 1.5
195760.5
```
It's only used in this logic, where I was probably leaning on the high side:
```go
	txCost := new(big.Int).Mul(suggestedGasPrice, big.NewInt(forwarderClaimGas))
	if balance.Cmp(txCost) < 0 {
		return fmt.Errorf("balance %s ETH is under the minimum %s ETH to relay claim",
			coins.FmtWeiAsETH(balance), coins.FmtWeiAsETH(txCost))
	}
```
Let me know what you think the value should be.

scripts/run-unit-tests.sh
```diff
@@ -11,7 +11,7 @@ start-ganache
 # run unit tests
 echo "running unit tests..."
 rm -f coverage.txt
-go test ./... -v -short -timeout=30m -count=1 -covermode=atomic -coverprofile=coverage.txt
```
dimalinux, line 14: As of go 1.10 you can use the `-coverpkg` flag I added to get cross-package coverage stats. Project wide, it didn't make a huge difference (1.4%), but it now shows coverage in `protocol/xmrmaker/claim.go` that we get from the daemon package.

net/initiate.go
```diff
@@ -87,9 +88,11 @@ func (h *Host) handleProtocolStream(stream libp2pnetwork.Stream) {
...
 
+	curPeer := stream.Conn().RemotePeer()
+
 	log.Debug(
 		"received message from peer, peer=",
-		stream.Conn().RemotePeer(),
+		curPeer,
 		" type=",
 		message.TypeToString(msg.Type()),
 	)
```
noot, line 98: change this to a `Debugf`? formatting is kinda weird rn lol

net/relay.go
```diff
@@ -68,6 +93,9 @@ func (h *Host) handleRelayStream(stream libp2pnetwork.Stream) {
 
 // SubmitClaimToRelayer sends a request to relay a swap claim to a peer.
 func (h *Host) SubmitClaimToRelayer(relayerID peer.ID, request *RelayClaimRequest) (*RelayClaimResponse, error) {
+	// The timeout should be short enough, that the Maker can try multiple relayers
+	// before T1 expires even if the receiving node accepts the relay request and
+	// just sits on it without doing anything.
```
noot, line 98: note that the timeout here is specifically for connecting and opening the stream, not for receiving the response. we'd need a different timer for that

net/types.go
```diff
@@ -27,16 +29,21 @@ type (
...
 	HandleRelayClaimRequest(msg *RelayClaimRequest) (*RelayClaimResponse, error)
 }
 
 type swap struct {
 	swapState SwapState
 	stream    libp2pnetwork.Stream
+	// isTaker is true if we initiate the swap (created the outbound stream). If
+	// the value is false, it is not safe to assume that we are the maker, as
+	// the swap entry will get created before validation of the first received
+	// message.
```
noot, line 47: how would we not be the maker if this is false?

protocol/xmrmaker/claim.go
```diff
@@ -33,59 +34,62 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
 	if s.offerExtra.UseRelayer || weiBalance.Cmp(big.NewInt(0)) == 0 {
 		// relayer fee was set or we had insufficient funds to claim without a relayer
 		// TODO: Sufficient funds check above should be more specific
-		txHash, err = s.discoverRelayersAndClaim()
+		receipt, err = s.claimWithRelay()
 		if err != nil {
-			log.Warnf("failed to claim using relayers: %s", err)
+			return nil, fmt.Errorf("failed to claim using relayers: %w", err)
 		}
+		log.Debugf("relayed claim transaction %s", common.ReceiptInfo(receipt))
```
noot, line 69:
```suggestion
		log.Infof("claim transaction was relayed:\ %s", common.ReceiptInfo(receipt))
```

protocol/xmrmaker/claim.go
```diff
@@ -150,19 +169,59 @@ func (s *swapState) discoverRelayersAndClaim() (ethcommon.Hash, error) {
...
+	)
+	if err != nil {
+		return nil, err
 	}
 
-	return ethcommon.Hash{}, errors.New("failed to submit transaction to any relayer")
+	receipt, err := s.claimWithAdvertisedRelayers(request)
+	if err != nil {
+		log.Warnf("failed to relay with DHT-advertised relayers: %s", err)
+		log.Infof("falling back to XMR taker as relayer")
```
noot, line 211:
```suggestion
		log.Infof("falling back to swap counterparty as relayer")
```

protocol/xmrmaker/claim.go
```diff
@@ -94,50 +98,65 @@ func (s *swapState) claimFunds() (ethcommon.Hash, error) {
...
+		return nil, err
+	}
+
+	if len(relayers) == 0 {
+		return nil, errors.New("no relayers found to submit claim to")
+	}
+	log.Debugf("Found %d relayers to submit claim to", len(relayers))
+	for _, relayerPeerID := range relayers {
+		if relayerPeerID == s.info.PeerID {
+			log.Debugf("skipping DHT-advertised relayer that is the XMR taker of the claim")
```
noot, line 149:
```suggestion
			log.Debugf("skipping DHT-advertised relayer that is our swap counterparty")
```

protocol/xmrmaker/claim.go
noot: okay, i'd say `included in block and validated` then? because when i read the initial log, it wasn't clear that the transaction had actually been included in the chain, it just sounds like something was validated

protocol/xmrmaker/claim.go
noot: it's not necessary to store the peer ID here - the network is able to get it based on the offer ID and the stored swap info. I suppose that might make the network interface more complex though, since you'd need a function that submits the request based on peer ID, and one that submits it based on offer ID.

relayer/claim_request.go
noot: if `130507` was the actual gas value you got, i'd just put that or *slightly* higher (like 131000). the gas cost will always be the same unless the contract changes

relayer/submit_transaction.go
```diff
@@ -57,6 +61,10 @@ func ValidateAndSendTransaction(
 		return nil, err
 	}
 
+	if err = checkForMinClaimBalance(ctx, ec); err != nil {
+		return nil, err
+	}
+
 	// Lock the wallet's nonce until we get a receipt
```
noot, line 68: use the same gas price in this function in `txOpts` below, otherwise it could fail still

noot: `txOpts.GasPrice = suggestedGasPrice`

noot (commented):
looks good! only some minor comments.

net/types.go
dimalinux: I'll remove that message.  I thought we were creating the entry before processing the initial received message, but I just looked at the code and it is fully processing the first message before creating the entry and calling `handleProtocolStreamInner`.

net/relay.go
```diff
@@ -34,27 +34,52 @@ func (h *Host) DiscoverRelayers() ([]peer.ID, error) {
...
+	if req.OfferID == nil && !h.isRelayer {
 		return
+	} else if req.OfferID != nil {
+		h.swapMu.RLock()
+		swap, ok := h.swaps[*req.OfferID]
+		h.swapMu.RUnlock()
+
+		found := ok && swap.isTaker
+		if !found || curPeer != swap.stream.Conn().RemotePeer() {
+			log.Debugf("received taker-specific claim request from invalid peer=%s offerID=%s swap-found=%t",
```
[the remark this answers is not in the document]
dimalinux: Resolving. Implemented exactly as suggested.

protocol/xmrmaker/claim.go
dimalinux: New version is:
```
log.Infof("DHT relayer's claim included and validated %s", common.ReceiptInfo(receipt))
```

protocol/xmrmaker/claim.go
dimalinux: Resolving. This log ended up in a new location and I used the suggested wording on that location.

relayer/claim_request.go
dimalinux: I reduced it from 195000 to 156000 (roughly 16% above measured gas units).

net/initiate.go
dimalinux: Done, new message is:
```
log.Debugf("received message from peer=%s type=%s", curPeer, message.TypeToString(msg.Type()))
```

net/relay.go
dimalinux: Comment updated with a link to the issue to fix the problem across all 3 streams.

net/types.go
dimalinux: Verbiage about the maker completely removed.

relayer/submit_transaction.go
dimalinux: Unfortunately, it won't fail. (What I discovered before adding this function.) It it failed when creating the transaction, that would be awesome, but unless you don't have the base fee (which you probably do) it will create a transaction that might not get mined quickly or at all.

That said, I implemented the suggestion in this commit:
https://github.com/AthanorLabs/atomic-swap/pull/366/commits/d23664fc7030078cf637a33d0ebe71905f679146

noot (approved):

Issue #326, opened 2023-02-21
First mainnet beta release


dimalinux:
- [x] have release tags on our own dependent repos
- [x] make sure all RPC API are hardened and don't need near-term breaking changes
* ... add items here as we discover them ...

PR #368, opened 2023-03-31
add license header to files; update readme disclaimer

noot:
related to #326


dimalinux (approved):
Since this doesn't change the license, I'm approving it, but I'm not sure the LGPL is appropriate for golang programs which do not use dynamic linking. Since the majority of open source go code is using the MIT or similar licenses, they won't be legally able to use a package of the project as a library or borrow a few files from the code base, yet we probably want to allow usage to help the Monero community.

https://www.makeworld.space/2021/01/lgpl_go.html
https://medium.com/@henvic/opensource-and-go-what-license-f6b36c201854

noot:
@dimalinux from what I understand, LGPL code can be used in other applications as a library without needing to reveal the source code as long as the library code has not been modified. in the case where the code has been modified from the source, the application would need to release that modified LGPL code, and in the case of static linking then they'd need to release the entire app's source. so there is nothing preventing a project that isn't (L)GPL from using this code unmodified. if there is ever a case where the license seriously becomes an issue, I'm happy to discuss it further.

PR #371, opened 2023-04-03
remove `swap_refund` in favour of `swap_cancel`

noot:
These two functions essentially did the same thing, except `swap_refund` was xmrtaker-only.

### DIFF
--- a/docs/rpc.md
+++ b/docs/rpc.md
@@ -362,10 +362,10 @@ curl -X POST http://127.0.0.1:5002 -d '{"jsonrpc":"2.0","id":"0","method":"perso
 
 ### `swap_cancel`
 
-Attempts to cancel an ongoing swap.
+Attempts to cancel an ongoing swap. Note that depending on the swap's stage, it may not be possible to cancel the swap and receive a refund.
 
 Parameters:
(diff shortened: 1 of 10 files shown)


docs/rpc.md
```diff
@@ -362,10 +362,10 @@ curl -X POST http://127.0.0.1:5002 -d '{"jsonrpc":"2.0","id":"0","method":"perso
 
 ### `swap_cancel`
 
-Attempts to cancel an ongoing swap.
+Attempts to cancel an ongoing swap. Note that depending on the swap's stage, it may not be possible to cancel the swap and receive a refund.
```
dimalinux, line 365: It might be helpful to show the non-refund scenario. I assume the outer JSON RPC has an error?

dimalinux (approved):

docs/rpc.md
noot: what this does is basically queue up an `Exit` event, so the next event will be the swap attempting to exit. that doesn't return the resulting exit state, so the user would need to get the swap status another way. I think in general this function's UX needs to be improved, since it basically looks like it's hanging while attempting to exit. I'll open an issue for improving/testing this function

PR #372, opened 2023-04-04
SwapFactory to SwapCreator rename, upgrade to solc 0.8.19

dimalinux:
Changes:
* Renames SwapFactory to SwapCreator
* Upgrades from solc 0.8.17 to 0.8.19
* Updates stagenet/Sepolia contract addresses
* Harmonizing naming of the address variables for the contracts
* Provides github secrets support for an `ETH_SEPOLIA_ENDPOINT` environment variable used by unit tests validating the deployed contract


docs/developing.md
```diff
@@ -12,19 +12,28 @@ You can use the script `scripts/setup-env.sh` to quickly set up local monerod-te
 ganache instances. If you need to later kill `ganache`, `monerod`, `monero-wallet-rpc`, or
 `swapd` instances, you can use `scripts/cleanup-test-processes.sh`.
 
-## Deploying or using deployed SwapFactory.sol
-
-The swap program uses a "factory" contract for the Ethereum side to reduce gas costs from deploying a new contract for each swap. The contract can be found in [here](../ethereum/contracts/SwapFactory.sol). For each new swap, the eth-holding party will call `NewSwap` on the factory contract, initiating a swap instance inside the contract.
```
dimalinux, line 17: Other than "factory" => "creator" substitution, no other content changes were made in this file. Added newlines are causing the larger diff. This was a 340 character long line before.

.github/workflows/unit-tests.yml
```diff
@@ -59,6 +59,7 @@ jobs:
       - name: Run unit tests
         env:
           ETH_MAINNET_ENDPOINT: ${{ secrets.ETH_MAINNET_ENDPOINT }}
+          ETH_SEPOLIA_ENDPOINT: ${{ secrets.ETH_SEPOLIA_ENDPOINT }}
```
dimalinux, line 62: This is all setup if you add a github actions secret for it. The stagenet contract check tests will happen even if it is not set, but they will use `https://rpc.sepolia.org/`, which may or may not be reliable enough.

dimalinux:
The -2.29% project-wide code-coverage stat is some kind of bug. If you look at the impacted files, there is nothing that can explain where it was lost.

.github/workflows/unit-tests.yml
noot: added

net/host.go
```diff
@@ -26,7 +26,7 @@ import (
 const (
 	// ProtocolID is the base atomic swap network protocol ID prefix. The full ID
 	// includes the chain ID at the end.
-	ProtocolID          = "/atomic-swap/0.2"
+	ProtocolID          = "/atomic-swap/0.3"
```
noot, line 29: is this in anticipation of the release? :p

protocol/backend/backend.go
```diff
@@ -100,34 +100,34 @@ type backend struct {
 	perSwapXMRDepositAddrRWMu sync.RWMutex
 	perSwapXMRDepositAddr     map[types.Hash]*mcrypto.Address
 
-	// swap contract
-	contract     *contracts.SwapFactory
-	contractAddr ethcommon.Address
-	swapTimeout  time.Duration
+	// swap swapCreator
```
noot, line 103: swap contract?

noot (approved):
looks good!

net/host.go
dimalinux: Yes, but going forward we need to update the protocol version every time p2p field names change or when the SwapCreator contract bytecode changes (and this PR does both).

protocol/backend/backend.go
dimalinux: I updated it to "swap contract".

PR #378, opened 2023-04-09
net: add initiate, query and relay timeouts

### LINKED ISSUE #375: Timeout for responses on p2p net stream responses
Query, init and relay all need timeouts on the response.

noot:
closes #375


net/query.go
```diff
@@ -56,17 +57,23 @@ func (h *Host) Query(who peer.ID) (*QueryResponse, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("error reading QueryResponse: %w", err)
-	}
-
-	resp, ok := msg.(*QueryResponse)
-	if !ok {
-		return nil, fmt.Errorf("expected %s message but received %s",
-			message.TypeToString(message.QueryResponseType),
-			message.TypeToString(msg.Type()))
+	const queryResponseTimeout = time.Second * 15
```
dimalinux, line 60: We have a `queryTimeout` at the package level that appears to be only for the connect. That should probably be renamed. I'd either put both of them together at the package level, or move the connect timeout to the function level.

net/host.go
```diff
@@ -220,3 +220,24 @@ func readStreamMessage(stream libp2pnetwork.Stream, maxMessageSize uint32) (comm
...
+			msg, err := readStreamMessage(stream, maxMessageSize)
+			if err != nil {
+				log.Warnf("failed to read stream message: %s", err)
+				close(ch)
+				return
+			}
+
+			ch <- msg
+		}
+	}()
```
dimalinux, line 240: How does this goroutine get cleaned up when the Hosts's context is cancelled. Will the stream be closed elsewhere, triggering `readStreamMessage` to return an error, and that will in turn close the channel?

net/host.go
```diff
@@ -220,3 +220,24 @@ func readStreamMessage(stream libp2pnetwork.Stream, maxMessageSize uint32) (comm
...
+// nextStreamMessage returns a channel that will receive the next message from the stream.
+// if there is an error reading from the stream, the channel will be closed, thus
+// the received value will be nil.
+func nextStreamMessage(stream libp2pnetwork.Stream, maxMessageSize uint32) <-chan common.Message {
+	ch := make(chan common.Message)
+	go func() {
+		for {
+			msg, err := readStreamMessage(stream, maxMessageSize)
+			if err != nil {
+				log.Warnf("failed to read stream message: %s", err)
```
dimalinux, line 233: There is at least one place this method is being called, where it is not an error for `readStreamMessage` to return `io.EOF`.  Our underlying `p2pnet.ReadStreamMessage` method already differentiates between `io.EOF` (stream closed between messages) and `io.ErrUnexpectedEOF` (truncated message).

I didn't research how to fix the issue. Maybe we don't write a warning here if the error is `io.EOF`, and we ensure that any channel reader logs a good message when the lack of response is an error?

net/initiate.go
```diff
@@ -164,3 +181,16 @@ func (h *Host) handleProtocolStreamInner(stream libp2pnetwork.Stream, s SwapStat
...
 	}
 }
+
+func (h *Host) handleProtocolStreamClose(stream libp2pnetwork.Stream, s SwapState) {
+	log.Debugf("closing stream: peer=%s protocol=%s", stream.Conn().RemotePeer(), stream.Protocol())
+	_ = stream.Close()
+
+	log.Debugf("exiting swap...")
+	if err := s.Exit(); err != nil {
+		log.Errorf("failed to exit protocol: err=%s", err)
```
dimalinux, line 191: We've been leaving the `err=` part of on newer messages. People already expect the error after the colon.

net/relay.go
```diff
@@ -123,17 +124,23 @@ func (h *Host) SubmitClaimToRelayer(relayerID peer.ID, request *RelayClaimReques
 }
 
 func receiveRelayClaimResponse(stream libp2pnetwork.Stream) (*RelayClaimResponse, error) {
-	msg, err := readStreamMessage(stream, maxRelayMessageSize)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read RelayClaimResponse: %w", err)
-	}
+	const relayResponseTimeout = time.Minute
```
dimalinux, line 127: Let's say it takes 20 minutes for the contract to get set to to ready, this will give Bob time to try just under 100 relayers assuming they all timeout? I may have the math wrong, that's why I'm asking.

net/host.go
noot: yeah, when the host is closed, it should close all of its connections and streams and thus it should get an EOF.

net/relay.go
noot: what do you mean? bob is only able to try to claim after ready and before t1, which is at minimum an hour right now. I reduced this to 45 seconds which would allow him to try 80 relayers if they're all done sequentially. we can add some concurrency potentially

net/host.go
noot: I updated this to not log anything and just return if the errors is `io.EOF`. another option would be to have this function return a channel for any errors that happen, which the caller can read from. what do  you think?

net/query.go
noot: updated to use one `connectionTimeout` var

net/relay.go
dimalinux: In my example, Bob had 100 minutes until T1, because T0 was still 40 minutes away when the contract was set to ready, plus the additional 60 minutes to reach T1.

net/host.go
dimalinux: I think the way you did it is good. Only the channel message reader has the context to know if the absence of a next message is an error or not. The reader's job is probably simpler without the 2nd channel since, from a protocol standpoint, they don't really care what the specific error was for their decision logic.


knows if the absence of a next message is an error or not.

dimalinux (approved):

PR #379, opened 2023-04-10
swapd version & shutdown

### LINKED ISSUE #280: stop RPC API to shutdown swapd
Not critical, but it might be nice to have a RPC method to shut `swapd` down.

### LINKED ISSUE #377: swapcli version and swapd version RPC call
The `swapcli version` command would report the following:
* The version of `swapcli` itself
* The version of `swapd` (assuming swapcli's version is compatible enough to query it from `swapd`)
* P2p protocol version (since 2 versions of `swapd` may be protocol compatible)
* The current environment (dev, stagenet, mainnet)
* The swap creator address that `swapd` is currently using (as the user may have configured it to an alternative address on start)

Maybe values. The risk in including these is that they could cause the whole RPC version request to fail:
* monerod version
* eth chainID (which is already in the protocol ID) and network ID?
* time delta since the last XMR and ETH block (probably nice to have, but arguably not version information)

stubbrn:
closes #280
closes #377

### DIFF
--- a/daemon/swap_daemon.go
+++ b/daemon/swap_daemon.go
@@ -45,8 +45,7 @@ type SwapdConfig struct {
 
 // RunSwapDaemon assembles and runs a swapd instance blocking until swapd is
 // shut down. Typically, shutdown happens because a signal handler cancels the
-// passed in context, but we may add a shutdown method to the RPC instance in
-// the future.
+// passed in context, or when the shutdown RPC method is called.
 func RunSwapDaemon(ctx context.Context, conf *SwapdConfig) (err error) {
 	// Note: err can be modified in defer blocks, so it needs to be a named return
 	//       value above.
@@ -158,8 +157,14 @@ func RunSwapDaemon(ctx context.Context, conf *SwapdConfig) (err error) {
 	})
 
 	log.Infof("starting swapd with data-dir %s", conf.EnvConf.DataDir)
-	err = rpcServer.Start() // blocks until server is shutdown or context is cancelled
-	if err != nil && !errors.Is(err, http.ErrServerClosed) {
+
+	err = rpcServer.Start()
+
+	if errors.Is(err, http.ErrServerClosed) {
+		// Returning nil to exit the program with exit code 0
+		// and to avoid a FATAL ERROR entry in log files
+		return nil
+	} else if err != nil {
 		return err
 	}
 
--- a/daemon/swap_daemon.go
+++ b/daemon/swap_daemon.go
@@ -161,11 +161,12 @@ func RunSwapDaemon(ctx context.Context, conf *SwapdConfig) (err error) {
 	err = rpcServer.Start()
 
 	if errors.Is(err, http.ErrServerClosed) {
-		// Returning nil to exit the program with exit code 0
+		// Set err to nil to exit the program with exit code 0
 		// and to avoid a FATAL ERROR entry in log files
-		return nil
-	} else if err != nil {
-		return err
+		//
+		// NOTE: The ErrServerClosed error happens only when the server is told
+		// to shutdown or close
+		err = nil
 	}
 
 	// err can get set in defer blocks, so return err or use an empty
--- a/daemon/swap_daemon_test.go
+++ b/daemon/swap_daemon_test.go
@@ -8,8 +8,6 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"math/big"
-	"os"
-	"os/signal"
 	"sync"
 	"syscall"
 	"testing"
@@ -181,16 +179,6 @@ func createTestConf(t *testing.T, ethKey *ecdsa.PrivateKey) *SwapdConfig {
 
 func launchDaemons(t *testing.T, timeout time.Duration, configs ...*SwapdConfig) context.Context {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	swapdCtx, swapdCancel := context.WithCancel(ctx)
-
-	go func() { // Handle signals
-		sigc := make(chan os.Signal, 1)
-		signal.Ignore(syscall.SIGHUP)
-		signal.Notify(sigc, syscall.SIGINT, syscall.SIGTERM)
-
-		<-sigc
-		swapdCancel()
-	}()
 
 	var wg sync.WaitGroup
 	t.Cleanup(func() {
@@ -207,7 +195,7 @@ func launchDaemons(t *testing.T, timeout time.Duration, configs ...*SwapdConfig)
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			err := RunSwapDaemon(swapdCtx, conf)
+			err := RunSwapDaemon(ctx, conf)
 			require.ErrorIs(t, err, context.Canceled)
 		}()
 		WaitForSwapdStart(t, conf.RPCPort)
--- a/daemon/swap_daemon_test.go
+++ b/daemon/swap_daemon_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/athanorlabs/atomic-swap/ethereum/block"
 	"github.com/athanorlabs/atomic-swap/ethereum/extethclient"
 	"github.com/athanorlabs/atomic-swap/monero"
+	"github.com/athanorlabs/atomic-swap/net"
 	"github.com/athanorlabs/atomic-swap/relayer"
 	"github.com/athanorlabs/atomic-swap/rpcclient"
 	"github.com/athanorlabs/atomic-swap/rpcclient/wsclient"
@@ -578,4 +579,21 @@ func TestRunSwapDaemon_CharlieIsBroke_AliceRelays(t *testing.T) {
 	bobBalance, err := bobConf.EthereumClient.Balance(ctx)
 	require.NoError(t, err)
 	require.Equal(t, bobExpectedBal.Text('f'), coins.FmtWeiAsETH(bobBalance))
+}
+
+// Tests the version and shutdown RPC methods
+func TestRunSwapDaemon_RPC_Version(t *testing.T) {
+	conf := createTestConf(t, tests.GetMakerTestKey(t))
+	protocolVersion := fmt.Sprintf("%s/%d", net.ProtocolID, conf.EthereumClient.ChainID())
+	timeout := time.Minute
+	ctx := launchDaemons(t, timeout, conf)
+
+	c := rpcclient.NewClient(ctx, fmt.Sprintf("http://127.0.0.1:%d", conf.RPCPort))
+	versionResp, err := c.Version()
+	require.NoError(t, err)
+
+	require.Equal(t, conf.EnvConf.Env, versionResp.Env)
+	require.NotEmpty(t, versionResp.SwapdVersion)
+	require.Equal(t, conf.EnvConf.SwapCreatorAddr, versionResp.SwapCreatorAddr)
+	require.Equal(t, protocolVersion, versionResp.P2PVersion)
 }
--- a/daemon/swap_daemon_test.go
+++ b/daemon/swap_daemon_test.go
@@ -8,7 +8,10 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"math/big"
+	"os"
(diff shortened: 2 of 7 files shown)


daemon/swap_daemon.go
```diff
@@ -158,8 +157,14 @@ func RunSwapDaemon(ctx context.Context, conf *SwapdConfig) (err error) {
...
 	log.Infof("starting swapd with data-dir %s", conf.EnvConf.DataDir)
-	err = rpcServer.Start() // blocks until server is shutdown or context is cancelled
-	if err != nil && !errors.Is(err, http.ErrServerClosed) {
+
+	err = rpcServer.Start()
+
+	if errors.Is(err, http.ErrServerClosed) {
+		// Returning nil to exit the program with exit code 0
+		// and to avoid a FATAL ERROR entry in log files
+		return nil
```
dimalinux, line 166: Yeah, this needed a fix.  Instead of returning `nil` here, assign `err` to `nil`. The error handling in this function is a little delicate, as errors can be appended by defer statements above to track problems shutting down other subsystems.  We can't return `nil`, because then any multierror appended in the defer statements won't be seen by the caller.

Given that you were confused and might not be the last person to be confused, perhaps we should convert `return` statements in this function to be empty (or at least the final return statement)? This function was originally written without naming the return value and then converted, but I didn't change all the return statements during conversion. Use your best discretion about what will be most obvious to readers.

rpc/daemon.go
```diff
@@ -0,0 +1,70 @@
...
+	P2PVersion      string             `json:"p2pVersion" validate:"required"`
+	Env             common.Environment `json:"env" validate:"required"`
+	SwapCreatorAddr ethcommon.Address  `json:"swapCreatorAddress" validate:"required"`
+}
+
+// Version returns version & misc info about swapd and its dependencies
+func (s *DaemonService) Version(_ *http.Request, req *VersionRequest, resp *VersionResponse) error {
+	resp.SwapdVersion = cliutil.GetVersion()
+
+	resp.P2PVersion = func() string {
```
dimalinux, line 53: For any given versioned build of `swapd`, the version of the P2P library is fixed, so we don't need that particular piece of information.

What we want is the `protocolID` passed to the p2p network layer (it is passed in `RunSwapDaemon` to the `net.NewHost` call). This identifier determines whether 2 versions of `swapd` can communicate with each other for running swaps. Two different versions of `swapd` can still communicate and run swaps together if we haven't bumped that protocol version. @noot Correct me if I'm wrong on anything in this paragraph. I was surprised when you told me that bootnodes can provide DHT support for multiple protocol IDs simultaneously.

rpc/server.go
```diff
@@ -101,11 +102,17 @@ func NewServer(cfg *Config) (*Server, error) {
...
 
-	return &Server{
+	s := &Server{
 		ctx:        cfg.Ctx,
 		listener:   ln,
 		httpServer: server,
-	}, nil
+	}
+
+	if err = rpcServer.RegisterService(NewDaemonService(s, cfg), "daemon"); err != nil {
```
dimalinux, line 111: In the interest of self documenting code, where it is easy to track who uses what information, we should probably pass down a separate config type with the minimal set of needed fields for `NewDaemonService`. If it was using the majority of the fields, it might not be worth the trouble, but I suspect that we are only using a small subset here.

rpc/daemon.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ShutdownRequest ...
+type ShutdownRequest struct{}
+
+// ShutdownResponse ...
+type ShutdownResponse struct{}
+
+// Shutdown swapd
+func (s *DaemonService) Shutdown(_ *http.Request, req *ShutdownRequest, resp *ShutdownResponse) error {
+	return s.server.Stop()
```
dimalinux, line 35: The method here is probably the simplest way for this section of the code to shutdown `swapd`, but just so you know your options, it is not the only way. If `DaemonService` had access to the cancel function of the outermost context, canceling it will also shut everything down.

That is how our signal handler shuts everything down:
https://github.com/AthanorLabs/atomic-swap/blob/master/cmd/swapd/signal_handler.go#L28

rpc/server.go
stubbrn: I ended up passing `cfg.ProtocolBackend` to the service as this is the only field that's used (same as the other services).

rpc/daemon.go
```diff
@@ -0,0 +1,56 @@
...
+// VersionResponse ...
+type VersionResponse struct {
+	SwapdVersion    string             `json:"swapdVersion" validate:"required"`
+	P2PVersion      string             `json:"p2pVersion" validate:"required"`
+	Env             common.Environment `json:"env" validate:"required"`
+	SwapCreatorAddr ethcommon.Address  `json:"swapCreatorAddress" validate:"required"`
+}
+
+// Version returns version & misc info about swapd and its dependencies
+func (s *DaemonService) Version(_ *http.Request, req *VersionRequest, resp *VersionResponse) error {
```
noot, line 50:
```suggestion
func (s *DaemonService) Version(_ *http.Request, _ *interface{}, resp *VersionResponse) error {
```

similarly here

rpcclient/daemon.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/athanorlabs/atomic-swap/rpc"
+)
+
+// Shutdown swapd
+func (c *Client) Shutdown() error {
+	const (
+		method = "daemon_shutdown"
+	)
+	if err := c.Post(method, nil, nil); err != nil {
+		return nil // Does not expect a response from swapd
```
noot, line 13: we should probably still return the error here, in case of other errors - I don't think it errors just because no response is expected? correct me if I'm wrong

noot (commented):
thanks for the contribution! only a few minor comments :)

rpcclient/daemon.go
dimalinux: Correct.  There is always an HTTP response to any HTTP request, even if the response body is empty. I could see making an exception for `errors.Is(err, syscall.ECONNREFUSED)` and not treating that as an error, as it may indicate that the server was already shutdown, but I would shy away from that implementation until someone shows their use case of why the behavior is needed.

rpc/daemon.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// ShutdownRequest ...
+type ShutdownRequest struct{}
+
+// ShutdownResponse ...
+type ShutdownResponse struct{}
+
+// Shutdown swapd
+func (s *DaemonService) Shutdown(_ *http.Request, req *ShutdownRequest, resp *ShutdownResponse) error {
```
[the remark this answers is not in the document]
dimalinux: What @noot wrote, but you can make it even simpler with `any`:
```
func (s *DaemonService) Shutdown(_ *http.Request, _ any, _ any) error {
```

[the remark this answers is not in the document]
dimalinux: I'm doing some testing in this PR's code, and don't rush to make this change. Our libraries fail to auto-identify the method name if the request doesn't have a concrete type. Figuring that out, and if we would even want to change the library behavior, is probably out-of-scope for this PR.

rpc/daemon.go
dimalinux: Here's a unit test that you can add to the `daemon` package to test the RPC version call end-to-end. You'll need add the chainID to the protocol version for it to work:
```
// Tests the version and shutdown RPC methods
func TestRunSwapDaemon_RPC_Version(t *testing.T) {
	conf := createTestConf(t, tests.GetMakerTestKey(t))
	protocolVersion := fmt.Sprintf("%s/%d", net.ProtocolID, conf.EthereumClient.ChainID())
	timeout := time.Minute
	ctx := launchDaemons(t, timeout, conf)

	c := rpcclient.NewClient(ctx, fmt.Sprintf("http://127.0.0.1:%d", conf.RPCPort))
	versionResp, err := c.Version()
	require.NoError(t, err)

	require.Equal(t, conf.EnvConf.Env, versionResp.Env)
	require.NotEmpty(t, versionResp.SwapdVersion)
	require.Equal(t, conf.EnvConf.SwapCreatorAddr, versionResp.SwapCreatorAddr)
	require.Equal(t, protocolVersion, versionResp.P2PVersion)
}
```
Interesting note.  I was playing around with @noot's suggestion.  You have to use `_ *any` on the request type.  If you try to use `_ any`, you'll get the error `rpc: can't find method "daemon.Version"`.

rpcclient/daemon.go
dimalinux: Here's an end-to-end unit test for the shutdown RPC method that can be placed in the `daemon` package:
```
// Tests the shutdown RPC method
func TestRunSwapDaemon_RPC_Shutdown(t *testing.T) {
	conf := createTestConf(t, tests.GetMakerTestKey(t))
	timeout := time.Minute
	ctx := launchDaemons(t, timeout, conf)

	c := rpcclient.NewClient(ctx, fmt.Sprintf("http://127.0.0.1:%d", conf.RPCPort))
	err := c.Shutdown()
	require.NoError(t, err)

	err = c.Shutdown()
	require.ErrorIs(t, err, syscall.ECONNREFUSED)
}
```
The test isn't working yet, so it may need further modifications. In the test helper method `launchDaemons`, you'll also need to change a couple lines to allow `nil` as an allowed return value from `RunSwapDaemon`:
```
			err := RunSwapDaemon(ctx, conf)
			if err != nil {
				require.ErrorIs(t, err, context.Canceled)
			}
```
I can debug it more if you want. It almost looks like the server is shutting down before it sends the RPC response. If that's the case, we'll need to figure out how to trigger a graceful shutdown that finishes serving the request, or switch to the alternative shutdown method of calling cancel on the outermost server context.

rpc/daemon.go
```diff
@@ -0,0 +1,49 @@
...
+func NewDaemonService(server *Server, pb ProtocolBackend) *DaemonService {
+	return &DaemonService{
+		server,
+		pb,
+	}
+}
+
+// Shutdown swapd
+func (s *DaemonService) Shutdown(req *http.Request, _ *any, _ *any) error {
+	return syscall.Kill(syscall.Getpid(), syscall.SIGINT)
```
stubbrn, line 31: Replaced `server.Stop()` with a signal (handled here https://github.com/AthanorLabs/atomic-swap/blob/master/cmd/swapd/signal_handler.go) because it was shutting down the RPC server before sending a response to the shutdown request, and caused a EOF error on the client.

Maybe it would be better to manage to call the cancel function of the outermost context directly instead of relying on the signal handler.

daemon/swap_daemon_test.go
```diff
@@ -177,6 +181,16 @@ func createTestConf(t *testing.T, ethKey *ecdsa.PrivateKey) *SwapdConfig {
...
+	swapdCtx, swapdCancel := context.WithCancel(ctx)
+
+	go func() { // Handle signals
+		sigc := make(chan os.Signal, 1)
+		signal.Ignore(syscall.SIGHUP)
+		signal.Notify(sigc, syscall.SIGINT, syscall.SIGTERM)
+
+		<-sigc
+		swapdCancel()
+	}()
```
stubbrn, line 193: The daemon running in tests didn't have a signal handler, added a simple one here (for the shutdown test).

If we end up keeping the signal method to shutdown the server, it would be better to use the actual signal handler used in swapd, I didn't do that here because it caused circular dependencies errors, but we could work on that if you want.

noot (approved):
looks good to me!

rpc/daemon.go
dimalinux: I thought about it some more, and we can just cancel the RPC server's context, not the outermost context. The code changes are relatively minor: [PR on top of this PR's branch](https://github.com/stubbrn/atomic-swap/pull/1/files)

daemon/swap_daemon_test.go
dimalinux: Yeah, I usually keep signal handlers in the `main` package of my daemons.  For unit tests, we run multiple swapd instances in the same linux process. Check out my other comment. I think I found a solution that is only minor changes from the work that you've already done and doesn't send a signal.

dimalinux (approved):

PR #380, opened 2023-04-11
ERC20 balance support

dimalinux:
Added `--token` flag to `swapcli`'s `balances` request allowing the user to query their token balance.  Also added an end-to-end ERC20 unit test to the `daemon` package.

### DIFF
--- a/cmd/swapcli/main.go
+++ b/cmd/swapcli/main.go
@@ -40,6 +40,7 @@ const (
 	flagProvidesAmount = "provides-amount"
 	flagUseRelayer     = "use-relayer"
 	flagSearchTime     = "search-time"
+	flagToken          = "token"
 	flagDetached       = "detached"
 )
 
@@ -76,6 +77,11 @@ var (
 				Action:  runBalances,
 				Flags: []cli.Flag{
 					swapdPortFlag,
+					&cli.StringSliceFlag{
+						Name:    flagToken,
+						Aliases: []string{"t"},
+						Usage:   "Token address to include in the balance response",
+					},
 				},
 			},
 			{
@@ -375,7 +381,18 @@ func runPeers(ctx *cli.Context) error {
 
 func runBalances(ctx *cli.Context) error {
 	c := newRRPClient(ctx)
-	balances, err := c.Balances()
+
+	request := &rpctypes.BalancesRequest{}
+	tokens := ctx.StringSlice(flagToken)
+	for _, token := range tokens {
+		asset, err := types.NewEthAsset(token)
+		if err != nil {
+			return err
+		}
+		request.TokenAddrs = append(request.TokenAddrs, asset)
+	}
+
+	balances, err := c.Balances(request)
 	if err != nil {
 		return err
 	}
@@ -383,6 +400,15 @@ func runBalances(ctx *cli.Context) error {
 	fmt.Printf("Ethereum address: %s\n", balances.EthAddress)
 	fmt.Printf("ETH Balance: %s\n", balances.WeiBalance.AsEtherString())
 	fmt.Println()
+
+	for _, tokenBalance := range balances.TokenBalances {
+		fmt.Printf("Token: %s\n", tokenBalance.TokenInfo.Address)
+		fmt.Printf("Name: %q\n", tokenBalance.TokenInfo.Name)
+		fmt.Printf("Symbol: %q\n", tokenBalance.TokenInfo.Symbol)
+		fmt.Printf("Balance: %s\n", tokenBalance.AsStandard().Text('f'))
+		fmt.Println()
+	}
+
 	fmt.Printf("Monero address: %s\n", balances.MoneroAddress)
 	fmt.Printf("XMR Balance: %s\n", balances.PiconeroBalance.AsMoneroString())
 	fmt.Printf("Unlocked XMR balance: %s\n",
@@ -393,7 +419,7 @@ func runBalances(ctx *cli.Context) error {
 
 func runETHAddress(ctx *cli.Context) error {
 	c := newRRPClient(ctx)
-	balances, err := c.Balances()
+	balances, err := c.Balances(nil)
 	if err != nil {
 		return err
 	}
@@ -408,7 +434,7 @@ func runETHAddress(ctx *cli.Context) error {
 
 func runXMRAddress(ctx *cli.Context) error {
 	c := newRRPClient(ctx)
-	balances, err := c.Balances()
+	balances, err := c.Balances(nil)
 	if err != nil {
 		return err
 	}
--- a/coins/coins.go
+++ b/coins/coins.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 
 	"github.com/cockroachdb/apd/v3"
+	ethcommon "github.com/ethereum/go-ethereum/common"
 )
 
 // PiconeroAmount represents some amount of piconero (the smallest denomination of monero)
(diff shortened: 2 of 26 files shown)


coins/test_support.go
```diff
@@ -43,3 +43,14 @@ func IntToWei(amount int64) *WeiAmount {
 	}
 	return NewWeiAmount(big.NewInt(amount))
 }
+
+// Sub returns the value of a-b in a newly allocated WeiAmount variable.
+// If a or b is NaN, this function will panic.
```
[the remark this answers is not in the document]
dimalinux: Done: I added a check to `coins.ValidatePositive` that we use on min/max/provided/exchange-rate values to verify that the passed value is finite.

cmd/swapcli/main.go
```diff
@@ -519,35 +549,55 @@ func runMake(ctx *cli.Context) error {
 		return err
 	}
 
+	ethAssetStr := ctx.String("eth-asset")
```
dimalinux, line 552: This just got moved around. We should update that to use a constant. It might be nice if the flag passed to `balances` (currently, `--token`) matched this flag to `make`. I'm not sure `--eth-asset` is an intuitive flag name for `balances` though. What would you think about using `--token` for both?

noot: i think `--token` for this flag would make sense as well, might be more intuitive actually. if it's not set then we're just using eth, otherwise a token.

cmd/swapcli/main.go
```diff
@@ -883,7 +957,7 @@ func printOffer(o *types.Offer, index int, indent string) error {
 	}
 
 	fmt.Printf("%sOffer ID: %s\n", indent, o.ID)
-	fmt.Printf("%sProvides: %s\n", indent, o.Provides)
+	fmt.Printf("%sProvides: %s\n", indent, o.Provides) // TODO: This code is not correctly handling ERC20 tokens.
```
noot, line 960: do you want to update this here?

coins/coins.go
```diff
@@ -182,105 +201,172 @@ func (a *WeiAmount) BigInt() *big.Int {
...
 // in ETH.
 func FmtWeiAsETH(wei *big.Int) string {
 	return NewWeiAmount(wei).AsEther().Text('f')
 }
 
+// ERC20TokenInfo stores the token contract address and basic info that most
+// ERC20 tokens support
+type ERC20TokenInfo struct {
+	Address     ethcommon.Address `json:"address" validate:"required"`
+	NumDecimals uint8             `json:"decimals"` // digits after the Decimal point needed for smallest denomination
```
noot, line 264: not required?

common/types/ethasset.go
```diff
@@ -13,10 +13,30 @@ import (
 // EthAsset represents an Ethereum asset (ETH or a token address)
 type EthAsset ethcommon.Address
 
+// NewEthAsset converts an unquoted string to the EthAsset type
```
noot, line 16: a json string? or eth address?

ethereum/extethclient/eth_wallet_client.go
```diff
@@ -148,37 +149,56 @@ func (c *ethClient) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
...
-	name, err = tokenContract.Name(c.CallOpts(ctx))
+
+	symbol, err := tokenContract.Symbol(c.CallOpts(ctx))
 	if err != nil {
-		return "", "", 18, err
+		return nil, err
 	}
-	symbol, err = tokenContract.Symbol(c.CallOpts(ctx))
+
+	// TODO: Do we support ERC20 tokens that do not have this method?
```
noot, line 186: are decimals 18 if they're not specified? not sure though

protocol/xmrtaker/swap_state_test.go
```diff
@@ -425,11 +449,29 @@ func TestExit_invalidNextMessageType(t *testing.T) {
...
+	err = s.setXMRMakerKeys(
+		xmrmakerKeysAndProof.PublicKeyPair.SpendKey(),
+		xmrmakerKeysAndProof.PrivateKeyPair.ViewKey(),
+		xmrmakerKeysAndProof.Secp256k1PublicKey,
+	)
 	require.NoError(t, err)
-	allowance, err := contract.Allowance(&bind.CallOpts{}, s.ETHClient().Address(), s.SwapCreatorAddr())
+
+	// approve is called by NewSwap() in lockAsset()
+	s.lockAsset()
```
noot, line 468: no error check?

rpc/swap.go
```diff
@@ -50,9 +50,11 @@ func NewSwapService(
 }
 
 // PastSwap represents a past swap returned by swap_getPast.
+// TODO: This struct does not have needed fields for swaps of ERC20 tokens.
```
noot, line 53: such as? maybe add to comment?

tests/erc20_integration_test.go
```diff
@@ -5,70 +5,106 @@ package tests
...
-	balance := new(big.Int).Mul(big.NewInt(9999999), decimals)
-	erc20Addr, erc20Tx, _, err := contracts.DeployERC20Mock(auth, conn, "ERC20Mock", "MOCK", addr, balance)
+	// Query Charlie's Ethereum address
+	charlieCli := rpcclient.NewClient(ctx, defaultCharlieSwapdEndpoint)
+	balResp, err := charlieCli.Balances(nil)
 	require.NoError(t, err)
-	_, err = block.WaitForReceipt(context.Background(), conn, erc20Tx.Hash())
+	charlieAddr := balResp.EthAddress
+
+	// Transfer half of the supply to Charlie (using Alice's extended ethereum client)
```
noot, line 58: why does charlie need tokens? he doesn't swap right?

coins/coins.go
dimalinux: Required would mean that the value can never be zero.

tests/erc20_integration_test.go
dimalinux: There is a test where Alice and Charlie try to take Bob's offer at the same time.

coins/coins.go
dimalinux: I found an example: https://etherscan.io/address/0xaec2e87e0a235266d9c5adc9deb4b2e29b54d009#code
I don't think there are any ones that people would want to use though, if we decide to ban 0-decimal tokens.

ethereum/extethclient/eth_wallet_client.go
dimalinux: With the normal ERC20 contracts, the decimals function will always be present and it will default to 18 decimals if you don't configure it otherwise in the constructor. As best as I can guess from [the standard](https://eips.ethereum.org/EIPS/eip-20), if the decimals function does not exist on the contract, you would treat it as zero decimals and there would never be a fractional part on the user representation. (But the standard doesn't explicitly state that, and I was unable to find any deployed token where the function is missing.)

rpc/swap.go
dimalinux: Sorry, I forgot to remove that TODO. I added `EthAsset` to both `PastSwap` and `OngoingSwap` as part of this PR.

cmd/swapcli/main.go
dimalinux: Done. We are now using `--token` for `make` and `balances`.

common/types/ethasset.go
dimalinux: I removed the method. I was using it outside of tests in an earlier version of the PR.

protocol/xmrtaker/swap_state_test.go
dimalinux: Added.

noot (approved):
looks good, nice work!

daemon/test_support.go
```diff
@@ -6,18 +6,109 @@
...
+
+	"github.com/athanorlabs/atomic-swap/common"
+	contracts "github.com/athanorlabs/atomic-swap/ethereum"
+	"github.com/athanorlabs/atomic-swap/ethereum/extethclient"
+	"github.com/athanorlabs/atomic-swap/monero"
+	"github.com/athanorlabs/atomic-swap/rpcclient"
+	"github.com/athanorlabs/atomic-swap/tests"
+)
+
+// map indexes for our mock tokens
```
dimalinux, line 31: DAI has 18 decimals and USDT has 6, so it seemed like a good initial pair to be writing lots of tests against.

PR #382, opened 2023-04-14
contract updates and cleanup

noot:
- update SwapCreator.sol to reject 0-value swaps
- update SwapCreator.sol to use solidity errors instead of require statements (this is recommended w solidity 0.8+ and saves gas)
- update solidity versions to 0.8.19
- add mainnet forwarder address + update forwarder code check to check for this address if we're on mainnet


scripts/license.sh
dimalinux, line 1: Can you document what this script is attempting to do?

dimalinux: When you make the change, also add the execute bit before committing: `chmod u+x scripts/license.sh` (assuming it is designed to be run as an executable).

ethereum/check_swap_creator_contract.go
```diff
@@ -104,6 +105,18 @@ func CheckForwarderContractCode(
...
+	// with solidity 0.8.7, but we're using 0.8.19 for SwapCreator.sol, we can just
+	// check that the address is what's expected.
+	chainID, err := ec.ChainID(ctx)
+	if err != nil {
+		return err
+	}
+
+	if contractAddr == common.MainnetConfig().ForwarderAddr && chainID.Uint64() == common.MainnetChainID {
+		return nil
+	}
```
dimalinux, line 118: Maybe not in this PR, but since the behavior can change depending on the compiler version, it might be worth figuring out a way to get the identically compiled contract in dev/stagenet.

scripts/license.sh
noot: oops i didn't mean to commit this lol, will remove

dimalinux (approved):

PR #383, opened 2023-04-14
implement database rpc service for querying recovery DB

### LINKED ISSUE #370: implement rpc method(s) for querying database
There should be an easy way for the user to query items from the database, especially w.r.t swap secrets, in case the user needs to manually call the contract for whatever reason. There should be a method `personal_queryDatabase` that accepts a database method + offer ID (if required) that fetches data from the database.

noot:
as title says - add methods for querying on-chain swap info and swap secrets from the db

closes #370


rpc/database.go
```diff
@@ -0,0 +1,116 @@
...
+// GetContractSwapInfoRequest ...
+type GetContractSwapInfoRequest struct {
+	ID types.Hash `json:"id" validate:"required"`
+}
+
+// GetContractSwapInfoResponse ...
+type GetContractSwapInfoResponse struct {
+	StartNumber     *big.Int                   `json:"startNumber" validate:"required"`
+	SwapID          types.Hash                 `json:"swapID" validate:"required"`
+	Swap            *contracts.SwapCreatorSwap `json:"swap" validate:"required"`
```
dimalinux, line 46: Why are we passing the `SwapID` if we are also passing the object that the `SwapID` can be quickly derived from?

rpc/database.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// GetContractSwapInfoRequest ...
+type GetContractSwapInfoRequest struct {
+	ID types.Hash `json:"id" validate:"required"`
+}
+
+// GetContractSwapInfoResponse ...
+type GetContractSwapInfoResponse struct {
+	StartNumber     *big.Int                   `json:"startNumber" validate:"required"`
```
dimalinux, line 44: Can we document this? Is it the height to start scanning monero blocks at?

rpc/database.go
```diff
@@ -0,0 +1,116 @@
...
+	resp.StartNumber = info.StartNumber
+	resp.SwapID = info.SwapID
+	resp.Swap = info.Swap
+	resp.SwapCreatorAddr = info.SwapCreatorAddr
+	return nil
+}
+
+// GetSwapPrivateKeyRequest ...
+type GetSwapPrivateKeyRequest struct {
+	ID types.Hash `json:"id" validate:"required"`
```
dimalinux, line 70: Which of our many ID values is being passed here?

rpc/database.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// GetSwapPrivateKeyResponse ...
+type GetSwapPrivateKeyResponse struct {
+	PrivateKey *mcrypto.PrivateSpendKey `json:"privateKey" validate:"required"`
```
dimalinux, line 75: Given that Monero has 2 private keys and the spend key isn't always derived from the view key, can we be precise of the field name?

rpc/database.go
noot: the user would need to manually write code to hash the struct to check their on-chain status which is bad

rpc/database.go
noot: it's documented here: https://github.com/AthanorLabs/atomic-swap/blob/master/db/types.go#L16 should i copy paste the comments?

rpc/database.go
noot: `OfferID`/`SwapID`

rpc/database.go
noot: maybe this should just be called `SwapSecret` or something since it's never actually used as a monero spend key?

rpc/database.go
dimalinux: Apologies. I was confused thinking this was a request when I wrote that. I'll resolve this.

rpc/database.go
dimalinux: I forgot that we were storing that. `newSwapBlockNum` would be unambiguous, albeit 4 characters longer. How does the value get used? Would it make sense the store `newSwapTxID` instead?

rpc/database.go
dimalinux: I don't understand. I just dug through the code and that ID is the OfferID. The SwapID is this: https://github.com/AthanorLabs/atomic-swap/blob/master/ethereum/utils.go#L58-L125.

Can we call the field `OfferID` (json `offerID`)?

rpc/database.go
dimalinux: Based on your response and studying the surrounding code, I now understand that this is "our" half of the private spend key. Here's a way you could blend your idea of using the phrase `SwapSecret` and be obvious as to whether the secret is ours/theirs/combined/view/spend:

```
type GetOurSwapSecretRequest struct {
	OfferID types.Hash `json:"offerID" validate:"required"`
}

type GetOurSwapSecretResponse struct {
	PrivSpendKey *mcrypto.PrivateSpendKey `json:"secret" validate:"required"`
}

type GetTheirSwapSecretRequest struct {
	OfferID types.Hash `json:"offerID" validate:"required"`
}

type GetTheirSwapSecretResponse struct {
	PrivSpendKey *mcrypto.PrivateSpendKey `json:"PrivSpendKey" validate:"required"`
}
```
I don't have anything against the word "counterparty", but these names are already on the long side and "counterparty" doesn't have a nice symmetric antonym.

rpc/database.go
noot: oh I've been referring to that ID as the contractSwapID lol. yeah I'll change it to offerID

rpc/database.go
noot: I don't really think the user needs this necessarily, but I'll change the name to `startBlockNumber`

rpc/database.go
noot: updated to be `GetSwapSecret`, I removed the counterparty secret functions since they aren't actually usable by the user to recover

dimalinux (approved):

PR #388, opened 2023-04-18
small additions to contract unit tests

noot:
- test that we can't claim before ready/t0
- test that we can't refund between t0 to t1

these are implicitly tested elsewhere but wanted to make it explicit in the contract tests


ethereum/swap_creator_test.go
```diff
@@ -279,6 +279,14 @@ func testClaim(t *testing.T, asset ethcommon.Address, newLogIndex int, value *bi
 		Nonce:        nonce,
 	}
 
+	// ensure we can't claim before setting contract to Ready
+	var s [32]byte
+	secret := proof.Secret()
+	copy(s[:], secret[:])
+	txOpts = *authOrig
+	_, err = contract.Claim(&txOpts, swap, s)
```
dimalinux, line 287: `proof.Secret()` returns an array and `contract.Claim` accepts an array. The copy on line 285 isn't needed.

ethereum/swap_creator_test.go
```diff
@@ -427,8 +432,8 @@ func testRefundAfterT1(t *testing.T, asset ethcommon.Address, erc20Contract *Tes
 	t0, t1, err := GetTimeoutsFromLog(receipt.Logs[newLogIndex])
 	require.NoError(t, err)
 
-	<-time.After(time.Until(time.Unix(t1.Int64()+1, 0)))
-
+	// ensure we can't refund before T1
```
dimalinux, line 435: The next line is waiting for T0 to expire. Should this message say?
```
// verify that we can't refund between T0 and T1 if contract is not set to ready
```

ethereum/swap_creator_test.go
```diff
@@ -441,12 +446,19 @@ func testRefundAfterT1(t *testing.T, asset ethcommon.Address, erc20Contract *Tes
...
-	// now let's try to refund
 	var s [32]byte
 	secret := proof.Secret()
 	copy(s[:], secret[:])
 	tx, err = contract.Refund(auth, swap, s)
 	require.NoError(t, err)
+	_, err = block.WaitForReceipt(context.Background(), conn, tx.Hash())
+	require.ErrorContains(t, err, "VM Exception while processing transaction: revert")
+
+	<-time.After(time.Until(time.Unix(t1.Int64()+1, 0)))
```
dimalinux, line 457: Did the +1 make a difference? More curious than anything else.

ethereum/swap_creator_test.go
```diff
@@ -409,7 +414,7 @@ func testRefundAfterT1(t *testing.T, asset ethcommon.Address, erc20Contract *Tes
 	}
 
 	nonce := big.NewInt(0)
-	timeout := big.NewInt(1) // T1 expires before we get the receipt for new_swap TX
+	timeout := big.NewInt(3) // T1 expires before we get the receipt for new_swap TX
```
dimalinux, line 417: Is that comment still true after increasing the timeout to 3 seconds?

dimalinux (approved):
Approved, but I left comments that should be acted on before merging.

ethereum/swap_creator_test.go
noot: yeah, it sometimes fails without it due to submitting a bit too early

ethereum/swap_creator_test.go
noot: no, can remove

PR #390, opened 2023-04-18
upgraded go version and dependencies

dimalinux:
I saw the docker image PR using 1.19 and was reminded that Go 1.20 has been released for a couple months and that it is probably time to upgrade. I also refreshed our other dependencies


noot:
cool i'll update the docker image to 1.20 when this is done!

ethereum/block/error_from_block.go
```diff
@@ -26,21 +27,21 @@ func ErrorFromBlock(ctx context.Context, ec *ethclient.Client, receipt *ethtypes
 	if err != nil {
 		return fmt.Errorf("unable to determine error in mined block, %w", err)
 	}
-	txMessage, err := tx.AsMessage(ethtypes.LatestSignerForChainID(chainID), nil)
+	txMessage, err := core.TransactionToMessage(tx, ethtypes.LatestSignerForChainID(chainID), nil)
```
dimalinux, line 30: This PR is upgrading `go-ethereum` from `v1.11.3` to `v1.11.5`. It looks like this change happened in the `v1.11.4` released.

noot: I think we were using that version for abigen - can you make sure installing abigen with this version and generating the bindings works? otherwise we need to pin the abigen version

dimalinux: Yes, `abigen` builds fine and running `make bindings` generates identical binding files to what we were already using.

noot (approved):

PR #391, opened 2023-04-18
dockerfile adjustments

dimalinux:
This is a PR to `noot/docker`, not a PR directly to `master`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,25 @@
-FROM golang:1.19.8-alpine as builder
-RUN apk add gcc musl-dev linux-headers git bash curl
-RUN git clone https://github.com/AthanorLabs/atomic-swap
-WORKDIR atomic-swap
-RUN bash scripts/install-monero-linux.sh
-RUN go build --tags=prod,netgo,osusergo --ldflags '-extldflags "-static"' ./cmd/swapd/
+FROM golang:latest as builder
+# Download monero-wallet-rpc. We need bzip2 to unpack the tar file.
+RUN apt update && apt install -y bzip2
+RUN arch=$(uname -m | sed 's/x86_64/linux64/; s/aarch64/linuxarm8/') && \
+    curl -sSL "https://downloads.getmonero.org/cli/${arch}" -o monero.tar.bz2
+RUN tar xvjf monero.tar.bz2 --no-anchored monero-wallet-rpc --strip-components=1
 
-FROM ubuntu:22.04
+# Build the swapd and swapcli binaries
+RUN git clone --depth=1 https://github.com/AthanorLabs/atomic-swap
+WORKDIR atomic-swap
+RUN make build
+
+FROM debian:bullseye-slim
 RUN apt-get update && apt-get install -y ca-certificates
-WORKDIR /root/
-COPY --from=builder /go/atomic-swap/monero-bin/monero-wallet-rpc /usr/local/bin
-COPY --from=builder /go/atomic-swap/swapd .
-RUN ls /usr/local/bin
-RUN ./swapd --version
-CMD ["/root/swapd", "--env", "stagenet", "--ethereum-endpoint", "https://rpc.sepolia.org/"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,15 +1,25 @@
-FROM golang:1.19.8-alpine as builder
```
dimalinux, line 1: We might as well use the default golang container that is based off of Debian bullseye and has glibc installed, since it matches what our final container will be. This is definitely the way to go if the final container will be alpine, but `monero-wallet-rpc` depends on glibc, so it's not the best option.

Dockerfile
```diff
@@ -1,15 +1,25 @@
...
+RUN apt update && apt install -y bzip2
+RUN arch=$(uname -m | sed 's/x86_64/linux64/; s/aarch64/linuxarm8/') && \
+    curl -sSL "https://downloads.getmonero.org/cli/${arch}" -o monero.tar.bz2
+RUN tar xvjf monero.tar.bz2 --no-anchored monero-wallet-rpc --strip-components=1
+
+# Build the swapd and swapcli binaries
+RUN git clone --depth=1 https://github.com/AthanorLabs/atomic-swap
 WORKDIR atomic-swap
-RUN bash scripts/install-monero-linux.sh
-RUN go build --tags=prod,netgo,osusergo --ldflags '-extldflags "-static"' ./cmd/swapd/
```
dimalinux, line 6: Since we know the container has to have glibc installed inside it for `monero-wallet-rpc`, a static build will just produce a larger container without removing the glibc dependency. The easiest way to get a static build if we didn't depend on glibc for `monero-wallet-rpc` would be:
```
CGO_ENABLED=0 make build
```

Dockerfile
```diff
@@ -1,15 +1,25 @@
-FROM golang:1.19.8-alpine as builder
-RUN apk add gcc musl-dev linux-headers git bash curl
-RUN git clone https://github.com/AthanorLabs/atomic-swap
+FROM golang:latest as builder
```
noot, line 1: should we pin the version to 1.20?

Dockerfile
noot: I don't think we can disable CGO since go-ethereum depends on it, which is why I had the static build flags

Dockerfile
```diff
@@ -1,15 +1,25 @@
...
-CMD ["/root/swapd", "--env", "stagenet", "--ethereum-endpoint", "https://rpc.sepolia.org/"]
\ No newline at end of file
+COPY --from=builder /go/monero-wallet-rpc /usr/local/bin/
+COPY --from=builder /go/atomic-swap/bin/ /usr/local/bin/
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+RUN groupadd --gid "${USER_GID}" atomic && \
+    useradd --no-log-init --home-dir /atomic-swap \
+    --uid "${USER_UID}" --gid "${USER_GID}" -m atomic
+USER atomic
```
noot, line 22: why is this added?

Dockerfile
dimalinux: We either don't use the features in `go-ethereum` that require CGO, or their code is configured to fallback to a go implementation of certain parts, as our binaries build okay:
```
rm -rf bin
CGO_ENABLED=0 make build
ldd bin/swapd
```
and ldd outputs:
```
not a dynamic executable
```

Dockerfile
dimalinux: It is up to you. There is a small, but non-zero, risk that our code stops building or working when 1.21 is released. In which case, we'd want to fix it as quickly as possible.  Locking it would prevent that risk, but also mean that we need to update the Dockerfile on every release. It would be nice if we could get CI to automatically move to the lastest stable/released version of go.  I didn't look into it when creating the 1.20 go upgrade PR.

dimalinux: I pinned it to 1.20, since I was making one more edit to the file anyways.

Dockerfile
noot: good to know!

PR #395, opened 2023-04-19
Simulate relayer tx and check return value

### LINKED ISSUE #360: relayer: simulate transaction and check return value before submitting transaction
The relayer currently performs signature checks and other checks before submitting a tx, but it does not actually check whether the internal call to `claimRelayer` will succeed before submitting the tx, possibly resulting in wasted gas if the internal call fails.

Note that even if the internal call fails, the tx will NOT revert, so it will not be caught by `EstimateGas` etc.

To check if the internal call to `claimRelayer` will succeed (on the current head block), we need to call `ethclient.CallContract` (https://pkg.go.dev/github.com/ethereum/go-ethereum/ethclient#Client.CallContract) simulating the tx and checking the abi-encoded return value, which will be the return value of `forwarder.Execute()` https://github.com/AthanorLabs/go-relayer/blob/master/impls/gsnforwarder/contracts/Forwarder.sol#L64 
The `success` bool returned will tell us whether the call to `claimRelayer` succeeded or not.

stubbrn:
I'll add tests when you think the code is adequate

closes #360

### DIFF
--- a/relayer/submit_transaction.go
+++ b/relayer/submit_transaction.go
@@ -5,12 +5,17 @@ package relayer
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
+	"strings"
 
 	"github.com/athanorlabs/go-relayer/impls/gsnforwarder"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 
 	"github.com/athanorlabs/atomic-swap/coins"
 	"github.com/athanorlabs/atomic-swap/common"
@@ -76,6 +81,19 @@ func ValidateAndSendTransaction(
 	}
 	txOpts.GasPrice = gasPrice
 
+	_, err = simulateExecute(
+		ctx,
+		ec,
+		&reqForwarderAddr,
+		txOpts,
+		*forwarderReq,
+		*domainSeparator,
+		req.Signature,
+	)
+	if err != nil {
+		return nil, err
+	}
+
 	tx, err := reqForwarder.Execute(
 		txOpts,
 		*forwarderReq,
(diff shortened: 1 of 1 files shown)


relayer/submit_transaction.go
```diff
@@ -119,3 +137,68 @@ func checkForMinClaimBalance(ctx context.Context, ec extethclient.EthClient) (*b
...
+// https://pkg.go.dev/github.com/ethereum/go-ethereum/ethclient#Client.CallContract
+func simulateExecute(
+	ctx context.Context,
+	ec extethclient.EthClient,
+	reqForwarderAddr *ethcommon.Address,
+	txOpts *bind.TransactOpts,
+	forwarderReq gsnforwarder.IForwarderForwardRequest,
+	domainSeparator [32]byte,
+	sig []byte,
+) ([]byte, error) {
```
noot, line 152:
```suggestion
) error {
```

I'd get rid of this first return value since it's never used

noot (commented):
looks good so far!

relayer/submit_transaction.go
```diff
@@ -119,3 +137,68 @@ func checkForMinClaimBalance(ctx context.Context, ec extethclient.EthClient) (*b
...
+	ForwarderABI, err := abi.JSON(strings.NewReader(gsnforwarder.ForwarderMetaData.ABI))
```
dimalinux, line 153: You can simplify that with `gsnforwarder.ForwarderMetaData.GetAbi()`? I would use a lower-case letter for a local variable, unless it confirms to some style guide that you think is useful and want to convince us to adopt.

dimalinux:
> I'll add tests when you think the code is adequate

This should help your testing: https://github.com/stubbrn/atomic-swap/pull/2

relayer/submit_transaction.go
stubbrn: Nice, thank you

dimalinux (approved):
Looks good to me.  Thanks!

protocol/xmrmaker/claim_test.go
noot, line 1: why was this deleted? can you re-add this file pls

noot (commented):
the `execute` call looks good! please re-add the test file (i assume that was a mistake lol)

protocol/xmrmaker/claim_test.go
stubbrn: https://github.com/stubbrn/atomic-swap/pull/2

dimalinux: Yeah, the file was just moved, not removed. It moved over cleanly, because it had zero interactions with the `xmrmaker` code base that it was located in. Git isn't recognizing it as a move, because the test was renamed and I removed the dead code handling ERC20 tokens, since we don't support relaying ERC20 swaps.

noot (approved):

Issue #127, opened 2022-06-07
ui: make swapd ports configurable


noot:
make swapd http and ws ports configurable instead of hard-coded

dimalinux:
Can clarify the issue a little and the desired behavior? It looks like the ports are configurable already:
```
atomic-swap$ ./swapd --help | grep port
   --rpc-port value                port for the daemon RPC server to run on; default 5001 (default: 0)
   --ws-port value                 port for the daemon RPC websockets server to run on; default 8080 (default: 0)
   --libp2p-port value             libp2p port to listen on (default: 0)
```
I can see above that there is an issue with how the values are defaulted. Setting the default value correctly in the argument parsing code to the stated (above) value is a trivial change, but it looks like the default value is actually dependent on other flags:
```
	p = uint16(c.Uint(flagRPCPort))
	switch {
	case p != 0:
		rpcPort = p
	case devXMRTaker:
		rpcPort = defaultXMRTakerRPCPort
	case devXMRMaker:
		rpcPort = defaultXMRMakerRPCPort
	default:
		rpcPort = defaultRPCPort
	}

	wsPort := uint16(c.Uint(flagWSPort))
	switch {
	case wsPort != 0:
	case devXMRTaker:
		wsPort = defaultXMRTakerWSPort
	case devXMRMaker:
		wsPort = defaultXMRMakerWSPort
	default:
		wsPort = defaultWSPort
	}
```

noot:
hey @dimalinux , the issue is referencing the port for the UI only. currently it's hard-coded to send RPC calls to `http://localhost:5001`, see here: https://github.com/noot/atomic-swap/blob/master/ui/src/utils/rpcApi.ts#L12 

the intended behaviour is to have the UI select what port to use based on some env variable, falling back to 5001 if it's not set.

PR #396, opened 2023-04-19
repair ui

stubbrn:
That's a start


stubbrn:
From `stagenet.md`:

> Note: the swapd process directly interacts with an unlocked Monero wallet and Ethereum private key. This is to allow for a smoother swap process that doesn't require any interaction from you once initiated. However, this effectively gives swapd access to all your (testnet) funds. In the future, there will be a mode that does not access your keys/wallet, but will require user interaction during a swap.

However it seems like the front-end was working that way (i.e using metamask) at first, it was broken partly because of this, so tell me if what I did (Using `net_takeOfferAndSubscribe` instead of `signer_subscribe`) is adequate.
It's now supposed to be in a state of "barely working" again.

noot:
@stubbrn thanks for this! will review shortly :)

ui/src/components/TakeDealDialog.svelte
```diff
@@ -17,7 +17,6 @@
   const WS_ADDRESS = 'ws://127.0.0.1:5001/ws'
```
noot, line 17: can you change this to `ws://127.0.0.1:${SWAPD_PORT}/ws` where `SWAPD_PORT` is an environment variable? there's also a function for this `readSwapdPortNumber()` in utils/rpcApi.ts you may be able to use

noot: yeah it is! I think this was the only place it was still hard-coded, so this should fix #127 also

noot:
overall, I think this is good as it un-breaks the UI, but I think it should still be using `signer_subscribe`, as it's a bit weird right now that it connects to metamask but doesn't sign anything with it. Is `signer_subscribe` completely broken? I could help with debugging there if that'd be helpful for you

dimalinux (approved):

PR #397, opened 2023-04-19
feat: add bootnode binary

noot:
Example usage:
```
make build-all
./bin/bootnode --env stagenet
```

To get bootnode addresses:
```
./bin/swapcli addresses
```

This PR adds a `bootnode` binary that only runs the network service, specifically the DHT. It can be connected to and used for new peers for discovery.  It does not run any other swap services, except for a minimal RPC server that supports the `daemon` namespace and `net_addresses` and `net_peers`. On the network level it does not support any protocols other than the DHT protocol (ie query, initiate, etc).


cliutil/log.go
```diff
@@ -0,0 +1,59 @@
...
+	case levelError, levelWarn, levelInfo, levelDebug:
+	default:
+		return fmt.Errorf("invalid log level %q", level)
+	}
+
+	setLogLevels(level)
+	return nil
+}
+
+func setLogLevels(level string) {
```
dimalinux, line 38: Maybe now would be an ideal time to move this function to somewhere like `common` so we don't have to synchronize the entries in so many different places?

noot: where else is this used? i thought it was just swapd and now bootnode

dimalinux: 4 places: Here, `swapd`, `swaptester` and `daemon/swap_daemon_test.go`.

cmd/bootnode/main.go
```diff
@@ -0,0 +1,210 @@
...
+		Usage:                "A bootnode for the atomic swap p2p network.",
+		Version:              cliutil.GetVersion(),
+		Action:               runBootnode,
+		EnableBashCompletion: true,
+		Suggest:              true,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  flagDataDir,
+				Usage: "Path to store swap artifacts",
+				Value: "{HOME}/.atomicswap/{ENV}", // For --help only, actual default replaces variables
```
dimalinux, line 53: I'm looking at `docs/default_file_locations.md` and I forgot to update the relayer section when I removed the separate relayer binary. Let's repurpose that section for the bootnode.

Should we put the dedicated bootnode's files in `{home}/.atomicbootnode/{ENV}` or something similar? Our failure modes when 2 programs try to access the same DB at the same time could be cleaner. The fewer defaults that testers have to override to run a swapd and bootnode instance on the same host the better.

I'm wondering if we should use a different default port as well? Or have a default bootnode port, but use an OS assigned random port for clients by default instead of a fixed port.

cmd/bootnode/main.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	libp2pPort := uint16(c.Uint(flagLibp2pPort))
+
+	hostListenIP := "0.0.0.0"
+	if config.Env == common.Development {
+		hostListenIP = "127.0.0.1"
+	}
+
+	host, err := net.NewHost(&net.Config{
```
dimalinux, line 136: I know it will be a super tiny package, but can we separate out all the code that is not dealing with CLI or environment variables and push it into a `bootnode` package. I will definitely be writing unit tests that use the dedicated bootnode.

cmd/bootnode/main.go
noot: right, yeah i was thinking of putting it in `.atomicswap/bootnode`, can update that

a different libp2p port would be fine, but I don't think it matters hugely since a user can just change it on the command line, what do you think?

docs/bootnode.md
```diff
@@ -0,0 +1,25 @@
+# Bootnode
+
+The swap program uses a p2p network to discover offer-makers to do a swap with, and also the run the actual swap protocol. A node must know addresses of nodes already in the network to join the network. These nodes are often publicly posted and referred to as bootnodes. Bootnodes act as an entry-point into the p2p network.
```
dimalinux, line 3: I think the phrase "The swapd daemon" would be more obvious than "The swap program".

How are you editing this? Most people disable line-wrap in code editors and editing a 324 symbol line is unwieldy. Github will still free-flow the lines when viewed in the web browser if you add newlines. I'm been using an editor plugin that allows me to reformat documentation and code comments at 80 characters (a little narrower than our max line length for easier readability).

docs/bootnode.md
```diff
@@ -0,0 +1,25 @@
+# Bootnode
+
+The swap program uses a p2p network to discover offer-makers to do a swap with, and also the run the actual swap protocol. A node must know addresses of nodes already in the network to join the network. These nodes are often publicly posted and referred to as bootnodes. Bootnodes act as an entry-point into the p2p network. 
+
+This repo comes with a `bootnode` program that runs only the p2p components of a swap node, and thus can be used as a lightweight bootnode. 
+
+## Requirements
+- go 1.20+ (see [build instructions](./build.md) to download Go.)
```
dimalinux, line 8: Maybe we should say "See ... " for install help and minimum supported version?  It is something we have to update twice per year, so I'm trying to brainstorm ways that we could minimize the number of places that require updating. Obviously, `grep ^go go.mod` is the authoritative source for the minimum version.

docs/bootnode.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Build and run
+
+To build and run the bootnode binary:
+```bash
+make build-all
+./bin/bootnode --env <environment>
```
dimalinux, line 15: Maybe we should use manpage style:  `--env ENVIRONMENT`.  All caps for a user specified value. Square brackets are used if the syntax is optional.  Angle brackets (`<>`) have special meaning in Linux shells.

net/host.go
```diff
@@ -80,17 +81,19 @@ type Config struct {
...
 func NewHost(cfg *Config) (*Host, error) {
 	h := &Host{
-		ctx:       cfg.Ctx,
-		h:         nil, // set below
-		isRelayer: cfg.IsRelayer,
-		swaps:     make(map[types.Hash]*swap),
+		ctx:        cfg.Ctx,
+		h:          nil, // set below
+		isRelayer:  cfg.IsRelayer,
+		isBootnode: cfg.IsBootnode,
```
dimalinux, line 95: This is a super short function. We could remove `IsBootnode` from this config, add a `NewBootnodeOnlyHost` method, and then `panic` if `isRelayer` is set in the new bootnode-only method.

A regular host can be a bootnode without setting `IsBootnode`, but a bootnode-only host cannot be a relayer. We want to make this as self-documenting as possible while scanning the least amount of code. Right now it is a little confusing to know how to set the values correctly and what they actually mean.

rpc/net.go
```diff
@@ -72,6 +74,10 @@ func (s *NetService) Peers(_ *http.Request, _ *interface{}, resp *rpctypes.Peers
 
 // QueryAll discovers peers who provide a certain coin and queries all of them for their current offers.
 func (s *NetService) QueryAll(_ *http.Request, req *rpctypes.QueryAllRequest, resp *rpctypes.QueryAllResponse) error {
+	if s.isBootnode {
+		return errUnsupportedForBootnode
+	}
```
dimalinux, line 79: Would it be worth leaving this function in for troubleshooting? It don't see any dependencies on a local maker or taker.

rpc/net.go
```diff
@@ -108,6 +114,10 @@ func (s *NetService) discover(req *rpctypes.DiscoverRequest) ([]peer.ID, error)
 
 // Discover discovers peers over the network that provide a certain coin up for `SearchTime` duration of time.
 func (s *NetService) Discover(_ *http.Request, req *rpctypes.DiscoverRequest, resp *rpctypes.DiscoverResponse) error {
+	if s.isBootnode {
+		return errUnsupportedForBootnode
+	}
```
dimalinux, line 119: Same here. This seems like something a bootnode could handle and might help with troubleshooting.

rpc/server.go
```diff
@@ -50,46 +58,70 @@ type Config struct {
 	XMRMaker        XMRMaker
 	ProtocolBackend ProtocolBackend
 	RecoveryDB      RecoveryDB
+	Namespaces      map[string]struct{}
```
dimalinux, line 61: I think this would be more obvious if it was:
```go
EnabledNamespaces map[string]bool
```

rpc/net.go
```diff
@@ -126,8 +136,14 @@ func (s *NetService) Discover(_ *http.Request, req *rpctypes.DiscoverRequest, re
...
-func (s *NetService) QueryPeer(_ *http.Request, req *rpctypes.QueryPeerRequest,
-	resp *rpctypes.QueryPeerResponse) error {
+func (s *NetService) QueryPeer(
+	_ *http.Request,
+	req *rpctypes.QueryPeerRequest,
+	resp *rpctypes.QueryPeerResponse,
+) error {
+	if s.isBootnode {
+		return errUnsupportedForBootnode
+	}
```
dimalinux, line 146: This one too doesn't require a maker or a taker subsystem locally.

net/host.go
noot: maybe I can change the variable name to `IsBootnodeOnly` or something like that?

rpc/net.go
noot: the main issue is I didn't want the bootnode to support any stream protocols since they aren't necessary to its function, I could add the query protocol in if you think it's useful

rpc/server.go
noot: I think that's more complicated b/c the value can be true, false, or not set

rpc/net.go
noot: same comment as queryAll above

net/host.go
dimalinux: Yes, that was part of the suggestion. Changing the variable name eliminates most of the confusion. The rest of the suggestion wouldn't be much extra code, especially when you consider that the existing `NewHost` and `Config` struct no longer need updates, but those changes are less important.

rpc/net.go
dimalinux: I'm imagining scenarios where a swapd node can't connect to a bootnode, or 2 peers don't see each other's offers, or a swapd instance is not finding any relayers despite relayers existing. I can at least imagine scenarios where being able to query the state of the DHT on the bootnode and maybe even reachability of peers could aid in diagnosing where/how the problem is happening. We can always ticket it and add it later if you think it will be useful but don't want to take it on now.

rpc/server.go
dimalinux: It is less complicated because you no longer have to check if the value was set. You assign a single value from the map instead of two: https://go.dev/play/p/z7_qyFbzkJ0

noot: hmm, i suppose. i always thought this way was more idiomatic for if something is either set or not set

dimalinux: You are thinking that because, at least with earlier versions of go (and maybe even now), if you are managing enormous sets, you can get a very tiny performance increase using `struct{}`. In our use case, it is just a small map that isn't even accessed much. The performance win wouldn't even be measurable.

net/host.go
noot: I changed the variable name and added a check that `IsRelayer` and `IsBootnodeOnly` are exclusive!

rpc/net.go
noot: I think discover should be fine for now since that shows who's discoverable on the DHT, we can add query in later if necessary

rpc/server.go
dimalinux: The comment wasn't to change the type in isolation.  It was that the code would be more self-documenting if the map name was changed to `EnabledNamespaces`.  Then code that does a lookup in the map is asking a question: "Is this namespace enabled?" The answer to that question is a boolean, which lead to the type change suggestion. The looping code you are referring to can have its inner switch removed and just check if the namespaces it cares about are enabled.

dimalinux (approved):

PR #398, opened 2023-04-22
fix: a few more contract checks

noot:
- add more checks in `newSwap` for zero values for hashed pubkeys, timeouts, and the claimer address
- clean up contract comments a bit


ethereum/contracts/SwapCreator.sol
```diff
@@ -18,17 +18,18 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
     }
 
     struct Swap {
-        // individual swap creator, Alice
+        // the swap initiator, Alice
+        // Alice is the only one who can call `refund`
         address payable owner;
-        // address allowed to claim the ether in this contract
+        // address allowed to claim the ether for this swap, Bob
```
dimalinux, line 24: Saying that Bob is an address would sound okay if Alice was described the same way, but we said she is "the swap initiator".  Here's a starter idea on how synchronize description styles:
```
the swap offerer (maker?), who can claim the ether of this swap, Bob
```

ethereum/contracts/SwapCreator.sol
```diff
@@ -160,11 +174,11 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
     }
 
     // Bob can claim if:
-    // - Alice has set the swap to `ready` or it's past t_0 but before t_1
+    // - (Alice has set the swap to `ready` or it's past timeout0) and it's before timeout1
```
dimalinux, line 177: Did you mean to place the left parenthesis before the word `or`?

ethereum/contracts/SwapCreator.sol
```diff
@@ -160,11 +174,11 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
...
     function claim(Swap memory _swap, bytes32 _s) public {
         _claim(_swap, _s);
 
-        // send eth to caller (Bob)
+        // send ether to swap claimant
```
dimalinux, line 181: "Claimer" already is a noun. We probably shouldn't introduce a new noun. "Claimant" is usually only used in legal contexts.

I've actually been standardizing on ETH (all caps) in the Go code. When looking at modern documentation, it seems like ETH is used more often. Not that there is anything wrong with "ether". It is not why I wrote this comment.

ethereum/contracts/SwapCreator.sol
```diff
@@ -242,7 +258,7 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
         }
     }
 
-    function verifySecret(bytes32 _s, bytes32 pubKey) internal pure {
-        if (!mulVerify(uint256(_s), uint256(pubKey))) revert InvalidSecret();
+    function verifySecret(bytes32 _s, bytes32 _hashedPubkey) internal pure {
```
dimalinux, line 261: Would it be worth converting the documentation in this contract to NatSpec and ensuring that all functions and their parameters are documented?
https://docs.soliditylang.org/en/v0.8.19/style-guide.html#natspec

ethereum/swap_creator_test.go
```diff
@@ -31,6 +31,7 @@ var (
 	defaultTimeoutDuration = big.NewInt(60) // 60 seconds
 	ethAssetAddress        = ethcommon.Address(types.EthAssetETH)
 	defaultSwapValue       = big.NewInt(100)
+	fakeSwapKey            = [32]byte{1} // dummy non-zero value for claim/refund key
```
dimalinux, line 34: I like the name in the comment better than the one on the variable.  "dummy" vs "fake".

dimalinux (approved):
Please read and maybe act on the minor comments before merging, but it looks good.

ethereum/contracts/SwapCreator.sol
noot: no, i wasn't sure how to make this non-ambiguous in natural language lol

ethereum/contracts/SwapCreator.sol
noot: I updated the comment for Alice to match the one for Bob, I think that's clearer

ethereum/contracts/SwapCreator.sol
noot: yeah probably, can do in a follow up

PR #392, opened 2023-04-19
created test for issue385

dimalinux:
I wasn't able to replicate issue385 in a unit test, but I am seeing what looks like the same issue when cancelling Alice's server before she completes the transfer from the temporary swap wallet back to her primary wallet. In the included test, she's not recovering successfully on server restart, although, in theory, she has the necessary pieces of information to complete the swap on server restart.

This branch is only for playing with the issue and not intended for merging without significant changes.

PR #399, opened 2023-04-22
fix: add `SweepingXMR` status, fixes restart during sweep

### LINKED ISSUE #374: Separate Sweep event/status for XMR taker
Right now the maker appears to finish well before the taker.  It would be nice to show the taker status during the sweep stage.

### LINKED ISSUE #385: XMR Taker showing wrong status of past swap
I just had a stagenet token swap. Both sides completed successfully. Server was restarted before querying the past swaps. (That may or may not be relevant, I can't go back and time and query the past swap before restarting.) The Maker is showing that the swap completed successfully when running `swapcli past`, but the taker is showing that the status is `ContractReady`:

Proof that taker side completed successfully:
```
$ ./bin/swapcli take --peer-id 12D3KooWNseb7Ei8Xx1aBKjSFoZ9PGfdxN9MwQxfSRxsBAyA8op4 --offer-id 0x91d8a21a6a7682def6507d4e70d37e12cdfa991d9e7a777077bc815a9708db94 --provides-amount 15 --swapd-port 5001
Initiated swap with offer ID 0x91d8a21a6a7682def6507d4e70d37e12cdfa991d9e7a777077bc815a9708db94
2023-04-16-01:05:44 > Stage updated: ExpectingKeys
2023-04-16-01:06:13 > Stage updated: ETHLocked
2023-04-16-01:33:37 > Stage updated: ContractReady
2023-04-16-01:40:26 > Stage updated: Success
```
Here's the `swapcli past` swap query against the same taker server (after restart). Note that the End Time is also missing:
```
$ ./bin/swapcli past --offer-id 0x91d8a21a6a7682def6507d4e70d37e12cdfa991d9e7a777077bc815a9708db94 --swapd-port 5001
Past swaps:
ID: 0x91d8a21a6a7682def6507d4e70d37e12cdfa991d9e7a777077bc815a9708db94
Start time: 2023-04-16-01:05:43
End time: -
Provided: 15 "NOOT"
Received: 10 XMR
Exchange Rate: 1.5 ETH/XMR
Status: ContractReady
```
This test was in the `dimalinux/erc20-balance` development branch, but I do not think the incorrect status could be related to branch changes.

noot:
- as title says
- if we see the swap's status is `SweepingXMR` when restarting, we set the swap's status to completed (success if xmrtaker, refunded if xmrmaker). we don't currently store the sweep tx hash anywhere so we can't check it, I could add storing it in the db and checking it on restart if that seems useful

closes #374
closes #385
closes #392


protocol/claim_monero.go
```diff
@@ -20,6 +21,11 @@ var (
 	log = logging.Logger("protocol")
 )
 
+// SwapManager is the interface for the swap manager.
```
dimalinux, line 24: I'm not sure the comment says anything that isn't obvious from the name. Would this phrase be correct?
```
// SwapManager is the subset of the swap.Manager interface needed by ClaimMonero
```

protocol/claim_monero.go
```diff
@@ -87,3 +99,18 @@ func ClaimMonero(
...
+
+// setSweepStatus sets the swap's status as `SweepingXMR` and writes it to the db.
+func setSweepStatus(info *swap.Info, sm SwapManager) error {
+	info.SetStatus(types.SweepingXMR)
+	err := sm.WriteSwapToDB(info)
+	if err != nil {
+		return err
+	}
+
+	if info.StatusCh() != nil {
```
dimalinux, line 111: Are there code paths where this can be `nil` and everything is still okay? I'm wondering if there needs to be a logged message.

daemon/sweep_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/athanorlabs/atomic-swap/rpcclient/wsclient"
+	"github.com/athanorlabs/atomic-swap/tests"
+)
+
+// This test starts a swap between Bob and Alice. Bob completes the swap
+// successfully, and then both swapd daemons are shut down before Alice
+// completes the sweep from the swap wallet back to her primary wallet.
+// The Alice restarts, she should not try to sweep again and instead know
+// that the swap is already complete.
+func TestIssue385(t *testing.T) {
```
dimalinux, line 26: Then Alice restarts ...

Maybe call this: `TestAliceStoppedAndRestartedDuringXMRSweep`

dimalinux (approved):
I left feedback, so don't merge it immediately, but it looks good!

protocol/claim_monero.go
noot: it can be nil, for example if this is a restart of an ongoing swap, the `statusCh` will be nil

PR #400, opened 2023-04-22
release: add initial mainnet bootnodes

noot:
add a few bootnodes so that more bootnodes can join 

#326


common/config.go
```diff
@@ -70,7 +70,14 @@ func MainnetConfig() *Config {
 		SwapCreatorAddr: ethcommon.HexToAddress("0x"), // TODO
 		// ForwarderAddr is from https://docs.opengsn.org/networks/addresses.html
 		ForwarderAddr: ethcommon.HexToAddress("0xB2b5841DBeF766d4b521221732F9B618fCf34A87"),
-		Bootnodes:     []string{}, // TODO
+		Bootnodes: []string{
+			"/ip4/67.205.131.11/tcp/9909/p2p/12D3KooWGpCLC4y42rf6aR3cguVFJAruzFXT6mUEyp7C32jTsyJd",
```
dimalinux, line 74: I thought about this a lot when I was working on the p2p code and also looked at what other projects were doing with their bootnodes:

Since we support both TCP and UDP/QUIC, and we know that all clients will support TCP, but there may be a few that are firewall blocked from using UDP, we want the bootnode entries to be the TCP version. As soon as the client connects to the bootnode over TCP, it will be informed that UDP/QUIC is supported and it will prefer UDP/QUIC for future interaction. There is no value in adding the same node twice.

noot: okay, I wasn't sure about this either, will change to TCP

PR #404, opened 2023-04-23
fetch forwarder address from swap contract

### LINKED ISSUE #365: remove hardcoded forwarder address for network config; load from swap contract
as title says, we can remove the `ForwarderContractAddress` from the network configs and just load it from the contract at `SwapFactoryAddress`.

stubbrn:
closes #365

What should I do for mainnet? `SwapCreatorAddr` is "0x" and `ForwarderAddr` is used in `CheckForwarderContractCode`

https://github.com/AthanorLabs/atomic-swap/blob/ea8bae1416c2c4fad8a5a30cb03118675a52994e/common/config.go#L70-L72
https://github.com/AthanorLabs/atomic-swap/blob/ea8bae1416c2c4fad8a5a30cb03118675a52994e/ethereum/check_swap_creator_contract.go#L116


ethereum/check_swap_creator_contract_test.go
```diff
@@ -141,7 +141,8 @@ func TestSepoliaContract(t *testing.T) {
 	if errors.Is(err, errInvalidSwapCreatorContract) && sepoliaKey != "" {
 		pk, err := ethcrypto.HexToECDSA(sepoliaKey) //nolint:govet // shadow declaration of err
 		require.NoError(t, err)
-		forwarderAddr := common.StagenetConfig().ForwarderAddr
+		forwarderAddr, err := DeployGSNForwarderWithKey(context.Background(), ec, pk)
+		require.NoError(t, err)
```
stubbrn, line 145: This deploys a forwarder contract instead of using the hardcoded `ForwarderAddr`

dimalinux: See the comments I added to the ticket. Deploying contracts on `mainnet` is expensive and the forwarder contract is already deployed by its creators, so it probably doesn't make sense to ever deploy it except in ganache. It is cheap to deploy it on sepolia, but we should treat `stagenet` as close as possible to the way we treat `mainnet` and not plan to redeploy it ever again on sepolia, so that code can be removed. It was just convenience code code for us and only ever executed if you were running the unit test by hand with a key set.

noot:
since the mainnet contract isn't deployed yet, we should probably merge this after that's done (which should be very soon!)

dimalinux:
Thanks for the work and getting us to figure out the small details for the complete fix! I'm closing this PR, as https://github.com/AthanorLabs/atomic-swap/pull/427 ending up fixing the issue.

PR #405, opened 2023-04-23
dockerized bootnode

dimalinux:
Adds Dockerfile and scripts to dockerize the bootnode. Updates to swapd's docker setup were also done to harmonize the behavior of both images.


noot (approved):
works for me :D

PR #406, opened 2023-04-23
Modified struct initialization syntax

drank40:
This syntax is more concise and saves a bit on gas


noot (approved):
saves around 328 gas, thanks!

PR #423, opened 2023-04-25
fix xmrtaker key overwrite on restart

noot:
as title says, `generateAndSetKeys` should NOT have been called when restarting an ongoing swap state, as it overwrote the keys in the db. this would have only been an issue on xmrtaker double-restart during a swap.

added a unit test for this case, and also added a small fix for a case caught in the test in `xmrtaker.swapState.ready()`

### DIFF
--- a/daemon/double_restart_test.go
+++ b/daemon/double_restart_test.go
@@ -18,10 +18,10 @@ import (
 	"github.com/athanorlabs/atomic-swap/tests"
 )
 
-// This test starts a swap between Bob and Alice. The nodes restart *twice* after the contract
-// is ready but before Bob claims. Then, on second restart, Bob should have claimed and
+// This test starts a swap between Bob and Alice. The nodes restart *twice* after the xmr is locked
+// but before Bob claims. Then, on second restart, Bob should have claimed and
 // Alice should be able to get the XMR.
-func TestAliceDoubleRestartAfterContractReady(t *testing.T) {
+func TestAliceDoubleRestartAfterXMRLock(t *testing.T) {
 	minXMR := coins.StrToDecimal("1")
 	maxXMR := minXMR
 	exRate := coins.StrToExchangeRate("300")
@@ -94,12 +94,6 @@ func TestAliceDoubleRestartAfterContractReady(t *testing.T) {
 				if !status.IsOngoing() {
 					return
 				}
-
-				// if status == types.ETHLocked {
-				// 	cancel()
-				// 	t.Log("cancelling context of Alice's and Bob's servers")
-				// 	return
-				// }
 			case <-ctx.Done():
 				t.Logf("Alice's context cancelled before she completed the swap [expected]")
 				return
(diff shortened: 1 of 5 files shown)


daemon/double_restart_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/athanorlabs/atomic-swap/common/types"
+	"github.com/athanorlabs/atomic-swap/monero"
+	"github.com/athanorlabs/atomic-swap/rpcclient"
+	"github.com/athanorlabs/atomic-swap/rpcclient/wsclient"
+	"github.com/athanorlabs/atomic-swap/tests"
+)
+
+// This test starts a swap between Bob and Alice. The nodes restart *twice* after the xmr is locked
+// but before Bob claims. Then, on second restart, Bob should have claimed and
+// Alice should be able to get the XMR.
```
dimalinux, line 23: The next person maintaining this test will probably be wondering, "why 2?". You had a very specific reason, but I don't think I understood it well enough to describe it in the comment here.

protocol/xmrtaker/swap_state.go
```diff
@@ -378,7 +377,7 @@ func (s *swapState) exit() error {
 		s.clearNextExpectedEvent(types.CompletedAbort)
 		return nil
 	case EventXMRLockedType, EventETHClaimedType:
-		// for EventXMRLocked, we already deployed the contract,
+		// for EventXMRLocked, we already lock our ETH-asset,
```
dimalinux, line 380: locked

protocol/xmrtaker/swap_state.go
```diff
@@ -636,6 +635,22 @@ func (s *swapState) ready() error {
...
+		}
+
+		if stage == contracts.StageCompleted {
+			log.Infof("contract aleady set to completed, ignoring call to ready() and sending EventExit")
+			go func() {
+				err = s.Exit()
+				if err != nil {
+					log.Errorf("failed to handle EventExit: %s", err)
+				}
+			}()
```
dimalinux, line 650: Same question as the other PR. What is the background goroutine for? Exit takes a long time, fair locking, something else? Is the code safe if the entire go routine completes before we return from the outer function here?

dimalinux (approved):
Left feedback that should be looked at before merging.

daemon/double_restart_test.go
noot: yeah I'll add more info to the comment

protocol/xmrtaker/swap_state.go
noot: it's because this code is called from inside the event handler, so it causes a deadlock if we try to call `Exit()` as it'll try to put something in to the event channel, but the code path we're in is on a read from the event channel. the other solution would be to make the event channel buffered, as it only has capacity 1 right now, but I'm not sure how ideal that is. can try in another PR

dimalinux: In a follow-up PR, if we don't make it buffered, we should add code comments describing what you wrote above. That is subtle if you have to figure it out just by looking at the code.

PR #425, opened 2023-04-25
fix: set `GasLimit` for relayed transactions

### LINKED ISSUE #343: methods to determine sufficient ETH balance for various claim types
We need methods to determine if there is a sufficient ETH balance to make a claim as well as to determine if there is sufficient balance to make a claim as a relayer.

In the first case, if Bob as has insufficient ETH as determined by this new method, he should use a relayer.

In the second case, any relayer should stop advertising relaying as a service until their balance is replenished to be sufficient to make a claim as a relayer.

### LINKED ISSUE #422: relayer tx simulation call does not work on mainnet
See the following logs:
```
2023-04-25T19:06:44.045Z	DEBUG	relayer	relayer/submit_transaction.go:81	relaying tx with gas price 47898010734
2023-04-25T19:06:44.091Z	ERROR	relayer	relayer/submit_transaction.go:93	failed to simulate execute: err: insufficient funds for gas * price + value: address 0x9017f022AE58452949f092363505B0196142eC3b have 48333473886691647 want 26343905903700000000 (supplied gas 550000000)
2023-04-25T19:06:44.091Z	DEBUG	net	net/relay.go:82	did not handle relay request: err: insufficient funds for gas * price + value: address 0x9017f022AE58452949f092363505B0196142eC3b have 48333473886691647 want 26343905903700000000 (supplied gas 550000000)
```

I suspect it's due to the `txOpts.GasLimit` not being set (notice the log above is setting the gas limit to 550000000) :/

noot:
closes #422
closes #343 (should re-open another issue regarding second case) I can pull this into another PR if desired, I added it here since I needed it for testing

also fix setting of `nextExpectedEvent` to `XMRLocked` in `xmrmaker`, a restart during the transfer step caused the node to think nothing was locked on restart.


protocol/xmrmaker/claim.go
```diff
@@ -81,6 +88,40 @@ func (s *swapState) claimFunds() (*ethtypes.Receipt, error) {
 	return receipt, nil
 }
 
+// checkForMinClaimBalance check if we have enough balance to call claim.
+// return true if we do, false otherwise.
+func checkForMinClaimBalance(ctx context.Context, ec extethclient.EthClient) (bool, error) {
```
dimalinux, line 93: I was about to suggest that this function belongs in the `relayer` package, but then I just looked at the `relayer` package to see where it would go, and there already is a function a named `checkForMinClaimBalance` in it.

protocol/xmrtaker/claim.go
```diff
@@ -117,5 +117,11 @@ func (s *swapState) claimMonero(skB *mcrypto.PrivateSpendKey) (*mcrypto.Address,
 
 	close(s.claimedCh)
 	log.Infof("monero claimed and swept to original account %s", depositAddr)
+	go func() {
+		err = s.Exit()
+		if err != nil {
+			log.Warnf("failed to exit: %v", err)
+		}
+	}()
```
dimalinux, line 125: Why are we using a background goroutine here? I'm asking, because I also want to understand if the caller of this method is safe if the goroutine completely executes before we return.

relayer/submit_transaction.go
```diff
@@ -101,6 +103,7 @@ func ValidateAndSendTransaction(
 		req.Signature,
 	)
 	if err != nil {
+		log.Errorf("failed to call execute: %s", err)
```
dimalinux, line 106: The message seems a little vague. Maybe something like `failed to simulate forwarder call`? It says this line wasn't unit tested, but we have a unit test, `Test_ValidateAndSendTransaction`, that is supposed to trigger a simulation error. It tries to claim twice to trigger it. I don't have the branch checked out right now, but I could look into it if you don't have time.

protocol/xmrmaker/claim.go
noot: the one in `relayer` is specifically for claims that are relayed through the forwarder, it also has different return values and returns and error instead of true/false if there's not enough funds to relay. I don't think it makes sense to use that function here

protocol/xmrtaker/claim.go
noot: it's impossible for the goroutine to exit before we return. this code path is executed synchronously on a read from the eventCh, and `Exit()` puts something in the eventCh, which has capacity 1. so if this isn't in a goroutine, the code will deadlock

relayer/submit_transaction.go
noot: this isn't the simulation call, this is the actual call to the contract

dimalinux (approved):

PR #427, opened 2023-04-26
move to gsnforwarder that is bytecode identical on mainnet and dev

dimalinux:
The `go-relayer` project has been updated to have bindings for a gsnforwarder that is bytecode identical to the mainnet forwarder, deployed by opengsn (which we are using). The forwarder we were previously using in dev/stagenet was compiled with a different version of solidity and had subtle source code differences from the one deployed on mainnet.

The gsnforwarder changes in `go-relayer` will be merged and tagged first. Reviewing those changes is part of reviewing this PR:
https://github.com/AthanorLabs/go-relayer/pull/13

Since I would have needed to update the environment configs anyways, I removed their references in the config:
closes https://github.com/AthanorLabs/atomic-swap/issues/365

@stubbrn Apologies for stealing a PR out from under you. Since you already started working on the issue, your review on these code changes would be helpful/appreciated.


go.mod
```diff
@@ -9,11 +9,11 @@ require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/athanorlabs/go-dleq v0.1.0
 	github.com/athanorlabs/go-p2p-net v0.2.0
-	github.com/athanorlabs/go-relayer v0.1.0
+	github.com/athanorlabs/go-relayer v0.1.1-0.20230426015202-a005fed78fda
```
dimalinux, line 12: Will be updated to tagged release after https://github.com/AthanorLabs/go-relayer/pull/13 is merged.

common/config.go
```diff
@@ -69,7 +68,6 @@ func MainnetConfig() *Config {
 		},
 		SwapCreatorAddr: ethcommon.HexToAddress("0xD3d19539D61bB0e7617E499C7262594E71CA1c66"),
 		// ForwarderAddr is from https://docs.opengsn.org/networks/addresses.html
-		ForwarderAddr: ethcommon.HexToAddress("0xB2b5841DBeF766d4b521221732F9B618fCf34A87"),
```
noot, line 72: I'm kind of unsure now about removing this hardcoded value, it's useful for needing to redeploy, and I also think we can and should skip the code check if the addresses match

noot: I think we should still document the contract addresses somewhere if we remove it from here

dimalinux: We still have the constants, they are just in the repo with the code that creates and validates the forwarder. Moving the tiny fraction of `go-relayer` that we use into this repo, so we don't need to maintain the large amount of unused/not-so-tested code in a secondary repo, would be a code maintenance win in my opinion though. Do you want me to add back the forwarder addresses here as code comments?

dimalinux: Code comment updated.

PR #435, opened 2023-04-27
multi-arch build script and version fixes

dimalinux:


cliutil/utils.go
```diff
@@ -99,26 +99,42 @@ func GetVersion() string {
 		return "unknown-version"
 	}
 
-	commitHash := "???????"
```
dimalinux, line 102: We didn't have any release tags when I wrote this and I didn't know that the commit hash wouldn't be present if doing "go install" and passing the remote repository.

noot (approved):
works for me! can't test the non x64 ones, but the script looks fine

PR #441, opened 2023-04-27
feat: swapcli commands to manually claim/refund and for database functions

### LINKED ISSUE #424: swapcli database functionality
we should add the rpc `database` namespace functionality to swapcli.

### LINKED ISSUE #439: swapcli commands to manually claim/refund based on offer ID
this would be nice as it's a hassle to manually format the calldata for these funcs as they take both the swap struct + the swap secret. a swapcli command that takes the offer ID could be used to manually call these functions, in case of bad endpoints being used or other unforeseen errors.

noot:
new functions:
```
   claim                    manually call claim() in the contract for a given swap. WARNING: This should only be used if the normal swap process fails.
   refund                   manually call refund() in the contract for a given swap. WARNING: This should only be used if the normal swap process fails.
   get-contract-swap-info   Get information about a swap needed to call the contract functions. Returns the contract address, the swap's struct as represented in the contract, and the hash of the swap's struct, which is used as its contract identifier. Note: this is only useful if you plan to manually call the contract functions.
   get-swap-secret          Get the secret for a swap. WARNING: do NOT share this secret with anyone. Doing so may result in a loss of funds. You should not use this function unless you are sure of what you're doing. This function is only useful if you plan to try to manually recover funds.
```

I'm wondering if these should be gated under an "unsafe" or "dev" command/feature, since they shouldn't be used under normal circumstances, especially the manual claim/refund functions.

closes #439
closes #424

### DIFF
--- a/cmd/swapcli/main.go
+++ b/cmd/swapcli/main.go
@@ -291,34 +291,6 @@ func cliApp() *cli.App {
 				},
 			},
 			{
-				Name: "claim",
-				Usage: "manually call claim() in the contract for a given swap. " +
-					"WARNING: This should only be used if the normal swap process fails.",
-				Action: runClaim,
-				Flags: []cli.Flag{
-					&cli.StringFlag{
-						Name:     flagOfferID,
-						Usage:    "ID of swap for which to call claim()",
-						Required: true,
-					},
-					swapdPortFlag,
-				},
-			},
-			{
-				Name: "refund",
-				Usage: "manually call refund() in the contract for a given swap. " +
-					"WARNING: This should only be used if the normal swap process fails.",
-				Action: runRefund,
-				Flags: []cli.Flag{
-					&cli.StringFlag{
-						Name:     flagOfferID,
-						Usage:    "ID of swap for which to call refund()",
-						Required: true,
-					},
-					swapdPortFlag,
-				},
-			},
-			{
 				Name:   "set-swap-timeout",
 				Usage:  "Set the duration between swap initiation and t0 and t0 and t1, in seconds",
 				Action: runSetSwapTimeout,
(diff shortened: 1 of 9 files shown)


cmd/swapcli/main.go
```diff
@@ -945,6 +1037,50 @@ func runShutdown(ctx *cli.Context) error {
...
+	resp, err := c.GetContractSwapInfo(offerID)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Contract address: %s\n", resp.SwapCreatorAddr)
+	fmt.Printf("Block at which newSwap was called: %d\n", resp.StartNumber)
+	fmt.Printf("Swap ID as stored in the contract: %s\n", resp.SwapID)
+	fmt.Printf("Swap struct as stored in the contract:\n")
+	fmt.Printf("\tOwner: %s\n", resp.Swap.Owner)
```
dimalinux, line 1056: Most terminals have tabs set to 8 space, which might be a little much here. In other `swapcli` output, like `query-all`, we are using 2-space indents.

daemon/manual_tx_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/athanorlabs/atomic-swap/rpcclient/wsclient"
+	"github.com/athanorlabs/atomic-swap/tests"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test if Alice is able to manually call `refund()` via swapcli.
```
dimalinux, line 27: Test if Alice is able to call the `Refund()` RPC API (used by `swapcli refund`) immediately after locking her ETH.

cmd/swapcli/main.go
```diff
@@ -290,6 +290,34 @@ func cliApp() *cli.App {
 					swapdPortFlag,
 				},
 			},
+			{
+				Name: "claim",
+				Usage: "manually call claim() in the contract for a given swap. " +
```
dimalinux, line 295: We might need an FAQ item in our Wiki with an example of a use case of where it was successfully used.  It is calling claim outside the automated protocol, so I guess you could say it is a "manual" claim.

cmd/swapcli/main.go
```diff
@@ -290,6 +290,34 @@ func cliApp() *cli.App {
...
+						Usage:    "ID of swap for which to call claim()",
+						Required: true,
+					},
+					swapdPortFlag,
+				},
+			},
+			{
+				Name: "refund",
+				Usage: "manually call refund() in the contract for a given swap. " +
+					"WARNING: This should only be used if the normal swap process fails.",
```
dimalinux, line 310: I just tested, and you can place newlines in these strings. `swapcli` will correctly indent after the newline.

dimalinux (approved):
Look's good. Some minor comments that you should read before merging.

cmd/swapcli/main.go
noot: yeah we should emphasize these should definitely not be used normally, it's mostly for us when we find bugs testing or in the case a user runs into some unexpected and unrecoverable error

PR #442, opened 2023-04-27
fix: move counterparty-relay checks to backend

### LINKED ISSUE #421: xmrtaker is unable to do a "counterparty-relay" after restart
```
2023-04-25T19:04:20.848Z	DEBUG	net	net/relay.go:74	received invalid taker-specific claim request from peer=12D3KooWKNweo3pR2BSXGpNs5rDa7bG2MoueMhFZ5Hw4uXaTdJ6c offerID=0xabdd2c94ccda33d8a86a1bdcbd5f229eeea29f477c55a2d44ac5291a391b5d13 swap-found=false
```

Scenario: xmrmaker uses xmrtaker to do counterparty-specific relay (ie xmrtaker has relaying set to false). xmrtaker restarts their node at some point during the swap, before xmrmaker tries to send a claim relay request. then, once the xmrmaker sends the claim relay request to the xmrtaker, the xmrtaker doesn't accept its request because it doesn't think (on the network level) that it has an ongoing swap with them, because the `net.Host.swaps` map isn't re-propagated on restart. The log above is what the xmrtaker sees.

Easy solution is to move the counterparty-check to `backend.HandleRelayClaimRequest()` and use the `swapManager` to check 1. we are doing that swap currently 2. we are the taker and 3. the peer ID matches

noot:
closes #421

### DIFF
--- a/protocol/backend/backend.go
+++ b/protocol/backend/backend.go
@@ -15,7 +15,6 @@ import (
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/libp2p/go-libp2p/core/peer"
 
-	"github.com/athanorlabs/atomic-swap/coins"
 	"github.com/athanorlabs/atomic-swap/common"
 	"github.com/athanorlabs/atomic-swap/common/types"
 	mcrypto "github.com/athanorlabs/atomic-swap/crypto/monero"
@@ -253,7 +252,7 @@ func (b *backend) HandleRelayClaimRequest(
 			return nil, err
 		}
 
-		if info.Provides != coins.ProvidesETH {
+		if !info.IsTaker() {
 			return nil, fmt.Errorf("cannot relay taker-specific claim request; not the xmr-taker for swap %s", *request.OfferID)
 		}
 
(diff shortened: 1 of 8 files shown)


protocol/backend/backend.go
```diff
@@ -237,7 +238,31 @@ func (b *backend) ClearXMRDepositAddress(offerID types.Hash) {
...
+		if !has {
+			return nil, fmt.Errorf("cannot relay taker-specific claim request; no ongoing swap for swap %s", *request.OfferID)
+		}
+
+		info, err := b.swapManager.GetOngoingSwap(*request.OfferID)
+		if err != nil {
+			return nil, err
+		}
+
+		if info.Provides != coins.ProvidesETH {
```
dimalinux, line 256: Maybe we need `info.IsTaker()` and `info.IsMaker()` methods that does these checks? (Shorter and more obvious than these checks which are easy confused with eth-asset checks.) If you like the idea, I can implement it in a future PR.

dimalinux (approved):

protocol/backend/backend.go
noot: I added an `info.IsTaker()` method for this

Issue #437, opened 2023-04-27
ensure XMR taker has sufficient funds (including all gas costs)


dimalinux:
We should verify that the XMR taker has enough funds for all of the following before allowing her to take an offer:
Provided amount + NewSwap Gas + SetRead Gas + Refund Gas

Right now we only verify that the taker has enough ETH for the provided amount.

PR #445, opened 2023-04-29
min and max offer amount checks

### LINKED ISSUE #286: Need lower bound on minimum XMR offer
If the XMR amount in a swap is too low, there are insufficient funds to pay the fees to transfer the claimed or refunded funds from the swap wallet back to the primary address. We could avoid support questions if we just have a good error message early when the offer is created, that a min offer amount less than `X` XMR is not allowed.

I'm not sure how to come up with the ideal minimum value. Maybe someone reading this has ideas. My initial idea is 0.001 XMR ($0.20 if XMR is at $200 USD). Average fees appear to be around 0.0002 XMR (note the extra zero).

### LINKED ISSUE #434: confusing over offer limit error message
The current error message looks like this:
```text
0.04567890123456789 ETH provided is over offer maximum of 0.5 XMR
```
The first time the end user reads the message, they are confused, because they expect the left and right side to be in the same currency. The right side should either also be in ETH (probably good enough), or both values should be printed together.

### LINKED ISSUE #438: min swap amount should be greater than the relayer fee
Right now, you can end up in a situation where the maker can not claim, because they don't have sufficient ETH to claim and no relayer will help them because the swap value is insufficient to cover the relayer fee.

dimalinux:
Maker min/max changes:
* Checks applied everywhere offers are used:
  * Max offer amount limited to 1000 XMR (sanity upper bounds check, somewhat arbitrary)
  * Min offer amount limited to the relayer fee converted back to XMR using the exchange rate in the offer.
* Checks applied at time of offer
  * Maker must have an XMR balance strictly greater than the max offer amount (unchanged)
  * Maker must have enough ETH to do a token claim if the offer is for a token swap

Taker min/max validations before offer is taken:
* For ETH
  * Must have a balance that can cover the provides amount plus the gas for NewSwap, SetReady and Refund to take an offer
* For Tokens:
  * Must have enough tokens to cover the provides amount (unchanged)
  * Must have an ETH balance to cover Approve, NewSwap, SetReady and Refund.

Misc changes:
* Fixes error messages on provides amount under-min/over-max values. Both sides of the error message now use ETH.
* Fixes occasional CI timeouts in the `contracts` package because txOpts were being reused
* Provides constants for max gas amounts, assertions that we are under, and logging that shows deltas from the max.

Closes: #437
Closes #438 
Closes #286
Closes #434


coins/common.go
```diff
@@ -20,6 +22,13 @@ const (
 	MaxCoinPrecision = 100
 )
 
+// RelayerFeeWei and RelayerFeeETH are the fixed 0.009 ETH fee for using a swap
+// relayer to claim.
+var (
+	RelayerFeeWei = big.NewInt(9e15)
+	RelayerFeeETH = NewWeiAmount(RelayerFeeWei).AsEther()
```
dimalinux, line 29: We can't have these in the `relayer` package, because we need the value in `common/types` where the offer checks are. We can't move them to the `common` package, as `common/types` depends on `common`. Given the choice of prefixing the values with `types` or `coins`, I thought `coins` was slightly better.

ethereum/check_swap_creator_contract_test.go
```diff
@@ -6,46 +6,24 @@ package contracts
...
-	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/stretchr/testify/require"
 
 	"github.com/athanorlabs/atomic-swap/common"
 	"github.com/athanorlabs/atomic-swap/tests"
 )
 
-// deployContract is a test helper that deploys the SwapCreator contract and returns the
-// deployed address
-func deployContract(
```
dimalinux, line 25: consolidated with another test helper doing the same thing

ethereum/erc20_token_test.go
dimalinux, line 1: This was a rename of `ethereum/erc20_mock_test.go`, but I made enough changes that the diff view thinks it is new.

protocol/xmrmaker/backend_offers.go
```diff
@@ -24,6 +24,12 @@ func (inst *Instance) MakeOffer(
 		return nil, errUnlockedBalanceTooLow{o.MaxAmount, unlockedBalance}
 	}
 
+	// If it is an XMR-for-ETH offer, the min offer amount converted to ETH
+	// must be less than relayer fee
```
dimalinux, line 28: I'll delete this comment.  It is handled.

protocol/xmrmaker/backend_offers.go
```diff
@@ -24,6 +24,12 @@ func (inst *Instance) MakeOffer(
...
+
+	// If it is an XMR-for-TOKEN offer, the maker must have sufficient ETH
+	// to claim
```
dimalinux, line 31: I forgot to add this. I'll take care of it later today.

ethereum/consts.go
```diff
@@ -0,0 +1,15 @@
+package contracts
+
+// Gas prices for our operations. Most of these are set by the highest value we
+// ever see in a test, so you would need to adjust upwards a little to use as a
+// gas limit. We use these values to estimate minimum required balances.
```
dimalinux, line 5: I think the values below have stabilized, but if a test fails when working on a future PR, just increment the constant to the new recorded high.

ethereum/erc20_mock_test.go
dimalinux, line 1: Not deleted. Renamed to `erc20_token_test.go`.

ethereum/erc20_token_test.go
dimalinux, line 1: Renamed from `erc20_mock_test.go`, not a new file.

common/types/offer.go
```diff
@@ -141,6 +148,25 @@ func (o *Offer) validate() error {
...
+		return err
+	}
+	if o.MinAmount.Cmp(relayerFeeAsXMR) <= 0 {
+		return fmt.Errorf(
+			"min amount must be greater than %s ETH when converted (%s XMR)",
+			coins.RelayerFeeETH.Text('f'), relayerFeeAsXMR.Text('f'))
+	}
+
+	if o.MaxAmount.Cmp(maxOfferValue) > 0 {
+		return fmt.Errorf("%s XMR exceed max offer amount of %s XMR",
```
noot, line 166:
```suggestion
		return fmt.Errorf("%s XMR exceeds max offer amount of %s XMR",
```

common/types/offer.go
```diff
@@ -22,6 +22,12 @@ var (
 	// CurOfferVersion is the latest supported version of a serialised Offer struct
 	CurOfferVersion, _ = semver.NewVersion("1.0.0")
 
+	// Don't allow offers over 1000 XMR. Mainly to prevent fat-finger errors, it
+	// could be raised if users need it.
+	maxOfferValue = apd.New(1, 3) // 1000 XMR
```
noot, line 27: any reason for this number?

common/types/offer.go
```diff
@@ -141,6 +148,25 @@ func (o *Offer) validate() error {
...
+	// ETH to make the claim themselves. While relayers are not used with ERC20
+	// swaps, we still want a minimum swap amount, so we use the same value.
+	relayerFeeAsXMR, err := o.ExchangeRate.ToXMR(coins.RelayerFeeETH)
+	if err != nil {
+		return err
+	}
+	if o.MinAmount.Cmp(relayerFeeAsXMR) <= 0 {
+		return fmt.Errorf(
+			"min amount must be greater than %s ETH when converted (%s XMR)",
+			coins.RelayerFeeETH.Text('f'), relayerFeeAsXMR.Text('f'))
```
noot, line 162: maybe suggest the minimum xmr amount allowed in the error message?

protocol/xmrmaker/min_balance.go
```diff
@@ -0,0 +1,66 @@
...
+	ec extethclient.EthClient,
+	offerMaxAmt *apd.Decimal,
+	ethAsset types.EthAsset,
+) error {
+	piconeroBalance, err := mc.GetBalance(0)
+	if err != nil {
+		return err
+	}
+
+	// Maker needs a sufficient XMR balance regardless if it is an ETH or toke swap
```
noot, line 30:
```suggestion
	// Maker needs a sufficient XMR balance regardless if it is an ETH or token swap
```

noot (approved):
looks good, only minor comments!

common/types/offer.go
dimalinux: I was just looking for a semi-round number that's greater than what anyone, at the current time at least, would realistically be swapping. At today's price, 1000 XMR is around $155,410 USD. If you have a different suggestion, let me know and I'll plug it in. I could have just as easily picked 10,000 XMR.

common/types/offer.go
dimalinux: We print both values. The first value in ETH is a constant.  I treated the XMR value as secondary, putting it in parentheses, because it is calculated and depends on the exchange rate in the offer. If you have a suggestion on how to improve the wording, let me know.

PR #446, opened 2023-04-29
feat: close stream after NotifyETHLocked is sent

### LINKED ISSUE #431: nodes should close swap stream after `ETHLocked` message is sent
blocked by #421 

nodes should be allowed to close the swap stream or even disconnect entirely after the `ETHLocked` message is sent from taker->maker, as there are no network messages exchanged after this. currently, if the nodes disconnect at any point during the swap, it triggers an `Exit` event which may not be necessary.

noot:
- as title says
- net package no longer has `swapState.Exit()`, exits are only called from within the protocol packages (or in rpc takeOffer if sending an initiation message fails)

closes #431


net/host.go
```diff
@@ -183,16 +183,25 @@ func (h *Host) CloseProtocolStream(offerID types.Hash) {
...
 	)
+
 	_ = swap.stream.Close()
 }
 
+// DeleteOngoingSwap deletes an ongoing swap from the network's state.
+func (h *Host) DeleteOngoingSwap(offerID types.Hash) {
+	h.swapMu.Lock()
+	defer h.swapMu.Unlock()
+	delete(h.swaps, offerID)
```
dimalinux, line 202: Do we need to be closing the stream if it is not already closed here?

net/host.go
```diff
@@ -183,16 +183,25 @@ func (h *Host) CloseProtocolStream(offerID types.Hash) {
 	h.swapMu.RLock()
 	swap, has := h.swaps[offerID]
 	h.swapMu.RUnlock()
-	if !has {
+	if !has || swap.streamClosed {
 		return
 	}
 
+	swap.streamClosed = true
```
dimalinux, line 190: We need a write lock instead of a read lock to add this line.

net/host.go
noot: the caller of this function always closes the stream as well, I can make a note of that in the function comment

dimalinux: Can we ether add a panic (if we are certain it can never happen) or an error level log if the assertion in the comment turns out to be false at some point? That way if we introduce a slow resource leak in the future, it will get detected quickly.

net/initiate.go
```diff
@@ -126,29 +123,35 @@ func (h *Host) handleProtocolStream(stream libp2pnetwork.Stream) {
 		return
 	}
 
-	var s SwapState
-	s, resp, err := h.makerHandler.HandleInitiateMessage(curPeer, im)
-	if err != nil {
-		log.Warnf("failed to handle protocol message: err=%s", err)
+	h.swapMu.Lock()
+	if h.swaps[im.OfferID] != nil {
+		log.Warnf("ignoring attempting initiation of swap %s: %s", im.OfferID, errSwapAlreadyInProgress)
```
dimalinux, line 128: If I'm understanding correctly what this is checking, I think we want a clearer message. Would it be correct to say something like this:
`ignoring initiation from peerID=%s for offer ID=%s that is already taken`

It is a scenario that will probably happen frequently if the offer is really good.

net/initiate.go
```diff
@@ -126,29 +123,35 @@ func (h *Host) handleProtocolStream(stream libp2pnetwork.Stream) {
...
+	// HandleInitiateMessage requires the network to be aware of the swap's stream,
+	// since it sends the SendKeysMessage response using that stream.
+	h.swaps[im.OfferID] = &swap{
+		stream: stream,
+	}
+	h.swapMu.Unlock()
+
+	s, err := h.makerHandler.HandleInitiateMessage(curPeer, im)
+	if err != nil {
+		log.Warnf("failed to handle protocol message: err=%s", err)
```
dimalinux, line 144: It would be nice if we used a warning level log if something failed on our end, but a low log level for bad requests, since those are not our problem. We don't have that level of knowledge this high in the stack and that change is probably out of scope for this PR, but it is something we should be thinking about.

net/initiate_test.go
```diff
@@ -46,7 +46,7 @@ func TestHost_Initiate(t *testing.T) {
 
 	err = ha.Initiate(hb.h.AddrInfo(), createSendKeysMessage(t), new(mockSwapState))
 	require.NoError(t, err)
-	time.Sleep(time.Millisecond * 500)
+	time.Sleep(time.Second * 2)
```
dimalinux, line 49: Is CI really that slow on a localhost only initiate?

protocol/xmrmaker/net.go
```diff
@@ -75,6 +76,7 @@ func (inst *Instance) initiate(
 
 	symbol, err := pcommon.AssetSymbol(inst.backend, offer.EthAsset)
 	if err != nil {
+		_ = s.Exit()
```
dimalinux, line 79: Would that error get logged somewhere else, or are we just swallowing it? We could log the one that is not propagated up, or we could do this:
```
	if exitErr := s.Exit(); err != nil {
		return nil, multierror.Append(err, exitErr)
	}
```

protocol/xmrmaker/net.go
```diff
@@ -144,18 +146,25 @@ func (inst *Instance) HandleInitiateMessage(
...
-		return nil, nil, err
+		_ = state.Exit()
+		return nil, err
 	}
 
 	resp := state.SendKeysMessage()
-	return state, resp, nil
+	err = inst.backend.SendSwapMessage(resp, offer.ID)
+	if err != nil {
+		_ = state.Exit()
```
dimalinux, line 165: Whatever solution you use on the other swallowed error should be here too.

net/initiate.go
noot: the `errSwapAlreadyInProgress` says that the swap is already ongoing, should i change it to say "offer no longer available" or something like that? and yeah the offer isn't advertised when it's taken so this would only happen if two users try to initiate at basically the same time

protocol/xmrmaker/net.go
noot: we're swallowing it but in this case I think it's fine, the swap is being aborted anyways since initiation failed. can add multierror if you think that's useful

noot: will add a log in `s.Exit()` for now

dimalinux (approved):
Approved, but let's fix that one warning level message that I think people will actually see from time to time before merging.

PR #448, opened 2023-04-30
feat: update GetPast() and GetOngoing() to sort from newest to oldest

### LINKED ISSUE #436: past swaps should be in reverse chronological order
Right now viewing past swaps with `swapcli past` shows the oldest swaps first, but I think that most users would prefer to see the recent swaps first.

bingcicle:
closes #436 

Changes the RPC call to sort from newest to oldest.

While looking into this I realized that `GetOngoing()` also returns swaps sorted from oldest to newest:

https://github.com/AthanorLabs/atomic-swap/blob/906a5c69a386448071f048bc05b8b198f779ae9c/rpc/swap.go#L201-L203

Should this be updated as well in a separate PR for consistency?


dimalinux:
> Should this be updated as well in a separate PR for consistency?

Yes, it should be updated, but let's change them all in a single PR.

rpc/swap.go
```diff
@@ -121,7 +121,7 @@ func (s *SwapService) GetPast(_ *http.Request, req *GetPastRequest, resp *GetPas
 	}
 
 	sort.Slice(resp.Swaps, func(i, j int) bool {
-		return resp.Swaps[i].StartTime.UnixNano() < resp.Swaps[j].StartTime.UnixNano()
+		return resp.Swaps[j].StartTime.UnixNano() < resp.Swaps[i].StartTime.UnixNano()
```
dimalinux, line 124: I know you didn't write the original code, but `Time` has built-in methods `Before` and `After` that are the preferred way to do comparisons. I don't think the extra efficiency matters much here, but switching to the preferred methods might keep this style from spreading further in to the code base.

dimalinux (approved):

PR #449, opened 2023-04-30
feat: update SwapCreater.sol `claimRelayer` to no longer use forwarder

noot:
- update contract to no longer use forwarder :D this implementation should be front-running-proof.
- remove all forwarder usages from codebase
- update `claimRelayer` to instead accept a signature from the claimer, signing the relayer's address, fee, and the current SwapCreator address, along with their usual swap struct.
- update `relayer` package respectively
- the contract `RelaySwap` structure needs to contain the relayer payout address to prevent front-running. because of this, I had to add a `relayerquery` protocol to the network that sends the relayer's address to a querier so that it can be signed. (note that this is not used for counterparty-relays as we already know the counterparty's eth address)
- todo: maybe remove the simulation call if the contract call already does it? since we're not using a forwarder anymore


daemon/swap_daemon_test.go
```diff
@@ -83,14 +83,14 @@ func minimumFundAlice(t *testing.T, ec extethclient.EthClient, providesAmt *apd.
...
+	//   relayClaim: 83967
 	//
 	const (
-		aliceGasRation = 150000 // roughly 10% more than newSwap+setRead+refund
+		aliceGasRation = 120000 // roughly 10% more than newSwap+setRead+refund
 	)
 	// We give Alice enough gas money to refund if needed, but not enough to
 	// relay a claim:
-	//    150000 - (53787 + 34452) = 61761
+	//    120000 - (53787 + 34452) = 51761
```
noot, line 93: note to self: the math here is wrong, need to fix

ethereum/contracts/SwapCreator.sol
```diff
@@ -41,6 +40,20 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
...
 
+    // RelaySwap contains additional information required for relayed transactions.
+    // This entire structure is encoded and signed by the swap claimer, and the signature is
+    // passed to the `claimRelayer` function.
+    struct RelaySwap {
+        // the swap the transaction is for
+        Swap swap;
+        // the fee, in wei, paid to the relayer
+        uint256 fee;
+        // the address to the paid the relayer fee
```
noot, line 51: to be*

ethereum/contracts/SwapCreator.sol
```diff
@@ -190,45 +205,55 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
 
     // Bob can claim if:
     // - (Alice has set the swap to `ready` or it's past timeout0) and it's before timeout1
-    // This function is only callable by the trusted forwarder.
-    // It transfers the fee to the originator of the transaction.
-    function claimRelayer(Swap memory _swap, bytes32 _s, uint256 fee) public {
-        if (!isTrustedForwarder(msg.sender)) revert OnlyTrustedForwarder();
-        _claim(_swap, _s);
+    // It transfers the fee to the relayer address specified in `_relaySwap`.
+    // Note: this function will revert if the swa value is less than the relayer fee;
```
noot, line 209: swap*

ethereum/contracts/SwapCreator.sol
```diff
@@ -190,45 +205,55 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
...
 
         // send ether to swap claimer, subtracting the relayer fee
-        // which is sent to the originator of the transaction.
-        // tx.origin is okay here, since it isn't for authentication purposes.
-        if (_swap.asset == address(0)) {
-            _swap.claimer.transfer(_swap.value - fee);
-            payable(tx.origin).transfer(fee); // solhint-disable-line
+        if (_relaySwap.swap.asset == address(0)) {
+            _relaySwap.swap.claimer.transfer(_relaySwap.swap.value - _relaySwap.fee);
+            payable(_relaySwap.relayer).transfer(_relaySwap.fee); // solhint-disable-line
```
noot, line 227: can get rid of the nolints

net/relay.go
```diff
@@ -92,9 +164,7 @@ func (h *Host) SubmitClaimToRelayer(relayerID peer.ID, request *RelayClaimReques
 	if err != nil {
 		return nil, fmt.Errorf("failed to open stream with peer: err=%w", err)
 	}
-
 	defer func() { _ = stream.Close() }()
-	log.Debugf("opened relay stream: %s", stream.Conn())
```
noot, line 97: re add

protocol/backend/backend.go
```diff
@@ -285,3 +293,36 @@ func (b *backend) HandleRelayClaimRequest(
...
+) (*message.RelayClaimResponse, error) {
+	if offerID == nil {
+		// this isn't a counterparty-relay, get the relayer's eth address
+		relayerAddr, err := b.QueryRelayerAddress(relayerID)
+		if err != nil {
+			return nil, err
+		}
+
+		// set relayer address and sign as front-run prevention
+		relaySwap.Relayer = relayerAddr
```
noot, line 315: this will be set in the caller if offerID is set (add comment)

relayer/claim_request.go
```diff
@@ -37,30 +34,41 @@ func CreateRelayClaimRequest(
...
+	if relaySwap.Swap.Claimer != signerAddress {
+		return nil, fmt.Errorf("signing key %s does not match claimer %s", signerAddress, relaySwap.Swap.Claimer)
+	}
+
+	msg := relaySwap.Hash()
+
+	// signature format is (r || s || v), v = 27/28
+	signature, err := rcommon.NewKeyFromPrivateKey(claimerEthKey).Sign(msg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign forward request digest: %w", err)
```
noot, line 70: update comment

also can probably remove the reliance on go-relayer, this is the only place it's used

ethereum/contracts/SwapCreator.sol
```diff
@@ -190,45 +205,55 @@ contract SwapCreator is ERC2771Context, Secp256k1 {
...
             // WARN: this will FAIL for fee-on-transfer or rebasing tokens if the token
             // transfer reverts (i.e. if this contract does not contain _swap.value tokens),
             // exposing Bob's secret while giving him nothing.
-            IERC20(_swap.asset).transfer(_swap.claimer, _swap.value - fee);
-            IERC20(_swap.asset).transfer(tx.origin, fee); // solhint-disable-line
+            IERC20(_relaySwap.swap.asset).transfer(
+                _relaySwap.swap.claimer,
+                _relaySwap.swap.value - _relaySwap.fee
+            );
+            IERC20(_relaySwap.swap.asset).transfer(_relaySwap.relayer, _relaySwap.fee); // solhint-disable-line
```
dimalinux, line 236: Do we want this `else` branch in the contract now, when our code does not currently allow relaying ERC20 swaps? (Just asking, I'm not against it, but it might confuse someone into thinking that we support it.)

daemon/swap_daemon_test.go
dimalinux: I changed these to use our new constants. Then I was disturbed to see that the tests in our daemon code were generating higher gas usage than we've ever recorded from the tests in `contracts`. I adjusted the `MaxNewSwapETHGas` and `MaxSetReadyGas` in your branch with the new recorded highs, but haven't done an across the board check. Maybe we can figure out where the tests in `contracts` are not being realistic enough.

ethereum/contracts/SwapCreator.sol
noot: idk, doesn't hurt to have it, better than redeploying the contract if we decide to support it

dimalinux: I'm wondering if we might end up deploying several times before figuring out how to help a user that wants ERC20 tokens in an account that has no ETH to do anything with the tokens.  :)

net/host.go
dimalinux, line 1: We need to increase the protocol ID in this file, since we have a new contract.

net/host.go
```diff
@@ -237,7 +238,7 @@ func readStreamMessage(stream libp2pnetwork.Stream, maxMessageSize uint32) (comm
 // nextStreamMessage returns a channel that will receive the next message from the stream.
 // if there is an error reading from the stream, the channel will be closed, thus
 // the received value will be nil.
-func nextStreamMessage(stream libp2pnetwork.Stream, maxMessageSize uint32) <-chan common.Message {
+func nextStreamMessage(stream libp2pnetwork.Stream, maxMessageSize uint32) <-chan common.Message { //nolint:unparam
```
dimalinux, line 241: `maxMessageSize` is shadowing the package-local constant with the same name. You could give the new protocol its own constant and use it, then rename the existing constant from `maxMessageSize` to something like `swapMaxMessageSize`. That would remove the shadowing and the linter complaint that we should eliminate the parameter because we only ever pass in a constant.

net/host.go
noot: why? I'd rather not since we'd need everyone running bootnodes to restart, and bootnodes wouldn't be affected by this change anyways. also I only know of bootnodes + 1 relayer currently on the network (relayer will need to rebuild either way)

ethereum/contracts/SwapCreator.sol
noot: with account abstraction, they should technically be able to use ERC20s in an account with no tokens, but yeah the user needs to be aware of that lol

daemon/swap_daemon_test.go
noot: maybe zero values are being used in the contract tests?

ethereum/contracts/SwapCreator.sol
```diff
@@ -48,8 +48,8 @@ contract SwapCreator is Secp256k1 {
         Swap swap;
         // the fee, in wei, paid to the relayer
         uint256 fee;
-        // the address to be paid the relayer fee
-        address relayer;
+        // hash of (relayer's payout address || 4-byte salt)
+        bytes32 relayerHash;
```
dimalinux, line 52: The randomized portion usually goes at the front of the hash. If someone wants to test a particular result against a list of addresses, they'll have to hash more bytes if the salt is first. It would need to be a large list for it to matter though.

dimalinux (approved):

PR #450, opened 2023-05-02
time index shift

dimalinux:
Renames T0/T1 to T1/T2. This way T0 can represent the block timestamp that `newSwap` was mined into.


ethereum/contracts/SwapCreator.sol
```diff
@@ -156,8 +156,8 @@ contract SwapCreator is Secp256k1 {
             pubKeyClaim: _pubKeyClaim,
             pubKeyRefund: _pubKeyRefund,
             claimer: _claimer,
-            timeout0: block.timestamp + _timeoutDuration0,
-            timeout1: block.timestamp + _timeoutDuration0 + _timeoutDuration1,
+            timeout1: block.timestamp + _timeoutDuration0,
+            timeout2: block.timestamp + _timeoutDuration0 + _timeoutDuration1,
```
dimalinux, line 160: I'm not sure which indices on these duration values would make them easiest to understand. It depends on whether you think of it as the duration starting from T0 and T1, or if you think of it as the duration until T1 and T2.

PR #451, opened 2023-05-02
eth endpoint defaults and error handling

dimalinux:
This PR uses our environment config to set the ETH endpoint defaults for dev and stagenet. For mainnet, instead of giving an error because the code cannot connect to the default ganache port on localhost, we tell them that they need to set the endpoint.

Notable change: Our tests are using the http endpoint starting with this PR instead of the ws endpoint.

### DIFF
--- a/cmd/swapd/main.go
+++ b/cmd/swapd/main.go
@@ -427,7 +427,9 @@ func createEthClient(c *cli.Context, envConf *common.Config) (extethclient.EthCl
 	}
 	if ethEndpoint == "" {
 		// Message is mainnet specific, because we have defaults for dev/stagenet
-		return nil, errors.New("missing ETH endpoint; completely open endpoints are not reliable for swaps")
+		return nil, fmt.Errorf(
+			"--%s flag required, note that open endpoints are unreliable for mainnet swaps", flagEthEndpoint,
+		)
 	}
 
 	var ethPrivKey *ecdsa.PrivateKey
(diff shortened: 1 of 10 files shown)


cmd/swapd/main.go
```diff
@@ -421,10 +421,16 @@ func createMoneroClient(c *cli.Context, envConf *common.Config) (monero.WalletCl
...
-	ethEndpoint := common.DefaultEthEndpoint
-	if c.String(flagEthEndpoint) != "" {
+	ethEndpoint := envConf.EthEndpoint
+	if c.IsSet(flagEthEndpoint) {
 		ethEndpoint = c.String(flagEthEndpoint)
 	}
+	if ethEndpoint == "" {
+		// Message is mainnet specific, because we have defaults for dev/stagenet
+		return nil, fmt.Errorf(
+			"--%s flag required, note that open endpoints are unreliable for mainnet swaps", flagEthEndpoint,
```
noot, line 431:
```suggestion
			"--%s flag required, note that public endpoints are generally unreliable for mainnet swaps", flagEthEndpoint,
```

noot (approved):
looks good, minor comment on making the mainnet message clearer

cmd/swapd/main.go
noot: omit the generally if you want :p but I think public is clearer than open?

PR #454, opened 2023-05-03
makefile targets for release binaries

dimalinux:
Added Makefile targets to build the latest tagged release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,24 @@ test:
 test-integration: 
 	./scripts/run-integration-tests.sh 2>&1 | tee test-integration.log
 
+# Instead of building from the local checked-out source, this will install
+# the most recent commit with a release tag. Use the most recent tagged release
+# for production swaps.
+.PHONY: build-release
+build-release:
+	go install -tags=prod github.com/athanorlabs/atomic-swap/cmd/...@latest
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -43,6 +43,24 @@ test:
...
 	./scripts/run-integration-tests.sh 2>&1 | tee test-integration.log
 
+# Instead of building from the local checked-out source, this will install
+# the most recent commit with a release tag. Use the most recent tagged release
+# for production swaps.
+.PHONY: build-release
+build-release:
+	go install -tags=prod github.com/athanorlabs/atomic-swap/cmd/...@latest
+	mkdir -p bin
+	mv $(GOPATH)/bin/swapd $(GOPATH)/bin/swapcli $(GOPATH)/bin/bootnode bin/
```
noot, line 53: why not just build directly into bin/ instead of moving?

noot (approved):
looks good, minor comment

Makefile
dimalinux: Brain flop on my part. Fixed. I had worked on that multi-platform script recently and you can't set GOBIN when doing a cross-platform build, but since this is a native build, that restriction does not apply.

PR #456, opened 2023-05-03
fix: start timeout watchers for xmrtaker when restarting ongoing swap

noot:


protocol/xmrtaker/swap_state.go
```diff
@@ -560,6 +565,10 @@ func (s *swapState) generateAndSetKeys() error {
 
 // getSecret secrets returns the current secret scalar used to unlock funds from the contract.
 func (s *swapState) getSecret() [32]byte {
+	if s.dleqProof == nil {
+		return [32]byte(common.Reverse(s.privkeys.SpendKey().Bytes()))
+	}
```
dimalinux, line 570: If there are comments that could help the next person editing this code understand when the variables might be in these different states, it might be good to add them.

noot: yeah, basically the solidity expects the bytes to be big endian, and the ed25519 lib uses little endian

PR #459, opened 2023-05-04
fix: add `Claimed` log watcher to xmrmaker

noot:
updates xmrmaker to exit the swap when it sees a `Claimed` log, meaning it successfully claimed the ETH.


protocol/xmrmaker/watcher.go
```diff
@@ -40,7 +41,18 @@ func (s *swapState) runContractEventWatcher() {
 				log.Debugf("EventETHRefunded sent, returning from event watcher")
 				return
 			}
+		case l := <-s.logClaimedCh:
+			eventSent, err := s.handleClaimedLogs(&l)
+			if err != nil {
+				log.Errorf("failed to handle claim logs: %s", err)
+			}
```
dimalinux, line 48: Was there supposed to be a return above?

dimalinux (approved):

protocol/xmrmaker/watcher.go
noot: added a comment

PR #460, opened 2023-05-04
wait for claim relayer receipt

dimalinux:
Renamed `waitForClaimReceipt` to `waitForClaimRelayerReceipt` and added fixes.

### DIFF
--- a/protocol/xmrmaker/claim.go
+++ b/protocol/xmrmaker/claim.go
@@ -137,7 +137,7 @@ func (s *swapState) relayClaimWithXMRTaker() (*ethtypes.Receipt, error) {
 		return nil, err
 	}
 
-	receipt, err := waitForClaimReceipt(
+	receipt, err := waitForClaimRelayerReceipt(
 		s.ctx,
 		s.ETHClient().Raw(),
 		response.TxHash,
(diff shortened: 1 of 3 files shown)


protocol/xmrmaker/claim.go
```diff
@@ -234,63 +234,58 @@ func waitForClaimReceipt(
...
+		if err != nil && !errors.Is(err, ethereum.NotFound) {
 			return nil, err
 		}
-
-		_, isPending, err := ec.TransactionByHash(ctx, txHash)
+		// If err is still set, the error was ethereum.NotFound, which is returned
+		// by TransactionReceipt even if our endpoint sees the TX as pending.
 		if err != nil {
-			// allow up to 5 NotFound errors, in case there's some network problems
-			if errors.Is(err, ethereum.NotFound) && notFoundCount >= maxNotFound {
```
dimalinux, line 257: I was wrong in my initial assessment. The check on what was formerly line 270 is correct, because it is `!isPending`. The actual problem was on this line, which should have been continuing the loop if `notFoundCount` was less than `maxNotFound`.

protocol/xmrmaker/claim.go
```diff
@@ -234,63 +234,58 @@ func waitForClaimReceipt(
 	secret [32]byte,
 ) (*ethtypes.Receipt, error) {
 	const (
-		checkInterval = time.Second // time between transaction polls
-		maxWait       = time.Minute // max wait for the tx to be included in a block
-		maxNotFound   = 10          // max failures where the tx is not even found in the mempool
+		checkInterval = 15000 * time.Second // 1.5 seconds between poll attempts
```
noot, line 237:
```suggestion
		checkInterval = 15000 * time.Millisecond // 1.5 seconds between poll attempts
```

noot: i pushed this

PR #461, opened 2023-05-04
fix: handle "insufficient funds" error on claim

noot:
for some reason, although I had more than enough funds to claim, I still got an `insufficient funds` error. this PR updates the code to handle that case, and use a relayer if we're out of funds


protocol/xmrmaker/claim.go
```diff
@@ -60,9 +61,19 @@ func (s *swapState) claimFunds() (*ethtypes.Receipt, error) {
 		sc := s.getSecret()
 		receipt, err = s.sender.Claim(s.contractSwap, sc)
 		if err != nil {
-			return nil, err
+			if strings.Contains(err.Error(), "insufficient funds for gas * price + value") {
```
dimalinux, line 64: We can update this to `core.ErrInsufficientFunds` from `go-ethereum` if we want to avoid the message contents check.

PR #467, opened 2023-05-06
bootnode version fix

### LINKED ISSUE #202: deprecate `contract-addresses.json`

### LINKED ISSUE #465: bootnode fails when you query its version
The bootnode fails right now when you query its version, because it is trying to use the ethereum endpoint to calculate the chainID for the version. The fix is to just calculate the chainID from the environment.

dimalinux:
`daemon_version` RPC requests on the bootnode where causing it to crash. It was accessing the `protocolBackend` interface, which is nil for bootnodes to get the chain-ID as well as the contract address. We also had Prometheus metrics that were added that were accessing `protocolBackend`.

Implemented solutions:
* We now get the chain-ID from the environment for `daemon_version`. We verify that the chain-ID matches the environment on start, so this should always be accurate.
* Bootnodes do not report a contract address in the version response. `swapd` instances report the contract address in the version because they can change it. Bootnodes never access the contract, so it makes sense for them to just omit the value.
* Prometheus stats are not started for bootnodes. If we want some stats for bootnodes in the future, we'll need to add them in a separate PR.
* Bootnodes were integrated into the integration tests, and some unit tests, so we don't run into this issue again. The new tests immediately caught the issue with the Prometheus stats.
* Since the integration test startup script was getting a partial rewrite, I removed our dependency on the `contract-addresses.json` closing an extra ticket from the backlog.

Closes #465 
Closes #202


cmd/bootnode/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/athanorlabs/atomic-swap/rpcclient"
+)
+
+func getFreePort(t *testing.T) uint16 {
+	port, err := common.GetFreeTCPPort()
+	require.NoError(t, err)
+	return uint16(port)
+}
+
+func TestBootnode(t *testing.T) {
```
dimalinux, line 27: I'd like to add unit tests to the `bootnode` package and maybe even use them in the `daemon` package's tests, but this seemed like the most bang-for-the-buck place to introduce initial bootnode unit tests.

rpc/daemon.go
```diff
@@ -28,19 +34,20 @@ func (s *DaemonService) Shutdown(_ *http.Request, _ *any, _ *any) error {
...
-	SwapCreatorAddr ethcommon.Address  `json:"swapCreatorAddress" validate:"required"`
+	SwapCreatorAddr *ethcommon.Address `json:"swapCreatorAddress,omitempty"`
 }
 
 // Version returns version & misc info about swapd and its dependencies
 func (s *DaemonService) Version(_ *http.Request, _ *any, resp *VersionResponse) error {
 	resp.SwapdVersion = cliutil.GetVersion()
-	resp.P2PVersion = fmt.Sprintf("%s/%d", net.ProtocolID, s.pb.ETHClient().ChainID())
-	resp.Env = s.pb.Env()
-	resp.SwapCreatorAddr = s.pb.SwapCreatorAddr()
```
dimalinux, line 44: pb was `nil` on bootnodes, so those 3 lines were problematic.

scripts/run-integration-tests.sh
```diff
@@ -32,41 +32,62 @@ create-eth-keys() {
 		echo "${GANACHE_KEYS[${i}]}" >"${key_file}"
 	done
 }
-# This is the local multiaddr created when using ./tests/alice-libp2p.key on the default libp2p port
-ALICE_MULTIADDR=/ip4/127.0.0.1/tcp/9933/p2p/12D3KooWAAxG7eTEHr2uBVw3BDMxYsxyqfKvj3qqqpRGtTfuzTuH
-ALICE_LIBP2PKEY=./tests/alice-libp2p.key
```
dimalinux, line 37: I removed `alice-libp2p.key` from the repo in this PR. We query the boootnode for it's p2p address using swapcli.

noot (approved):
nice! this looks good to me

PR #469, opened 2023-05-12
added a bootnode to the daemon tests

dimalinux:
The daemon tests were previously using the first `swapd` instance created by `LaunchDaemons` as the bootnode for all subsequent daemons. This made writing more complicated tests fragile. If you passed `LaunchDaemons(... aliceConf, bobConf)`, you couldn't stop and restart Bob's swapd instance unless Alice's was still online as one example. I'm working on tests that need this feature and wanted to get the infrastructure changes in as a separate PR.


daemon/retake_failed_swap_test.go
```diff
@@ -0,0 +1 @@
+package daemon
```
noot, line 1: why is this file empty?

noot (approved):

daemon/retake_failed_swap_test.go
dimalinux: Sorry, didn't mean for that file to make it into the PR. Removed. it.

PR #470, opened 2023-05-13
fix: store `newSwap` tx in db, handle pending `newSwap` on restart

### LINKED ISSUE #443: write newSwap TX ID to database as soon as TX is sent to mempool
Unexpected things can happen in production like a NewSwap transaction getting mined 2 hours after it was sent. We should save the TX ID to the database before waiting for its receipt, so out tooling still help the user make a refund if the TX gets mined, even if swapd is shutdown before the TX is included in a block.

noot:
- write `newSwap` tx to db when it's broadcast and before it's included
- on restart, if we're in a state where we see the `newSwap` tx was broadcast but not yet included before we shut down, we attempt to refund or cancel the swap
- in the case the tx was included during the time we were offline, we attempt to refund 
- in the case it hasn't yet been included, we attempt to cancel the tx by sending a replacement tx with the same nonce but higher gas

closes #443


protocol/xmrtaker/instance.go
```diff
@@ -119,6 +146,148 @@ func (inst *Instance) abortOngoingSwap(s *swap.Info) error {
...
+		return false, fmt.Errorf("failed to get cancel transaction receipt: %w", err)
+	}
+
+	if receipt.Status == 0 {
+		// this is okay, it means newSwap was included, and we can refund it in the calling function
+		log.Infof("failed to cancel swap, attempting to refund")
+		return false, nil
+	}
+
+	// TODO: check for receipt success; there's still a case newSwap might be included
```
noot, line 286: remove

common/consts.go
```diff
@@ -28,9 +28,10 @@ const (
 
 // SwapCreator.sol event signatures
 const (
-	ReadyEventSignature    = "Ready(bytes32)"
-	ClaimedEventSignature  = "Claimed(bytes32,bytes32)"
-	RefundedEventSignature = "Refunded(bytes32,bytes32)"
+	NewSwapFunctionSignature = "newSwap(bytes32,bytes32,address,uint256,uint256,address,uint256,uint256)"
+	ReadyEventSignature      = "Ready(bytes32)"
+	ClaimedEventSignature    = "Claimed(bytes32,bytes32)"
+	RefundedEventSignature   = "Refunded(bytes32,bytes32)"
```
dimalinux, line 34: These are all available programmatically if we don't want to hard code them. Or we could could have a unit test that fails if their values don't match the current ABI.

```go
func TestABISigs(t *testing.T) {
	for name, event := range SwapCreatorParsedABI.Events {
		t.Logf("%s: %s", name, event.Sig)
	}
	t.Log()
	for name, method := range SwapCreatorParsedABI.Methods {
		t.Logf("%s: %s", name, method.Sig)
	}
}
```

daemon/cancel_or_refund_test.go
```diff
@@ -0,0 +1,139 @@
...
+// Bob should have aborted the swap in all cases.
+func TestXMRTakerCancelOrRefundAfterKeyExchange(t *testing.T) {
+	minXMR := coins.StrToDecimal("1")
+	maxXMR := minXMR
+	exRate := coins.StrToExchangeRate("300")
+	providesAmt, err := exRate.ToETH(minXMR)
+	require.NoError(t, err)
+
+	ec, err := ethclient.Dial("http://localhost:8545")
+	require.NoError(t, err)
```
dimalinux, line 41: You can replace those 2 lines with this:
```go
	ec, _ := tests.NewEthClient(t)
```

daemon/cancel_or_refund_test.go
```diff
@@ -0,0 +1,139 @@
...
+	makeResp, bobStatusCh, err := bws.MakeOfferAndSubscribe(minXMR, maxXMR, exRate, types.EthAssetETH, false)
+	require.NoError(t, err)
+
+	aliceStatusCh, err := aws.TakeOfferAndSubscribe(makeResp.PeerID, makeResp.OfferID, providesAmt)
+	require.NoError(t, err)
+
+	var statusWG sync.WaitGroup
+	statusWG.Add(2)
+
+	go func() {
```
dimalinux, line 69: Can we add this go-routine to the waitgroup to ensure that we never start the next test with a previous-test go-routine still running? The context can get cancelled by the timeout passed to `LaunchDaemons`, so it is not a guarantee that this go-routine is the one shutting down the other 2. In background go routines we should only use `assert.*` or `t.Errorf(...)` to ensure that we safely cleanup all go processes. (There is a `require` a few lines down that we should convert.)

daemon/cancel_or_refund_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	t.Logf("daemons stopped, now re-launching them")
+	_, _ = LaunchDaemons(t, 3*time.Minute, bobConf, aliceConf)
+
+	pastSwap, err := ac.GetPastSwap(&makeResp.OfferID)
+	require.NoError(t, err)
+	t.Logf("Alice past status: %s", pastSwap.Swaps[0].Status)
+	if pastSwap.Swaps[0].Status.String() != types.CompletedRefund.String() {
+		t.Errorf("Alice should have refunded the swap")
+	}
```
dimalinux, line 133: You could rewrite this block to be panic free and print the same message if we don't get the expected value:
```go
	pastSwap, err := ac.GetPastSwap(&makeResp.OfferID)
	require.NoError(t, err)
	require.NotEmpty(t, pastSwap.Swaps)
	require.Equal(t, types.CompletedRefund.String(), pastSwap.Swaps[0].Status.String(),
		"Alice should have refunded the swap")
```

ethereum/extethclient/eth_wallet_client.go
```diff
@@ -298,6 +301,30 @@ func (c *ethClient) Raw() *ethclient.Client {
...
+	signer := ethtypes.LatestSignerForChainID(c.chainID)
+	signedTx, err := ethtypes.SignTx(tx, signer, c.ethPrivKey)
+	if err != nil {
+		return ethcommon.Hash{}, fmt.Errorf("failed to sign tx: %w", err)
+	}
+
+	err = c.ec.SendTransaction(ctx, signedTx)
+	if err != nil {
+		return ethcommon.Hash{}, fmt.Errorf("failed to send tx: %w", err)
+	}
```
dimalinux, line 323: I see that we are waiting for the receipt in the caller, but maybe we should do it here? (And in the Transfer method as well?) It hasn't really been cancelled until the tx is included in a block. In the monero wallet methods we are waiting for the transactions to be mined. If multiple callers would will all be writing the exact same boilerplate code for the receipt, that would be an argument for doing it here. If you think different callers will want distinct receipt handling, the way it is now would be better.

protocol/xmrtaker/instance.go
```diff
@@ -119,6 +146,148 @@ func (inst *Instance) abortOngoingSwap(s *swap.Info) error {
...
+		return false, fmt.Errorf("failed to create or send cancel tx: %w", err)
+	}
+
+	log.Infof("submit cancel tx %s", cancelTx)
+	receipt, err := block.WaitForReceipt(inst.backend.Ctx(), inst.backend.ETHClient().Raw(), cancelTx)
+	if err != nil {
+		return false, fmt.Errorf("failed to get cancel transaction receipt: %w", err)
+	}
+
+	if receipt.Status == 0 {
```
dimalinux, line 280: There are constants for these in the ethtypes package:
```go
const (
	// ReceiptStatusFailed is the status code of a transaction if execution failed.
	ReceiptStatusFailed = uint64(0)

	// ReceiptStatusSuccessful is the status code of a transaction if execution succeeded.
	ReceiptStatusSuccessful = uint64(1)
)
```

protocol/xmrtaker/swap_state.go
```diff
@@ -669,7 +661,39 @@ func (s *swapState) lockAsset() (*ethtypes.Receipt, error) {
 		return nil, err
 	}
 
-	log.Infof("locked %s in swap contract, waiting for XMR to be locked", providedAmt.StandardSymbol())
+	log.Infof("locked %s in swap contract, waiting for XMR to be locked", s.providedAmount.StandardSymbol())
+	return receipt, nil
+}
+
+func (s *swapState) lockAndWaitForReceipt(
+	cmtXMRMaker, cmtXMRTaker [32]byte,
```
dimalinux, line 669: Should we match these names to the names in the contract?
`pubKeyClaim` and `pubKeyRefund`

It's helpful that the above names have Maker and Taker in them, but I'm not sure everyone will know that `cmt` means commitment and what exactly the maker/taker committed to.

protocol/xmrtaker/swap_state.go
```diff
@@ -669,7 +661,39 @@ func (s *swapState) lockAsset() (*ethtypes.Receipt, error) {
 		return nil, err
 	}
 
-	log.Infof("locked %s in swap contract, waiting for XMR to be locked", providedAmt.StandardSymbol())
+	log.Infof("locked %s in swap contract, waiting for XMR to be locked", s.providedAmount.StandardSymbol())
+	return receipt, nil
+}
+
+func (s *swapState) lockAndWaitForReceipt(
```
dimalinux, line 668: The maker function is called `lockFunds`. `lockXMRFunds` and `lockETHFunds` might make it easier to search for the function you want. I think the receipt portion of the function name is implied, as the funds are not locked if the transaction is still in the mempool.

ethereum/extethclient/eth_wallet_client.go
noot: yeah I wrote it as is because the callers might want to handle the receipt checking differently

protocol/xmrtaker/swap_state.go
noot: it would make more sense to change the names in the contract, since what's stored is a hash of the swap public key, not the public key

protocol/xmrtaker/swap_state.go
noot: this function is called inside the function `lockAsset`, should i rename that to `lockFunds`?

protocol/xmrtaker/instance.go
```diff
@@ -119,6 +146,148 @@ func (inst *Instance) abortOngoingSwap(s *swap.Info) error {
...
+	}
+
+	log.Infof("refunded swap %s successfully: %s", s.OfferID, common.ReceiptInfo(receipt))
+
+	// set status to refunded
+	s.Status = types.CompletedRefund
+	return inst.backend.SwapManager().CompleteOngoingSwap(s)
+}
+
+func (inst *Instance) maybeCancelNewSwap(txHash ethcommon.Hash) (bool, error) {
```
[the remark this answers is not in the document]
noot: this function only has one caller, `refundOrCancelNewSwap`, and its purpose is specifically to cancel a `newSwap` transaction. the generic function is `extethclient.CancelTxWithNonce`

protocol/xmrtaker/swap_state.go
dimalinux: This function appears to just be a wrapper around calling `newSwap` directly on the the contract. Looking at the functions below it that do very similar things, we are just calling them the same name as the function on the contract. Maybe calling it `newSwap` would make the code consistent?  Also: The next function down should be renamed to `setReady`. If you look at its documentation, it appears to have been written at a point in time when the contract method was called `ready`.

protocol/xmrtaker/swap_state.go
dimalinux: We could put the 2 values on separate lines and add a code comment after each one saying something like `commitment to public claim key` and `commitment to public refund key`.

ethereum/extethclient/eth_wallet_client.go
dimalinux: Feel free to disagree, but I think in this case we'd be best off being consistent and waiting for the receipt here. Then, when a secondary use case requiring different behavior becomes apparent, we can do the necessary code refactoring. There is a high probability that the alternate handing would be done completely in this package to promote maximum code reuse. I've spent so many hours of my life tracking down race conditions in the code base because receipts were not being waited on, that I'm in favor of making it hard for anyone accidentally not do it.  :)

daemon/cancel_or_refund_test.go
```diff
@@ -56,10 +54,11 @@ func TestXMRTakerCancelOrRefundAfterKeyExchange(t *testing.T) {
...
-	var statusWG sync.WaitGroup
-	statusWG.Add(2)
+	var wg sync.WaitGroup
+	wg.Add(3)
 
 	go func() {
+		defer wg.Done()
 		for {
 			count, err := ec.PendingTransactionCount(ctx) //nolint:govet
 			require.NoError(t, err)
```
dimalinux, line 64: We can't use `require` in a background go-process or the test won't fail for 30 minutes. Just converting to `assert` is also bad, because we don't want the code after it executed (we want the nested function to return).  There are multiple ways to fix it, but one way would be something like this:
```go
			if err != nil {
				t.Errorf("PendingTransactionCount failed: %s", err)
				return
			}
```

protocol/xmrtaker/instance.go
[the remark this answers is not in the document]
dimalinux: Knowing that, maybe we can make the errors returned by this function specific enough that the parent caller doesn't need to wrap the result?  The parent caller's wrapped message, that I was hoping we might remove, says: `"failed to maybe cancel newSwap: %w", err` which is likely confusing for end users with the word "maybe". Based on your answer, this function has more context to give a precise error message without help from the caller.

daemon/cancel_or_refund_test.go
noot: updated!

dimalinux (approved):
There are still a couple comments that I left that might be worth addressing, but looks good!

protocol/xmrtaker/swap_state.go
noot: I think renaming it `newSwap` is way too unclear, I don't think these function names need to match the contract functions. I'll rename/update the below function though

ethereum/extethclient/eth_wallet_client.go
noot: I'm actually going to argue we need it to be separate here. there may not actually ever be a receipt for the cancellation transaction in the case `newSwap` is executed, as the tx will simply be dropped from the mempool (and same for the newSwap tx in the case the cancellation is included) it would be better here to have a select statement that chooses between which of the two receipts is received, then check the status from there. I don't think the `WaitForReceipt` function should be used here at all, as only one of the two txs will ever be included, and thus one will return `ethereum.NotFound` and then loop inside `WaitForReceipt` for one hour :/

noot: I think I need to actually remove usage of `WaitForReceipt` here entirely because of that actually

protocol/xmrtaker/swap_state.go
noot: I think it's pretty well documented already in the swap contract and code-wise but can add if you think it's useful

PR #471, opened 2023-05-13
ensure that the status channel is always initialized

dimalinux:
This PR adds a new test case, `TestXMRNotLockedAndETHRefundedAfterAliceRestarts`, that was initially broken before the rest of the PR. The taker's status subscribe channel was not getting the refund status change after restart.

I decided to remove the status channel from both the `OfferExtra` and `swap.Info` struct types. Both of those types are designed for serialization (and deserialization on restart), but status channels are ephemeral to a single run of `swapd`. Instead of trying to keep the status channels synchronized across two different data structures, I made management of the channel part of `swap.Manager`.

### DIFF
--- a/common/types/offer.go
+++ b/common/types/offer.go
@@ -225,8 +225,6 @@ func (o *Offer) UnmarshalJSON(data []byte) error {
 
 // OfferExtra represents extra data that is passed when an offer is made.
 type OfferExtra struct {
-	StatusCh chan Status `json:"-"`
-
 	// UseRelayer forces the XMR maker to claim using the relayer even when he
 	// has enough funds to make the claim himself. Setting it to false will not
 	// prevent the relayer from being used if there are insufficient ETH funds
@@ -234,31 +232,9 @@ type OfferExtra struct {
 	UseRelayer bool `json:"useRelayer,omitempty"`
 }
 
-// NewStatusChannel creates a status channel using the the correct size
-func NewStatusChannel() chan Status {
-	// The channel size should be large enough to handle the max number of
-	// stages a swap can potentially go through.
-	const statusChSize = 6
-	return make(chan Status, statusChSize)
-}
-
-// NewOfferExtra creates an OfferExtra instance. Use this method, instead of
-// creating objects directly, to ensure that the status channel is initialized
-// correctly.
+// NewOfferExtra creates an OfferExtra instance
 func NewOfferExtra(forceUseRelayer bool) *OfferExtra {
 	return &OfferExtra{
-		StatusCh:   NewStatusChannel(),
 		UseRelayer: forceUseRelayer,
 	}
-}
-
-// UnmarshalJSON provides custom unmarshalling of OfferExtra to ensure that the
-// StatusCh field is always initialized.
-func (o *OfferExtra) UnmarshalJSON(data []byte) error {
-	type _OfferExtra OfferExtra
-	if err := vjson.UnmarshalStruct(data, (*_OfferExtra)(o)); err != nil {
-		return err
-	}
-	o.StatusCh = NewStatusChannel()
-	return nil
 }
--- a/common/types/offer.go
+++ b/common/types/offer.go
@@ -174,12 +174,6 @@ func (o *Offer) validate() error {
 	return nil
 }
 
-// OfferExtra represents extra data that is passed when an offer is made.
-type OfferExtra struct {
-	StatusCh   chan Status `json:"-"`
-	UseRelayer bool        `json:"useRelayer,omitempty"`
-}
-
 // UnmarshalOffer deserializes a JSON offer, checking the version for compatibility before
 // attempting to deserialize the whole blob.
 func UnmarshalOffer(jsonData []byte) (*Offer, error) {
@@ -227,4 +221,44 @@ func (o *Offer) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	return o.validate()
+}
+
+// OfferExtra represents extra data that is passed when an offer is made.
+type OfferExtra struct {
+	StatusCh chan Status `json:"-"`
+
+	// UseRelayer forces the XMR maker to claim using the relayer even when he
+	// has enough funds to make the claim himself. Setting it to false will not
+	// prevent the relayer from being used if there are insufficient ETH funds
+	// to claim.
+	UseRelayer bool `json:"useRelayer,omitempty"`
+}
+
+// NewStatusChannel creates a status channel using the the correct size
+func NewStatusChannel() chan Status {
+	// The channel size should be large enough to handle the max number of
+	// stages a swap can potentially go through.
+	const statusChSize = 6
+	return make(chan Status, statusChSize)
+}
+
+// NewOfferExtra creates an OfferExtra instance. Use this method, instead of
+// creating objects directly, to ensure that the status channel is initialized
+// correctly.
+func NewOfferExtra(forceUseRelayer bool) *OfferExtra {
+	return &OfferExtra{
+		StatusCh:   NewStatusChannel(),
+		UseRelayer: forceUseRelayer,
+	}
+}
+
+// UnmarshalJSON provides custom unmarshalling of OfferExtra to ensure that the
+// StatusCh field is always initialized.
+func (o *OfferExtra) UnmarshalJSON(data []byte) error {
+	type _OfferExtra OfferExtra
+	if err := vjson.UnmarshalStruct(data, (*_OfferExtra)(o)); err != nil {
+		return err
+	}
+	o.StatusCh = NewStatusChannel()
+	return nil
 }
--- a/daemon/test_support.go
+++ b/daemon/test_support.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"math/big"
 	"net"
+	"path"
 	"sync"
 	"syscall"
 	"testing"
@@ -87,13 +88,15 @@ func CreateTestBootnode(t *testing.T) (uint16, string) {
 		wg.Wait()
 	})
 
+	dataDir := t.TempDir()
+
 	conf := &bootnode.Config{
 		Env:           common.Development,
 		DataDir:       t.TempDir(),
 		Bootnodes:     nil,
 		P2PListenIP:   "127.0.0.1",
 		Libp2pPort:    0,
-		Libp2pKeyFile: common.DefaultLibp2pKeyFileName,
+		Libp2pKeyFile: path.Join(dataDir, common.DefaultLibp2pKeyFileName),
 		RPCPort:       uint16(rpcPort),
 	}
 
--- a/protocol/backend/backend.go
+++ b/protocol/backend/backend.go
@@ -262,7 +262,7 @@ func (b *backend) ClearXMRDepositAddress(offerID types.Hash) {
 // HasOngoingSwapAsTaker returns nil if we have an ongoing swap with the given peer where
 // we're the xmrtaker, otherwise returns an error.
 func (b *backend) HasOngoingSwapAsTaker(remotePeer peer.ID) error {
-	swaps, err := b.swapManager.GetOngoingSwaps()
+	swaps, err := b.swapManager.GetOngoingSwapsSnapshot()
 	if err != nil {
 		return err
 	}
@@ -293,7 +293,7 @@ func (b *backend) HandleRelayClaimRequest(
 			return nil, fmt.Errorf("cannot relay taker-specific claim request; no ongoing swap for swap %s", *request.OfferID)
 		}
 
-		info, err := b.swapManager.GetOngoingSwap(*request.OfferID)
+		info, err := b.swapManager.GetOngoingSwapSnapshot(*request.OfferID)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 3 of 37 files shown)


common/types/offer.go
```diff
@@ -228,3 +222,19 @@ func (o *Offer) UnmarshalJSON(data []byte) error {
...
+	// to claim.
+	UseRelayer bool `json:"useRelayer,omitempty"`
+}
+
+// NewOfferExtra creates an OfferExtra instance
+func NewOfferExtra(forceUseRelayer bool) *OfferExtra {
+	return &OfferExtra{
+		UseRelayer: forceUseRelayer,
+	}
+}
```
dimalinux, line 240: I was doing custom marshaling to ensure that the `StatusCh` was always initialized in an early version of this PR. Otherwise I wouldn't have made so many edits here. Everyone is calling `NewOfferExtra` now instead of creating the structure directly.

daemon/test_support.go
```diff
@@ -87,13 +88,15 @@ func CreateTestBootnode(t *testing.T) (uint16, string) {
...
+	dataDir := t.TempDir()
+
 	conf := &bootnode.Config{
 		Env:           common.Development,
 		DataDir:       t.TempDir(),
 		Bootnodes:     nil,
 		P2PListenIP:   "127.0.0.1",
 		Libp2pPort:    0,
-		Libp2pKeyFile: common.DefaultLibp2pKeyFileName,
+		Libp2pKeyFile: path.Join(dataDir, common.DefaultLibp2pKeyFileName),
```
dimalinux, line 99: This was a bug in my previous PR. Tests were littering network key files in the current working directory of the test.

protocol/backend/backend.go
```diff
@@ -262,7 +262,7 @@ func (b *backend) ClearXMRDepositAddress(offerID types.Hash) {
 // HasOngoingSwapAsTaker returns nil if we have an ongoing swap with the given peer where
 // we're the xmrtaker, otherwise returns an error.
 func (b *backend) HasOngoingSwapAsTaker(remotePeer peer.ID) error {
-	swaps, err := b.swapManager.GetOngoingSwaps()
+	swaps, err := b.swapManager.GetOngoingSwapsSnapshot()
```
dimalinux, line 265: This was another change that I made when I was still working on a design that kept the status channel inside the `swap.Info` structure, but it seems worth keeping. Methods that return deep copies and are therefore completely thread safe, have the word `Snapshot` on the end.

protocol/swap/status_manager.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/athanorlabs/atomic-swap/common/types"
+)
+
+// statusManager provides lookup for the status channels. Status channels are
+// ephemeral between runs of swapd.
+type statusManager struct {
+	mu             sync.Mutex
+	statusChannels map[types.Hash]chan Status
+}
```
dimalinux, line 14: I isolated the status channel management into this separate `statusManager` "class" and gave `swap.manager` an "is-a" relationship with it. The base swap manager was already using its mutex across multiple independent maps. Separating this, hopefully, lowers the code complexity.

protocol/swap/status_manager.go
```diff
@@ -0,0 +1,63 @@
...
+	return sm.getStatusChan(offerID)
+}
+
+// DeleteStatusChan deletes any status channel associated with the offer ID.
+func (sm *statusManager) DeleteStatusChan(offerID types.Hash) {
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+
+	delete(sm.statusChannels, offerID)
+}
```
dimalinux, line 49: No one is using this yet and I'm not entirely clear what the lifecycle should be. For `net_makeOfferAndSubscribe`, I think you want it to live across multiple takers if the first one fails. If we can figure out the lifecycle before merging, that might be good, but I don't think anyone is going to run enough swaps without restarting for this to budge resource usage noticeably.

protocol/xmrmaker/message_handler.go
```diff
@@ -72,7 +72,7 @@ func (s *swapState) setNextExpectedEvent(event EventType) error {
 		panic("status corresponding to event cannot be UnknownStatus")
 	}
 
-	s.info.SetStatus(status)
+	s.UpdateStatus(status)
 	err := s.Backend.SwapManager().WriteSwapToDB(s.info)
```
dimalinux, line 76: When do we write the the DB after updating the status and when do we not do it?

rpc/net.go
```diff
@@ -37,12 +38,12 @@ type NetService struct {
 	net        Net
 	xmrtaker   XMRTaker
 	xmrmaker   XMRMaker
-	sm         SwapManager
+	sm         swap.Manager
```
dimalinux, line 41: Nothing here. I just removed the type alias.

protocol/swap/status_manager.go
noot: I would expect it to be deleted when the swap ends (no matter the status) and if it's retaken, then a new channel would be created

protocol/swap/swap_manager.go
```diff
@@ -140,40 +147,86 @@ func (m *manager) GetPastSwap(id types.Hash) (*Info, error) {
...
-		swaps[i] = sCopy
-		i++
+		offerIDs = append(offerIDs, &s.OfferID)
 	}
+
+	return offerIDs, nil
+}
+
+// GetOngoingSwapsSnapshot returns a copy of all ongoing swaps. If you need to
+// modify the result or read from its status channel, call `GetOngoingSwap` on
```
noot, line 204: `Info` no longer has a status channel?

protocol/swap/swap_manager.go
```diff
@@ -140,40 +147,86 @@ func (m *manager) GetPastSwap(id types.Hash) (*Info, error) {
...
+	s, has := m.ongoing[offerID]
+	if !has {
+		return nil, errNoSwapWithOfferID
+	}
+
+	// We have the read lock above when making this copy, but the swapState
+	// instances don't use that lock when they modify the same Info structure
+	// from a different go process. At some point we should update the swapState
+	// instances to make all changes to their ongoing Info structure via this
+	// manager.
```
noot, line 180: we should open an issue for this; ideally we don't need the snapshot functions and it's only copies that are ever returned, and the other processes modify the swap info (I think status is the only field modified after creation?) via this manager

protocol/swap/status_manager.go
```diff
@@ -0,0 +1,63 @@
...
+
+// PushNewStatus adds a new status to the offer ID's channel
+func (sm *statusManager) PushNewStatus(offerID types.Hash, status types.Status) {
```
noot, line 52: we should remove `info.SetStatus` and make a function on the swapManager that updates the status that also calls this function

protocol/xmrtaker/instance.go
```diff
@@ -64,12 +64,17 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
 	if err != nil {
 		return err
 	}
 
-	for _, s := range swaps {
+	for _, offerID := range ongoingIDs {
+		s, err := inst.backend.SwapManager().GetOngoingSwap(*offerID)
+		if err != nil {
+			return err
+		}
```
noot, line 76: why is this changed? can you add a function to get all ongoing swaps like before?

protocol/swap/swap_manager.go
dimalinux: Correct. There is some information in the PR description as to why.

protocol/xmrtaker/instance.go
dimalinux: `GetOngoingSwapsSnapshot` returns all the swaps (as a deep copy) and it is still there. I thought the code was easier to audit if we just got the IDs here and later pull the non-deep-copies on an individual bases when restarting a swap.

protocol/swap/swap_manager.go
dimalinux: > I think status is the only field modified after creation?

`EndTime`, `LastStatusUpdateTime`, `Timeout1` and `Timeout2` are also modified after creation.

protocol/swap/status_manager.go
dimalinux: > I would expect it to be deleted when the swap ends (no matter the status) and if it's retaken, then a new channel would be created

It is handled that way now.

protocol/swap/swap_manager.go
dimalinux: > we should open an issue for this

The locking is now in this PR, so we don't need to open an issue. I removed the comment above. The locking in here is exclusively for the map maintained by the manager. The lock for the `Info` structure is inside the struct and used by member functions of the struct.

protocol/swap/status_manager.go
dimalinux: See what I implemented and if you like it. I was originally thinking along the lines of what you wrote above, but I decided that this manager should only deal with maintaining the map of the status channel (which has a different lifecycle than the `Info` struct). Locking and changing the `Info` struct is handled by member functions that are invoked by the maker/taker `swapStage` go-processes.

protocol/swap/swap_manager.go
noot: yeah I meant you need to update the comment here

protocol/swap/types.go
```diff
@@ -63,8 +62,18 @@ type Info struct {
...
+	// rwMu handles synchronization when LastStatusUpdateTime, Timeout1,
+	// Timeout2 and EndTime are updated. This Info struct is modified by the
+	// maker or taker's swapState go process as the state of the swap
+	// progresses. The swapState go process does not need synchronization when
+	// reading its own changes, but it needs to grab a write lock when modifying
+	// the the structure. Readers from other go-processes only get copies of
+	// this structure. They exclusively use the DeepCopy method to get their
+	// copy, which grabs the read lock ensuring that they always capture the
+	// up-to-date state of this Info struct.
+	rwMu sync.RWMutex
```
noot, line 76: why did you decide to go with a lock in the info struct instead of having the swap manager `Get` functions be read-only (ie. only return copies) and having swap manager implement `Set` functions which modify the info fields for a given swap?

protocol/xmrmaker/instance.go
```diff
@@ -84,12 +84,17 @@ func NewInstance(cfg *Config) (*Instance, error) {
...
 	if err != nil {
 		return err
 	}
 
-	for _, s := range swaps {
+	for _, offerID := range ongoingIDs {
+		s, err := inst.backend.SwapManager().GetOngoingSwap(*offerID)
+		if err != nil {
+			return err
+		}
```
noot, line 96: why two iterations? why not do `GetOngoingSwapsSnapshot` directly?

protocol/xmrmaker/swap_state.go
```diff
@@ -452,6 +450,11 @@ func (s *swapState) SendKeysMessage() common.Message {
 	}
 }
 
+func (s *swapState) UpdateStatus(status types.Status) {
```
noot, line 453: doesn't need to be exported I don't think

protocol/xmrtaker/instance.go
noot: I don't think it makes a difference and this results in 2 iterations instead of 1, and we're iterating through all of them either way

noot: maybe implement a `GetOngoingSwaps` that returns all the swaps not as a copy?

protocol/xmrtaker/swap_state.go
```diff
@@ -318,6 +317,11 @@ func (s *swapState) SendKeysMessage() common.Message {
 	}
 }
 
+func (s *swapState) UpdateStatus(status types.Status) {
```
noot, line 320: doesn't need to be exported

noot (commented):
basically looks good, mostly minor comments, and an overall question about why the setters were left on `Info` instead of `swap.Manager`. I'm not sure if designing it like that would be nicer or worse :/

protocol/swap/types.go
dimalinux: The swap manager's existing locks are for the maps inside the manager. They act globally across all `Info` objects in the ongoing or past swap maps. To manage the state the state of an individual swap, we needed a lock that is specific to that individual `Info` object. The swap manager could maintain a map of locks for the individual `Info` structures, but I thought this solution was cleaner.

protocol/xmrmaker/message_handler.go
dimalinux: @noot Did you see this question? It's not critical to the PR. I just wanted to understand it better.

protocol/swap/swap_manager.go
dimalinux: Now I understand. I removed the part about the status channel from the comment. Thanks.

protocol/xmrtaker/instance.go
dimalinux: I downgraded the code from getting the full, modifiable object list here, because there is nothing at this high level in the code that needs access to anything but the offer IDs. By getting the modifiable copies lower down on the stack on an individual basis, it shortens the chain-of-custody for the modifiable objects and reduces the API surface area that returns modifiable objects. Auditing synchronous access to objects is hard and I was hoping these changes make the code easier to audit for correct use of locks.

protocol/xmrmaker/instance.go
dimalinux: See the comment on the maker. If, after reading it, you still want the change, I can make it.

protocol/xmrmaker/swap_state.go
dimalinux: Fixed!

protocol/xmrtaker/swap_state.go
dimalinux: Fixed!

noot (approved):

PR #472, opened 2023-05-17
unified RPC/websockets client

dimalinux:
Provides a united client for both the RPC and websocket connections. Some redundant API methods that were available over both ws and rpc were consolidated to just the RPC versions.

This PR is currently branched off of [dimalinux/status-subscribe-after-restart](https://github.com/AthanorLabs/atomic-swap/tree/dimalinux/status-subscribe-after-restart). Assuming that PR is merged first, I'll rebase this PR off of the master branch. Alternatively, we could merge this PR into the other one.


protocol/swap/status_manager.go
```diff
@@ -50,7 +50,14 @@ func (sm *statusManager) DeleteStatusChan(offerID types.Hash) {
...
 func (sm *statusManager) PushNewStatus(offerID types.Hash, status types.Status) {
-	sm.getStatusChan(offerID) <- status
+	ch := sm.getStatusChan(offerID)
+	ch <- status
+	// If the status is not ongoing, existing subscribers will get the status
+	// via the channel since they already have a reference to it. New
+	// subscribers will get the final status from the past swaps map.
+	if !status.IsOngoing() {
+		sm.DeleteStatusChan(offerID)
+	}
```
dimalinux, line 60: TODO: Push these changes into [the parent PR](https://github.com/AthanorLabs/atomic-swap/pull/471).

rpcclient/client.go
```diff
@@ -38,24 +38,26 @@ var (
...
 // NewClient creates a new JSON-RPC client for the specified endpoint. The passed context
 // is used for the full lifetime of the client.
-func NewClient(ctx context.Context, endpoint string) *Client {
+func NewClient(ctx context.Context, port uint16) *Client {
 	return &Client{
-		ctx:      ctx,
-		endpoint: endpoint,
+		ctx:        ctx,
+		endpoint:   fmt.Sprintf("http://127.0.0.1:%d", port),
+		wsEndpoint: fmt.Sprintf("ws://127.0.0.1:%d/ws", port),
```
dimalinux, line 52: I didn't want to complicate this PR, but if we want to in the future, I could put the websocket methods at the root too. It is possible to share.

rpcclient/mocks_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2023 The AthanorLabs/atomic-swap Authors
 // SPDX-License-Identifier: LGPL-3.0-only
 
-package rpc
+package rpcclient
```
dimalinux, line 4: I moved the tests, mostly unchanged, from `rpc` to `rpcclient` to avoid a circular dependency.

common/rpctypes/types.go
```diff
@@ -17,9 +17,6 @@ import (
 
 // JSON RPC method names that we serve on the localhost server
 const (
-	NetDiscover         = "net_discover"
-	NetQueryPeer        = "net_queryPeer"
-	SubscribeNewPeer    = "net_subscribeNewPeer"
 	SubscribeMakeOffer  = "net_makeOfferAndSubscribe"
 	SubscribeTakeOffer  = "net_takeOfferAndSubscribe"
```
dimalinux, line 21: If we want to maintain/test less code, I don't see how these 2 methods are any different than calling their `http` equivalent and then immediately calling `swap_subscribeStatus`. There is no race condition. If a status is added to the channel between the http and ws requests, it will still be waiting for the user when the 2nd request starts.

rpc/ws.go
```diff
@@ -90,42 +90,7 @@ func (s *wsServer) handleRequest(conn *websocket.Conn, req *rpctypes.Request) er
 		}
 
 		return s.handleSigner(s.ctx, conn, params.OfferID, params.EthAddress, params.XMRAddress)
-	case rpctypes.SubscribeNewPeer:
-		return errUnimplemented
```
dimalinux, line 94: Not sure what this was supposed to do, but I assume we can just add it back later if we decide that we need it.

common/rpctypes/types.go
noot: yeah, we can probably remove these in favour of the above, feel free to open an issue for it

noot (approved):
this is really nice, thanks!

common/rpctypes/types.go
dimalinux: I'll just leave it for now. It is not as simple as I though to combine them when I wrote that message. In its current implementation, `swap_subscribeStatus` assumes the swap is currently ongoing, but that is not the case when the maker listens for state changes on the offer he just made.

PR #473, opened 2023-05-18
stagenet bootnode and docker updates

dimalinux:
The libp2p keys of some of the stagenet bootnodes changed. This updates to the current values.

Our docker build scripts now target the latest release by grabbing the latest version outside of the `Dockerfile`.  When docker analyzes each command in the `Dockerfile` and decides whether it can use a cached execution, `latest` won't always give you the `latest`. I could have disabled caching in the docker build, but by changing the scripts to build a specific version, docker will handle the caching correctly.


cmd/swapcli/main.go
```diff
@@ -1096,7 +1096,10 @@ func runGetVersions(ctx *cli.Context) error {
 	fmt.Printf("swapd: %s\n", resp.SwapdVersion)
 	fmt.Printf("p2p version: %s\n", resp.P2PVersion)
 	fmt.Printf("env: %s\n", resp.Env)
-	fmt.Printf("swap creator address: %s\n", resp.SwapCreatorAddr)
+	// Bootnodes don't have a contract address
+	if resp.SwapCreatorAddr != nil {
+		fmt.Printf("swap creator address: %s\n", resp.SwapCreatorAddr)
+	}
```
dimalinux, line 1102: Before this change it was printing `swap creator address: <nil>` for bootnodes. Now it just won't print that line for bootnodes.

common/config.go
```diff
@@ -101,11 +101,10 @@ func StagenetConfig() *Config {
 		},
 		SwapCreatorAddr: ethcommon.HexToAddress("0xbf2B7a6dCE5598Cf002B3507a8D62cf2C35cE5c6"),
 		Bootnodes: []string{
-			"/ip4/134.122.115.208/tcp/9900/p2p/12D3KooWDqCzbjexHEa8Rut7bzxHFpRMZyDRW1L6TGkL1KY24JH5",
-			"/ip4/143.198.123.27/tcp/9900/p2p/12D3KooWSc4yFkPWBFmPToTMbhChH3FAgGH96DNzSg5fio1pQYoN",
-			"/ip4/67.207.89.83/tcp/9900/p2p/12D3KooWLbfkLZZvvn8Lxs1KDU3u7gyvBk88ZNtJBbugytBr5RCG",
-			"/ip4/134.122.115.208/tcp/9900/p2p/12D3KooWDqCzbjexHEa8Rut7bzxHFpRMZyDRW1L6TGkL1KY24JH5",
-			"/ip4/67.205.131.11/tcp/9900/p2p/12D3KooWT19g8cfBVYiGWkksU1ZojHCBNqTu3Hz5JLfhhytaHSwi",
```
dimalinux, line 108: Lines 104 and 107 were identical, so there is one less line in the updated version.

noot (approved):

PR #474, opened 2023-05-20
add relayer fee to `swapcli past`

### LINKED ISSUE #419: `swapcli past` should subtract relayer fee if needed
```
$ ./bin/swapcli past
Past swaps:
ID: 0xabdd2c94ccda33d8a86a1bdcbd5f229eeea29f477c55a2d44ac5291a391b5d13
Start time: 2023-04-25-14:23:37
End time: 2023-04-25-15:10:52
Provided: 0.105813953488 XMR
Received: 0.0091 ETH
Exchange Rate: 0.086 ETH/XMR
Status: Success
```

This is the current output of `swapcli past`, however, the `Received` amount does not subtract the relayer fee. we should leave a note that says `Received: 0.0091 ETH (minus 0.009 ETH relayer fee)` or something like that.

stubbrn:
closes #419

`swapcli past` looks like this:

```
ID: 0xfcef3b9908284a607b9fef10c86aa82aa0f0e86cb100b9a0c665a01088a321bd
Start time: 2023-05-21-01:00:57
End time: 2023-05-21-01:01:14
Provided: 0.1 XMR
Received: 0.1 ETH (minus 0.01 ETH relayer fee)
Exchange Rate: 1 ETH/XMR
Status: Success
```


dimalinux:
> Received: 0.1 ETH (minus 0.01 ETH relayer fee)

Question: Do we want to write it like this, or do we want to write the actual amount received, then in parenthesis write the amount paid in relayer fees? I don't think it is critical, but I'd lean towards the exact amount received being a more helpful value. Do we want the fees in parentheses or as a separate line item?

Additional questions: For the taker, do we want to summarize what they paid in ETH fees across the newSwap/setReady transactions?  Do we want to track the relatively low XMR fees? The XMR fees may be low enough that we don't care, not sure.

Whatever the answers, I'm not suggesting that they all make it into this PR.  But I think we should answer all the questions to know what smaller steps take us in the right direction.

protocol/xmrmaker/claim.go
```diff
@@ -235,8 +235,20 @@ func (s *swapState) claimWithRelay() (*ethtypes.Receipt, error) {
...
 		log.Infof("falling back to swap counterparty as relayer")
-		return s.relayClaimWithXMRTaker()
+		receipt, err = s.relayClaimWithXMRTaker()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Save the relayer fee to the database
+	s.info.RelayerFee = coins.RelayerFeeWei
```
dimalinux, line 245: The Info data structure is accessed by multiple threads. This go process executing the code here is the only thread that changes the data structure, so it doesn't require a read locks to always get correctly synchronized data, but it needs to grab the write lock (introduce recently in https://github.com/AthanorLabs/atomic-swap/pull/471) before changing data.  Create a setter function on the Info type that grabs the lock for writing before updating the value.  Other threads only read the data structure via `DeepCopy`, which grabs the read lock, so no additional changes are required.

dimalinux: If you use `RelayerFeeETH` above instead of `RelayerFeeWei`, you might make some math with the expected amount (which is in ETH) easier.

cmd/swapcli/main.go
```diff
@@ -900,7 +900,11 @@ func runGetPastSwap(ctx *cli.Context) error {
 		fmt.Printf("Start time: %s\n", info.StartTime.Format(common.TimeFmtSecs))
 		fmt.Printf("End time: %s\n", endTime)
 		fmt.Printf("Provided: %s %s\n", info.ProvidedAmount.Text('f'), providedCoin)
-		fmt.Printf("Received: %s %s\n", info.ExpectedAmount.Text('f'), receivedCoin)
+		fmt.Printf("Received: %s %s", info.ExpectedAmount.Text('f'), receivedCoin)
```
dimalinux, line 903: @stubbrn If you agree with the message that I wrote earlier that the actual received amount would be more helpful here, here's how you could calculate it:
```go
		receivedAmt := new(apd.Decimal)
		relayerFeeETH := coins.NewWeiAmount(info.RelayerFee).AsEther()
		_, err = coins.DecimalCtx().Sub(receivedAmt, info.ExpectedAmount, relayerFeeETH)
		if err != nil {
			return err
		}
```
Note: My other comment about storing the relayer fee in ETH instead of WEI would eliminate the 2nd line above.

noot:
@stubbrn yeah, I think putting the exact amount received makes sense, but it would be good to mention `received amount = (before fee swap amount - relayer fee)` somewhere so people aren't asking where the extra 0.01 ETH went lol.

cmd/swapcli/main.go
```diff
@@ -896,11 +896,27 @@ func runGetPastSwap(ctx *cli.Context) error {
 			endTime = info.EndTime.Format(common.TimeFmtSecs)
 		}
 
+		receivedAmt := new(apd.Decimal)
+		*receivedAmt = *info.ExpectedAmount
+		if info.RelayerFee != nil {
+			_, err = coins.DecimalCtx().Sub(receivedAmt, info.ExpectedAmount, info.RelayerFee)
```
dimalinux, line 902: Line 200 is probably safe, but I try not to make assumptions about libraries unless they use copy-by-value in their documentation. The copy isn't necessary:
```go
		receivedAmt := info.ExpectedAmount
		if info.RelayerFee != nil {
			receivedAmt := new(apd.Decimal)
			_, err = coins.DecimalCtx().Sub(receivedAmt, info.ExpectedAmount, info.RelayerFee)
```

dimalinux (approved):
Let's get rid of that apd-decimal copy-by-value (see new comment) before merging, but otherwise it looks good.

PR #475, opened 2023-05-21
redesign

stubbrn:
Here is what I think a better baseline for the UI.

I originally wanted to do just a redesign, but I found the `smui` library to be difficult to work with, so I replaced it with `flowbite-svelte` instead.

I've also added `vite`, so now each time you change something it doesn't take multiple seconds to see the results. 

preview

I think it makes more sense for Pairs to be separated (1 Table/Order-Book = 1 Pair), that's why there's a "ETH / XMR" indicating the current pair (Of course for now that's just a mock-up).


ui/src/utils/rpcApi.ts
```diff
@@ -8,7 +8,7 @@ export type JSONRPCResult<Data> = {
 }
 
 export const getPort = () : number => {
-    const port = Number(process.env.SWAPD_PORT)
+    const port = Number(import.meta.env.SWAPD_PORT)
     return isNaN(port) ? 5001 : port
```
noot, line 12: can you change this to 5000 (which is now the default port)?

noot (commented):
overall, I really like the design! the peer count in the corner is a nice addition. thanks so much for taking this on, definitely needed since I don't really do ui stuff and neither does @dimalinux afaik. 

general thoughts/comments:
- would be good to show the coin in the min/max columns, and also show the min/max amounts for both XMR and ETH (or put that in the swap pop-up) since right now it requires some estimation to figure out the ETH amount
- what are your plans for the pairs/wallet/docs sections at the top? specifically thinking about wallets, I suppose for pairs we can eventually show ERC20/XMR pairs if offers for those exist
- if I try to take a swap offer, I get a websocket closed error, is that expected?

noot:
> Well, wallets are handled by swapd, so it might make sense to implement a UI to list wallets and manage them easily.
Could also be useful to show the user's swap history.

makes sense, it could be useful to use the `swapcli transfer-(eth/xmr)` functions in the wallets tab so users can easily transfer the funds out

> Not sure how to reproduce your error, it's supposed to show up only when the socket closed during the swap.
This is what is supposed to happen:

yeah the error was my bad! it's working for me now :)

ui/src/lib/Identicon.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
+    import Identicon from 'identicon.js'
+  
+    export let size: number = 30
+    export let peerAddress: string = ''
+    var options = {
+      //foreground: [0, 0, 0, 255],               // rgba black
```
noot, line 7: can you remove this?

noot (approved):
looks good to me! it would be good to get the metamask integration working again, the previous flow was that swapd is started with the `--external-signer` flag, and in the UI, the WS method `signer_subscribe` is called which pushes messages to be signed, which are then signed by the user in metamask, and the resulting signed tx is returned to swapd. however, I suspect at least one if not multiple of these features are broken since they haven't really been maintained. can open an issue for this and try to test these features in the meantime!

PR #478, opened 2023-05-28
additonal erc20 and taker decimal checks

dimalinux:
When making an offer between XMR and ETH, it is impossible for the maker to propose min/max values in XMR that can not be represented in whole WEI units. This is because XMR is limited to 12 decimal places and the exchange rate is capped at 6 decimal places. This doesn't prevent all problems though. The taker can chose a value between the maker's min/max values that, when converted to ETH, result in fractional piconeros when divided by the exchange rate. We could allow rounding in this case (not done in this PR), as long as our code never expects `provided = ToETH(ToXMR(provided))`. This inverse-equal expectation could happen if, for example, the taker looks at the maker-locked XMR balance, converts it back to ETH and then compares the converted amount. If the piconero amount was rounded down for the locked amount, when converting back to ETH the value will be less than what the taker provided. In this PR, we just restrict taker values that could result in fractional piconero calculations to avoid these subtle rounding issues.

When making an offer between XMR and an ER20 token, even the existing maker limitations were insufficient. Some tokens, mainly stablecoins, only have 6 decimal places. That means that a maker could provide min/max values that, when converted into the token, would be more precise than the token supports. If rounding is used (not done in this PR), we'd need to be careful that min values are rounded up and max values are rounded down in the token units. Otherwise, the reverse calculation from the rounded token units leads to out-of-offer-range values when converting back to XMR.

In this PR we restrict all min/max amounts to values that are fully reversible: `x = ToXMR(ToEth(x))`.  We also restrict all taker provided amounts to values that are fully reversible: `y = ToETH(ToXMR(y))`.

If the taker tries to provide an amount that is not fully reversible, we include the closest possible alternative value that is reversible in the error message, as computing this value would be non-trivial for end users to do on their own.


cmd/swapcli/suite_test.go
```diff
@@ -28,7 +31,10 @@ func TestRunSwapcliWithDaemonTests(t *testing.T) {
 	s.conf = daemon.CreateTestConf(t, tests.GetMakerTestKey(t))
 	t.Setenv("SWAPD_PORT", strconv.Itoa(int(s.conf.RPCPort)))
 	daemon.LaunchDaemons(t, 10*time.Minute, s.conf)
-	s.mockTokens = daemon.GetMockTokens(t, s.conf.EthereumClient)
+	ec := s.conf.EthereumClient.Raw()
+	pk := s.conf.EthereumClient.PrivateKey()
+	s.mockTether = contracts.GetMockTether(t, ec, pk)
+	s.mockDAI = contracts.GetMockDAI(t, ec, pk)
```
dimalinux, line 37: This PR moves the mock tokens out of the `daemon` package and into the `contracts` package to make them available to more test packages without creating circular dependencies.

coins/coins.go
```diff
@@ -305,11 +305,11 @@ func NewERC20TokenAmount(amount int64, tokenInfo *ERC20TokenInfo) *ERC20TokenAmo
 	}
 }
 
-// NewERC20TokenAmountFromDecimals converts some amount for a token in its
+// NewTokenAmountFromDecimals converts some amount for a token in its
```
dimalinux, line 308: Just shortened the name, since it was extra long.

common/types/offer.go
```diff
@@ -152,14 +152,18 @@ func (o *Offer) validate() error {
 	// can't be used to complete the swap if the maker does not have sufficient
 	// ETH to make the claim themselves. While relayers are not used with ERC20
 	// swaps, we still want a minimum swap amount, so we use the same value.
-	relayerFeeAsXMR, err := o.ExchangeRate.ToXMR(coins.RelayerFeeETH)
+	minAmtAsETH, err := o.ExchangeRate.ToETH(o.MinAmount)
```
dimalinux, line 155: We've already validated that the exchange rate has no more than 6 decimal places and the min amount has no more than 12, so converting to ETH (a multiply) will always result in a non-rounded ETH value. The original calculation here was dividing the relayer fee by the exchange rate which can result in fractional piconeros.

daemon/bad_make_take_values_test.go
```diff
@@ -0,0 +1,55 @@
...
+	require.NoError(t, err)
+
+	// Fail because providesAmount has too much precision in the token's standard units
+	err = ac.TakeOffer(makeResp.PeerID, makeResp.OfferID, providesAmt)
+	require.ErrorContains(t, err, `"net_takeOffer" failed: "providesAmount" has too many decimal points; found=7 max=6`)
+
+	// Fail because the providesAmount has too much precision when converted into XMR
+	// 20.123456/13.3 = 1.51304[180451127819548872] (bracketed sequence repeats forever)
+	providesAmt = coins.StrToDecimal("20.123456")
+	err = ac.TakeOffer(makeResp.PeerID, makeResp.OfferID, providesAmt)
```
dimalinux, line 53: I found a bug in our client RPC handling thanks to this test that you'll see later down. There is a `TakeOfferRequest` type, but no expected response type. When there was no expected response type, we weren't decoding the server's JSON-RPC response and getting returned errors. The code here was always getting `nil` even when the server returned an error. We were most likely seeing the same error when @stubbrn did the Shutdown RPC PR.

coins/coins.go
```diff
@@ -283,71 +283,68 @@ func (t *ERC20TokenInfo) SanitizedSymbol() string {
 
 // ERC20TokenAmount represents some amount of an ERC20 token in the smallest denomination
 type ERC20TokenAmount struct {
-	Amount    *apd.Decimal    `json:"amount" validate:"required"` // in smallest non-divisible units of token
+	Amount    *apd.Decimal    `json:"amount" validate:"required"` // in standard units
```
dimalinux, line 286: I was getting confused by this, and I don't think I was the only one. In other places where we represent non-standard units using `*apd.Decimal`, we are using a type alias. This was the only place where we were declaring the field as `*apd.Decimal` and it was not in standard units. And the name of the field didn't clarify it either. I considered creating a new type alias or renaming the field, but just storing the value in standard units seemed cleaner.

daemon/swap_daemon_erc20_test.go
```diff
@@ -29,15 +30,17 @@ func TestRunSwapDaemon_ExchangesXMRForERC20Tokens(t *testing.T) {
...
+	tokenAsset := getMockTetherAsset(t, aliceConf.EthereumClient)
 
 	timeout := 7 * time.Minute
 	ctx, _ := LaunchDaemons(t, timeout, aliceConf, bobConf)
 
 	bc := rpcclient.NewClient(ctx, bobConf.RPCPort)
 	ac := rpcclient.NewClient(ctx, aliceConf.RPCPort)
 
+	bobStartTokenBal, err := bobConf.EthereumClient.ERC20Balance(ctx, tokenAsset.Address())
+	require.NoError(t, err)
```
dimalinux, line 42: Now that we have more token tests and are caching the token contract between tests, we can't assume that Bob's token balance is zero on start.

rpc/net.go
```diff
@@ -190,7 +190,7 @@ func (s *NetService) takeOffer(makerPeerID peer.ID, offerID types.Hash, provides
 
 	swapState, err := s.xmrtaker.InitiateProtocol(makerPeerID, providesAmount, offer)
 	if err != nil {
-		return fmt.Errorf("failed to initiate protocol: %w", err)
```
dimalinux, line 193: At least with the error samples I was looking at (see `TestBadMakeTakeValues` in the `daemon` package), this was making the error message longer/harder to read without adding helpful context. This particular error is passed to the RPC client who was issuing a TakeOffer request, so it is a more user facing error than the ones we log.

protocol/ethereum_asset_amount.go
```diff
@@ -22,13 +22,23 @@ func GetEthAssetAmount(
 	amt *apd.Decimal, // in standard units
 	asset types.EthAsset,
 ) (coins.EthAssetAmount, error) {
```
dimalinux, line 24: I'm not sure what to do with this function. It never had direct unit tests and the only caller left is in the `xmrmaker`'s `checkContract` function (different package). It was no longer a good fit in the other 2 places where it was being called, as both of them required the token info well before creating this ETH asset amount.

coins/round.go
```diff
@@ -5,18 +5,107 @@ package coins
 
 import (
 	"github.com/cockroachdb/apd/v3"
+	"github.com/ethereum/go-ethereum/common/math"
 )
 
-func roundToDecimalPlace(result *apd.Decimal, n *apd.Decimal, decimalPlace uint8) error {
-	result.Set(n) // already optimizes result == n
+func roundToDecimalPlace(n *apd.Decimal, decimalPlace uint8) (*apd.Decimal, error) {
```
dimalinux, line 11: The previous version had a defect in that it assumed that the `result` was not the same memory address as `n` (the number being rounded). We don't care about optimizing the number of apd.Decimal's that are allocated, so I just simplified things and changed the API to always return a freshly allocated value.

common/rpctypes/jsonrpc.go
```diff
@@ -39,5 +39,8 @@ type Error struct {
 
 // Error ...
 func (e *Error) Error() string {
-	return fmt.Sprintf("message=%s; code=%d; data=%v", e.Message, e.ErrorCode, e.Data)
+	if e.ErrorCode != 0 {
+		return fmt.Sprintf("message=%s; code=%d; data=%v", e.Message, e.ErrorCode, e.Data)
+	}
+	return e.Message
```
dimalinux, line 45: In this PR we are returning information important to the end user in the error returned from the server, so I'm trying to remove unhelpful wrapped verbiage that makes it harder for the end user to see the helpful portion of the message.

coins/coins.go
```diff
@@ -220,19 +234,24 @@ func (a *WeiAmount) AsEtherString() string {
 	return a.AsEther().Text('f')
 }
 
-// AsStandard is an alias for AsEther, returning the Wei amount as ETH
-func (a *WeiAmount) AsStandard() *apd.Decimal {
+// AsStd is an alias for AsEther, returning the Wei amount as ETH
+func (a *WeiAmount) AsStd() *apd.Decimal {
```
noot, line 238: why the rename?

coins/exchange_rate_test.go
```diff
@@ -13,41 +13,34 @@ import (
...
 	rate := StrToExchangeRate("0.333333")
-	ethAmount := StrToDecimal("3.1")
+	ethAssetAmt := StrToETHAsset("3.1", nil)
 
-	// 3.1/0.333333 calculated to 13 decimals is 9.3000093000093 (300009 repeats indefinitely)
-	// This calculator goes to 200 decimals: https://www.mathsisfun.com/calculator-precision.html
-	// XMR rounds at 12 decimal places to:
-	const expectedXMRAmount = "9.300009300009"
+	_, err := rate.ToXMR(ethAssetAmt)
+	expectedErr := "3.1 ETH / 0.333333 exceeds XMR's 12 decimal precision, try 3.099999999999899997"
```
noot, line 29: am I wrong or does 3.1/0.333333 have only 11 decimal places? i appears to be 9.30000930001

docs/mainnet.md
```diff
@@ -4,7 +4,7 @@ The swap operates on a maker/taker paradigm, where one party is a "market maker"
 
 Currently, due to protocol limitations, XMR holders who want ETH can only act as makers, and ETH holders who want XMR can only act as takers.
 
-> **Note:** a swap on mainnet currently takes around 20-25 minutes due to block time.
+> **Note:** a swap on mainnet currently take around 20-25 minutes due to block time.
```
noot, line 7: change this back?

docs/stagenet.md
```diff
@@ -1,8 +1,9 @@
 # Joining the Stagenet/Sepolia network
 
-Currently, an initial version of the swap is deployed onto the Sepolia (Ethereum testnet) and Monero Stagenet networks. To join the network and try out the swap, either as a maker or a taker, please see the following.
+Swaps can be performed on the Sepolia (Ethereum testnet) together with the
+Monero Stagenet network. This document describes how to do stagenet swaps.
 
-> Note: a swap on stagenet currently takes around 10-20 minutes due to block times.
+> Note: stagenet and mainnet swaps currently takes around 20 minutes due to the 2-minute block time and required 10 confirmations before funds can be spent.
```
noot, line 6: I think you wanted to change the "takes" here

noot: also why mention mainnet here?

protocol/ethereum_asset_amount.go
noot: you can move it to the xmrmaker package perhaps?

protocol/xmrmaker/net.go
```diff
@@ -110,44 +110,48 @@ func (inst *Instance) HandleInitiateMessage(
...
-	// TODO: If this is not ETH, we need quick/easy access to the number
-	//       of token decimal places. Should it be in the OfferExtra struct?
-	err := coins.ValidatePositive("providedAmount", coins.NumEtherDecimals, msg.ProvidedAmount)
+	offer, offerExtra, err := inst.offerManager.GetOffer(msg.OfferID)
 	if err != nil {
 		return nil, err
 	}
 
-	offer, offerExtra, err := inst.offerManager.GetOffer(msg.OfferID)
+	var maxDecimals uint8 = coins.NumEtherDecimals
```
noot, line 118:
```suggestion
	maxDecimals := coins.NumEtherDecimals
```

protocol/xmrtaker/net.go
```diff
@@ -7,47 +7,64 @@ import (
...
 // The input units are ether that we will provide.
 func (inst *Instance) InitiateProtocol(
 	makerPeerID peer.ID,
 	providesAmount *apd.Decimal,
 	offer *types.Offer,
 ) (common.SwapState, error) {
-	err := coins.ValidatePositive("providesAmount", coins.NumEtherDecimals, providesAmount)
-	if err != nil {
-		return nil, err
+	var maxDecimals uint8 = coins.NumEtherDecimals
```
noot, line 29: same here

coins/coins.go
dimalinux: This PR added `AsStandardString` to the `EthAssetAmount` interface's API, and that original name seemed too long.  Just using `String()` was causing issues, as `WeiAmount` implements the interface, but doesn't default to standard units.  Using `amt.AsStandard().Text('f')` was also too long and prone to developer mistakes or non-use.  STD is an acronym that most people are familiar with and wouldn't be confused with anything else, so I thought it was a good place to use it. `.AsStandard()` would be short enough without the acronym, but I wanted all the functions on the `EthAssetAmount` interface to be consistent in their naming. Having `AsStdString()` and `AsStandard()` on the same interface seemed inconsistent.

noot (approved):
looks good, only minor comments!

docs/mainnet.md
dimalinux: The noun in this sentence, `swap`, is singular. We could make it plural. I was just fixing the grammar with minimal changes.

docs/stagenet.md
dimalinux: I'll change this one to `take`

docs/mainnet.md
dimalinux: Will fix!

protocol/xmrmaker/net.go
dimalinux: `NumEtherDecimals` is not-typed constant. Would you prefer `maxDecimals := uint8(coins.NumEtherDecimals)` over the way it is now?

dimalinux: I switched it and the other identical one to use a cast on the non-typed constant instead of a `var` declaration.

protocol/ethereum_asset_amount.go
dimalinux: done

Issue #479, opened 2023-05-29
TestXMRTakerCancelOrRefundAfterKeyExchange flaky


dimalinux:
The recently added test, `TestXMRTakerCancelOrRefundAfterKeyExchange`, is flaky. The problem is this code:
```go
			count, err := ec.PendingTransactionCount(ctx) //nolint:govet
			if err != nil {
				t.Errorf("failed to get pending tx count: %s", err)
				return
			}

			if count > 0 {
				// the newSwap tx is in the mempool, shut down the nodes
				cancel()
				t.Log("cancelling context of Alice's and Bob's servers")
				return
			}
```
Multiple packages are being tested concurrently and the pending transaction that is found is sometimes an unrelated transaction that happens before the `newSwap` transaction that it is looking for is created. Then the `newSwap` transaction never gets created because the context is cancelled first. The swap ends up with the status of `aborted`.

noot:
closed by #480 ?

PR #480, opened 2023-06-02
fix TestXMRTakerCancelOrRefundAfterKeyExchange

dimalinux:
* Fixes TestXMRTakerCancelOrRefundAfterKeyExchange, closing https://github.com/AthanorLabs/atomic-swap/issues/479
* Updates dependencies
* Changes github actions to cancel previous started job when a new commit comes in to the identical workflow


.github/workflows/build-macos.yml
```diff
@@ -1,11 +1,15 @@
...
   pull_request:
     branches:
       - master
   push:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
dimalinux, line 11: We don't have to keep these changes if you don't like the concept.

I feel like we get more CI failures for reasons that look like resource depletion (context timeouts as if the test had been paused for a while) when people make a multiple commits in succession to their pull request branch. CI will run tests to completion on every single commit causing many parallel jobs. The changes in this PR will cancel the previous job when the new commit comes in, and only run the tests on the latest commit of the current workflow. (Other workflows are unaffected ... 2 PRs will not cancel each other's tests.)

We have 2 workflows: (1) commits to pull request (2) commits to master. The second flow isn't likely to be affected by these changes, as it would be hard to have a 2nd pull request finish its tests in time and be allowed to merge to master before the previous master tests completed.

Downside 1: From a Github UI perspective, the cancelled tests from the previous commit will have a red `x` by them. ... just discovered that testing this PR. A minor nuisance, but better than the tests on the most recent commit failing due to resource depletion.

Downside 2: I don't see github documentation that resources are shared. I do feel that there is correlation, but I can't prove it without a doubt.

daemon/cancel_or_refund_test.go
```diff
@@ -61,13 +63,13 @@ func TestXMRTakerCancelOrRefundAfterKeyExchange(t *testing.T) {
 	go func() {
 		defer wg.Done()
 		for {
-			count, err := ec.PendingTransactionCount(ctx) //nolint:govet
+			aliceNonce, err := ec.PendingNonceAt(ctx, aliceAddr) //nolint:govet
```
dimalinux, line 66: Another test from another package can be running and make us think the `newSwap` contract method was called when using `PendingTransactionCount`, but since Alice's ganache key is not shared between packages, an incremented nonce check is safe.

I looked at other options before this simple one. Without a transaction ID, I don't see any API to pull full pending transaction data from the mempool. I tried creating a separate client to the DB, which may not even be safe (don't know enough about the underlying DB), but the 2nd client wasn't seeing the new TX ID as soon as Alice's client added the entry ... there was significant delay. I could be wrong, but I don't think you can use filters with pending transactions.

go.mod
```diff
@@ -4,17 +4,17 @@ go 1.20
 
 require (
 	filippo.io/edwards25519 v1.0.0
-	github.com/ChainSafe/chaindb v0.1.5-0.20221010190531-f900218c88f8
-	github.com/MarinX/monerorpc v1.0.5
+	github.com/ChainSafe/chaindb v0.1.5
+	github.com/MarinX/monerorpc v1.0.6
```
dimalinux, line 8: That was a PR I did against MarinX's repo. (For incoming transfers, a feature we don't use.)

There are libp2p updates (only a single minor version increase, but backwards incompatible), that will involve code updates, so I didn't include them in this PR.

.github/workflows/build-macos.yml
noot: seems ok, downside 1 is sort of annoying but can just wait for the job to finish before pushing i guess

noot (approved):
looks good!

PR #481, opened 2023-06-04
ETH transfer fixes and sweep ETH addition

dimalinux:
* Fixes problems with the existing ETH transfer method
* Allows transfers to contracts or regular account address
* Adds a sweep ETH feature that's symmetric to the existing XMR sweep feature
* Provides unit tests to existing and new code


ethereum/contracts/TestERC20.sol
```diff
@@ -38,4 +38,13 @@ contract TestERC20 is ERC20 {
     function approveInternal(address owner, address spender, uint256 value) public {
         _approve(owner, spender, value);
     }
+
+    // You can send a zero-value transfer directly to the contract address to
+    // get a 100 standard unit tokens.
```
dimalinux, line 43: I needed a contract address that unit tests could transfer ETH to. Copying this idea from the Weenus token project seemed like the easiest way to enable that capability.

ethereum/extethclient/eth_wallet_client.go
```diff
@@ -301,62 +317,136 @@ func (c *ethClient) Raw() *ethclient.Client {
...
 	ctx context.Context,
-	nonce uint64,
-	gasPrice *big.Int,
-) (ethcommon.Hash, error) {
+	to ethcommon.Address,
+	amount *coins.WeiAmount,
+	gasLimit *uint64,
+) (*ethtypes.Receipt, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
```
dimalinux, line 330: The previous version of `Transfer` was grabbing the nonce-synchronization lock, but then it wasn't waiting for a confirmation, defeating the purpose of the lock. You wouldn't know if the transfer succeeded or not and even if it did succeed, it wouldn't show up in the balance while the TX was still in the mempool.  The new version still grabs the lock, but it also waits for the receipt or error. The previous version of `CancelTxWithNonce` was grabbing the nonce-synchronization lock, and that was removed. With a fixed nonce, there is no nonce to synchronize, and attempting to grab the lock is not what we want if a different method, that we wait to fail, already has the lock and is blocked trying to get a receipt.

ethereum/contracts/TestERC20.sol
```diff
@@ -38,4 +38,13 @@ contract TestERC20 is ERC20 {
...
+    receive() external payable {
+        mint(msg.sender, 100 * 10 ** uint(_decimals));
```
dimalinux, line 45: I had `10**uint(_decimals)` which I like better, but the solidity formatter separated them.

ethereum/extethclient/eth_wallet_client.go
```diff
@@ -51,12 +53,26 @@ type EthClient interface {
...
 
-	// transfers ETH to the given address
-	// does not need locking, as it locks internally
-	Transfer(ctx context.Context, to ethcommon.Address, amount *coins.WeiAmount) (ethcommon.Hash, error)
-
-	// attempts to cancel a transaction with the given nonce by sending a zero-value tx to ourselves
-	CancelTxWithNonce(ctx context.Context, nonce uint64, gasPrice *big.Int) (ethcommon.Hash, error)
+	// Transfer transfers ETH to the given address, nonce Lock()/Unlock()
+	// handling is done internally. The gasLimit field when the destination
+	// address is not a contract.
```
dimalinux, line 58: I don't like having to maintain comments on both the interface and the matching functions in the same file. If you agree, I'll remove this interface in a follow-up PR, since it is just the full public API of the interface's only implementation.

cliutil/utils.go
```diff
@@ -136,24 +137,56 @@ func GetVersion() string {
...
 		return nil, fmt.Errorf("value of flag --%s cannot be zero", flagName)
 	}
-	if bf.Negative {
-		return nil, fmt.Errorf("value of flag --%s cannot be negative", flagName)
+
+	return d, nil
+}
+
+// ReadETHAddress reads a string flag and parses to an ethereum Address type
+func ReadETHAddress(ctx *cli.Context, flagName string) (*ethcommon.Address, error) {
```
noot, line 176: why return pointer?

cmd/swapcli/main.go
```diff
@@ -378,6 +368,23 @@ func cliApp() *cli.App {
 						Usage:    "Amount of ETH to send",
 						Required: true,
 					},
+					&cli.Uint64Flag{
+						Name:  flagGasLimit,
+						Usage: "Set the gas limit (required if transferring to contract, otherwise ignored)",
```
noot, line 373: why is this ignored for transfers to EOAs/why is it enforced for contracts?

cmd/swapcli/main.go
```diff
@@ -1225,31 +1236,60 @@ func runSweepXMR(ctx *cli.Context) error {
...
 	amount, err := cliutil.ReadUnsignedDecimalFlag(ctx, flagAmount)
 	if err != nil {
 		return err
 	}
 
+	var gasLimit *uint64
+	if ctx.IsSet(flagGasLimit) {
+		gasLimit = new(uint64)
+		*gasLimit = ctx.Uint64(flagGasLimit)
+	}
```
noot, line 1253: this appears to not be ignored for EOAs, contradicting the flag comment

rpc/personal.go
```diff
@@ -167,24 +167,48 @@ func (s *PersonalService) SweepXMR(_ *http.Request, req *SweepXMRRequest, resp *
...
 
-	resp.TxHash = txHash
+	resp.TxHash = receipt.TxHash
+	return nil
+}
+
+// SweepETHRequest is JSON-RPC request object for SweepETH
+type SweepETHRequest struct {
+	To       ethcommon.Address `json:"to" validate:"required"`
+	GasLimit *uint64           `json:"gasLimit,omitempty"`
```
noot, line 197: this is unused

protocol/xmrtaker/instance.go
```diff
@@ -282,7 +282,7 @@ func (inst *Instance) maybeCancelNewSwap(txHash ethcommon.Hash) (bool, error) {
 	// loop for 1 hour and block as there will never be a receipt for the cancel tx.
 	// we should probably poll for our account nonce to increase, and when it does, check for receipts
 	// on both txs to see which one was successful.
-	receipt, err := block.WaitForReceipt(inst.backend.Ctx(), inst.backend.ETHClient().Raw(), cancelTx)
+	receipt, err := block.WaitForReceipt(inst.backend.Ctx(), inst.backend.ETHClient().Raw(), cancelTx.TxHash)
```
noot, line 285: `CancelTxWithNonce` now returns the receipt so this is unneeded, but the comment still applies, so leave the error check under this line for the receipt

noot (approved):
looks good, just minor comments!

cliutil/utils.go
dimalinux: I updated it to a non-pointer to make our code style consistent.

rpc/personal.go
dimalinux: Thanks, removed it.

protocol/xmrtaker/instance.go
```diff
@@ -271,26 +270,9 @@ func (inst *Instance) maybeCancelNewSwap(txHash ethcommon.Hash) (bool, error) {
...
-	cancelTx, err := inst.backend.ETHClient().CancelTxWithNonce(inst.backend.Ctx(), tx.Nonce(), gasPrice)
+	receipt, err := inst.backend.ETHClient().CancelTxWithNonce(inst.backend.Ctx(), tx.Nonce(), gasPrice)
 	if err != nil {
-		return false, fmt.Errorf("failed to create or send cancel tx: %w", err)
-	}
-
-	log.Infof("submit cancel tx %s", cancelTx)
-
-	// TODO: if newSwap is included instead of the cancel tx (unlikely), block.WaitForReceipt will actually
-	// loop for 1 hour and block as there will never be a receipt for the cancel tx.
```
dimalinux, line 282: I don't think this comment is correct, at least not on mainnet or sepolia, so I didn't move it to another point in the code. (Ganache doesn't have a mempool, and it might not handle things correctly.) Whichever transaction gets included in a block first will successfully get a receipt, while the other one will get an error waiting for a receipt because it uses the same nonce and can't be included in any future block.

protocol/xmrtaker/instance.go
```diff
@@ -271,23 +271,12 @@ func (inst *Instance) maybeCancelNewSwap(txHash ethcommon.Hash) (bool, error) {
...
-	// we should probably poll for our account nonce to increase, and when it does, check for receipts
-	// on both txs to see which one was successful.
-	receipt, err := block.WaitForReceipt(inst.backend.Ctx(), inst.backend.ETHClient().Raw(), cancelTx)
+	receipt, err := inst.backend.ETHClient().CancelTxWithNonce(inst.backend.Ctx(), tx.Nonce(), gasPrice)
 	if err != nil && !errors.Is(err, ethereum.NotFound) {
 		return false, fmt.Errorf("failed to get cancel transaction receipt: %w", err)
 	}
 
-	if errors.Is(err, ethereum.NotFound) || receipt.Status == ethtypes.ReceiptStatusFailed {
```
dimalinux, line 290: Our `block.WaitForReceipt` method doesn't return a receipt if the status failed. If you get a receipt, but the status is "failed", it means the transaction *was* included in a block despite the fact that it was reverted. Our `WaitForReceipt` method tries to pull the revert reason out of the block that the transaction was included in. I could be wrong, but I don't think we need to worry about it here. I don't think a simple transfer to a non-contract address can be reverted. If the gas limit is too low, the geth APIs won't even let you send the simple transfer into the mempool. The balance can't be too low, since we are sending 0 eth. The cancel tx should either get included successfully into a block, or not included at all. If I'm wrong, I'd be really curious to know how it can happen.

PR #484, opened 2023-06-08
USDT compatibility

dimalinux:
Mainnet testing with USDT found some issues:

1. The signatures of `transfer` and `transferFrom` in older token contracts do not return `bool`. While we were not checking the return value, as these functions should either return `true` or revert, we were still having compatibility issues for reasons [described here](https://ethereum.stackexchange.com/a/89567/32089) due to USDT's use of `onlyPayloadSize`.
2. USDT's `approve` call will fail if you don't first set the approved allowance to zero before changing the allowance to a new value.

Note: Openzepplin code is minimized to just the contract files that we use, but they are now in an @openzepplin folder which will make it easier to see if the contracts we use have changed when openzepplin makes new releases.

For stagenet testing, someone has created a replica of Tether on sepolia with an additional `_mint` capability added here. You can connect metamask directly to etherscan to mint yourself some TestTether.
https://sepolia.etherscan.io/address/0x7169d38820dfd117c3fa1f22a697dba58d90ba06

Additional changes needed before merge:
1. Deploy mainnet contract
2. Prevent swapd nodes from interacting with other swapd nodes that are using an incompatible version of the contract.


ethereum/contracts/@openzeppelin/contracts/utils/Context.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts v4.4.1 (utils/Context.sol)
 
-pragma solidity ^0.8.19;
+pragma solidity ^0.8.0;
```
dimalinux, line 4: Starting with this PR we don't modify or format the openzepplin contracts, making diffs with the official releases clean.

ethereum/contracts/@openzeppelin/contracts/token/ERC20/IERC20.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts (last updated v4.6.0) (token/ERC20/IERC20.sol)
+// OpenZeppelin Contracts (last updated v4.9.0) (token/ERC20/IERC20.sol)
```
dimalinux, line 2: Strange that the last modified version changed, but we don't see a diff anywhere else in the file.

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,81 @@
...
+        // after which Alice can refund if Bob did not claim.
         uint256 timeout2;
-        // the asset being swapped: equal to address(0) for ETH, or an ERC-20 token address
+        // asset is address(0) for ETH swaps, or an ERC-20 address if Alice is
+        // providing tokens for XMR.
         address asset;
-        // the value of this swap
+        // value is a wei amount that Alice locked in the contract, if asset is
+        // the zero address, otherwise it a unit amount for the token that Alice
+        // locked.
```
dimalinux, line 59: Note 1: Remove comma before first `if`. 
Note 2: ... otherwise it is ...

daemon/swap_daemon_erc20_test.go
```diff
@@ -21,26 +24,52 @@ import (
 // Tests the scenario where Bob has XMR and enough ETH to pay gas fees for the token claim. He
 // exchanges 2 XMR for 3 of Alice's ERC20 tokens.
 func TestRunSwapDaemon_ExchangesXMRForERC20Tokens(t *testing.T) {
-	minXMR := coins.StrToDecimal("1")
-	maxXMR := coins.StrToDecimal("2")
-	exRate := coins.StrToExchangeRate("1.5")
+	fundingEC := extethclient.CreateTestClient(t, tests.GetTakerTestKey(t))
```
dimalinux, line 27: I refactored this test to make sure that Alice uses a completely new ethereum key. At the time, I was trying to figure out if the failures were related to `approve`. I think the test is better this way, so I didn't revert the changes.

daemon/swap_daemon_erc20_test.go
```diff
@@ -21,26 +24,52 @@ import (
...
+	// any of those tokens by giving her a brand-new ETH key.
+	aliceEthKey, err := crypto.GenerateKey()
+	require.NoError(t, err)
+	aliceConf := CreateTestConf(t, aliceEthKey)
+	timeout := 7 * time.Minute
 
-	tokenAsset := getMockTetherAsset(t, aliceConf.EthereumClient)
+	// Fund Alice and Bob with a little ether for gas. Bob needs gas to claim,
+	// as ERC20 token swaps cannot use a relayer.
+	_, err = fundingEC.Transfer(context.Background(), aliceConf.EthereumClient.Address(), gasMoney, nil)
```
dimalinux, line 53: This line should be updated when we add the new token transfer API to the extended ETH client:
https://github.com/AthanorLabs/atomic-swap/issues/485

daemon/swap_daemon_erc20_test.go
```diff
@@ -21,26 +24,52 @@ import (
...
+	require.NoError(t, err)
+
+	// Fund Alice with the exact amount of token that she'll provide in the swap
+	// with Bob. After the swap is over, her token balance should be exactly
+	// zero.
+	erc20Iface, err := contracts.NewIERC20(tokenAddr, fundingEC.Raw())
+	require.NoError(t, err)
+	txOpts, err := fundingEC.TxOpts(context.Background())
+	require.NoError(t, err)
+	_, err = erc20Iface.Transfer(txOpts, aliceConf.EthereumClient.Address(), providesAmt.BigInt())
```
dimalinux, line 65: This line should be updated when we add the new token transfer API to the extended ETH client:
https://github.com/AthanorLabs/atomic-swap/issues/485

net/initiate.go
```diff
@@ -84,6 +84,9 @@ func (h *Host) receiveInitiateResponse(stream libp2pnetwork.Stream, s SwapState)
 
 		err := s.HandleProtocolMessage(msg)
 		if err != nil {
+			// todo: shorten this error and raise the level to ERROR
+			//nolint:lll
+			// 2023-06-08T00:12:42.499-0500	WARN	net	net/initiate.go:87	failed to handle protocol message: err=failed to handle EventKeysReceivedType: failed to lock ethereum asset in contract: failed to lock asset: failed to instantiate swap on-chain: new_swap tx creation failed, VM Exception while processing transaction: revert
```
dimalinux: I changed the log to an error and I removed the redundant "failed to lock asset" wrapping.

common/config.go
```diff
@@ -99,7 +99,7 @@ func StagenetConfig() *Config {
 				Port: 38081,
 			},
 		},
-		SwapCreatorAddr: ethcommon.HexToAddress("0xbf2B7a6dCE5598Cf002B3507a8D62cf2C35cE5c6"),
+		SwapCreatorAddr: ethcommon.HexToAddress("0xAD031B747757d5d308e0f28544AF162A89782B6E"),
```
noot, line 102: we need to update mainnet too i suppose?

ethereum/contracts/@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @dev Interface of the ERC20 Permit extension allowing approvals to be made via signatures, as defined in
+ * https://eips.ethereum.org/EIPS/eip-2612[EIP-2612].
+ *
+ * Adds the {permit} method, which can be used to change an account's ERC20 allowance (see {IERC20-allowance}) by
+ * presenting a message signed by the account. By not relying on {IERC20-approve}, the token holder account doesn't
+ * need to send a transaction, and thus is not required to hold Ether at all.
+ */
+interface IERC20Permit {
```
noot, line 14: why do we need this? can you remove (and remove `safePermit` from the `SafeERC20.sol` file?)

noot: is it more so we don't have to modify the openzeppelin deps?

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
 pragma solidity ^0.8.19;
 
-import {IERC20} from "./IERC20.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {Secp256k1} from "./Secp256k1.sol";
 
+// SwapCreator facilitates swapping between Alice, a party that has ETH or an
+// ERC-20 token that she wants to exchange cross-chain for Monero (XMR), and
+// Bob, a party that has XMR and wishes to exchange it for Alice's ETH or token.
```
noot, line 10: remove xmr mentions - the contract itself is generic, the other chain can be anything

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
+    // swaps maps from a swap ID to the swap's current Stage
+    mapping(bytes32 => Stage) public swaps;
+
+    // Swap stores the initial swap configuration, the hash of which forms the
+    // swap ID.
     struct Swap {
-        // the swap initiator, Alice
-        // address allowed to refund the ether for this swap
+        // owner is the address of Alice, who initiates the swap by calling
+        // `newSwap`. Only the owner is allowed to call `setReady` or `refund`.
```
noot, line 34:
```suggestion
        // `owner` is the address of the swap initiator (ie. the address which called `newSwap`).
        // Only the owner is allowed to call `setReady` or `refund`.
```

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
         address payable owner;
-        // address allowed to claim the ether for this swap, Bob
+        // claimer is the address of Bob. Only the claimer can call `claim` or
```
noot, line 36:
```suggestion
        // `claimer` is the address of Bob. Only the claimer can call `claim` or
```

same for other variable names in comments below

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
+        // timeout1 is the block timestamp before which Alice can call
+        // either `setReady` or `refund`.
         uint256 timeout1;
-        // timestamp after which Bob cannot claim, only Alice can refund
+        // timeout2 is the block timestamp after which Bob cannot claim, and
+        // after which Alice can refund if Bob did not claim.
         uint256 timeout2;
-        // the asset being swapped: equal to address(0) for ETH, or an ERC-20 token address
+        // asset is address(0) for ETH swaps, or an ERC-20 address if Alice is
+        // providing tokens for XMR.
```
noot, line 62: I think this was clearer before

ethereum/contracts/SwapCreator.sol
```diff
@@ -181,7 +225,8 @@ contract SwapCreator is Secp256k1 {
         return swapID;
     }
 
-    // Alice should call setReady() before timeout1 once she verifies the XMR has been locked
+    // Alice should call `setReady` before timeout1 and after verifying that Bob
+    // locked his XMR.
```
noot, line 229:
```suggestion
    // locked his swap funds.
```

ethereum/contracts/SwapCreator.sol
```diff
@@ -190,26 +235,27 @@ contract SwapCreator is Secp256k1 {
...
+    // Bob can call `claim` if either of these hold true:
     // (1) Alice has set the swap to `ready` and it's before timeout1
-    // (2) It is between timeout0 and timeout1
+    // (2) It is between timeout1 and timeout2
     function claim(Swap memory _swap, bytes32 _secret) public {
         if (msg.sender != _swap.claimer) revert OnlySwapClaimer();
         _claim(_swap, _secret);
 
-        // send ether to swap claimer
+        // Send ETH to swap claimer if token asset address is zero, otherwise
```
noot, line 245:
```suggestion
        // Send ETH to swap claimer if asset address is zero, otherwise
```

ethereum/contracts/SwapCreator.sol
```diff
@@ -262,9 +308,9 @@ contract SwapCreator is Secp256k1 {
         swaps[swapID] = Stage.COMPLETED;
     }
 
-    // Alice can claim a refund:
-    // - Until timeout1 unless she calls setReady
-    // - After timeout2
+    // Alice can `refund` her locked ethereum asset:
```
noot, line 311:
```suggestion
    // Alice can refund her swap funds:
```

noot (commented):
logic seems good, just smaller comments!

common/config.go
dimalinux: Yes, when there are no more changes to make, we'll update it. I made the `t.Skip(...)` in `TestMainnetContract` loud so that we don't forget.

ethereum/contracts/SwapCreator.sol
dimalinux: The code was using a mix of back ticks and trailing parenthesis when referencing function names. As part of this PR I went through the contract and harmonized on back ticks. I think we do want something special for method names, just like Natspec uses `{...}` around method names, but I'm a little worried that using the same syntax for method names and field/variable names and/or type names might not increase readability. I very deliberately crafted each field comment to start with the variable name, so it would be clear to the reader that it is the variable name. Any thoughts on this?

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
         INVALID,
         PENDING,
         READY,
         COMPLETED
     }
 
+    // swaps maps from a swap ID to the swap's current Stage
+    mapping(bytes32 => Stage) public swaps;
+
+    // Swap stores the initial swap configuration, the hash of which forms the
```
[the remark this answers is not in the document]
dimalinux: Committed suggestion exactly as proposed, I'll resolve this one.

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
+        // for.
         address payable claimer;
-        // the keccak256 hash of the expected public key derived from the secret `s_b`.
-        // this public key is a point on the secp256k1 curve
+        // pubKeyClaim is the Keccak-256 hash of the expected public key derived
+        // from the secret that Bob sends when claiming. This public key is a
+        // point on the secp256k1 curve. Bob gives Alice this public key hash
+        // and proves that he already knows the secret associated with it, and
+        // that the secret also forms a valid point on the ed25519 curve,
+        // off-chain before Alice initiates the swap on-chain.
```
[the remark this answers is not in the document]
dimalinux: I renamed the variables to the suggested names and shortened the comments. I didn't like the use of `s_b`, as we never defined it. To get the brevity you wanted, I think we can safely assume that the readers that care already know that elliptic curve private keys are scalars, and that you multiply them by a generator point to get the public key.
```text
        // claimCommitment is the Keccak-256 hash of the expected secp256k1
        // public key derived from the secret (private key) that Bob sends when
        // claiming. Alice receives this commitment off-chain.
        bytes32 claimCommitment;
```
```text
        // refundCommitment is the Keccak-256 hash of the expected secp256k1
        // public key derived from the secret (private key) that Alice sends if
        // refunding.
        bytes32 refundCommitment;
```

ethereum/contracts/SwapCreator.sol
```diff
@@ -1,61 +1,88 @@
...
+        // forms a valid point on the ed25519 curve, off-chain before starting
+        // the swap on-chain.
         bytes32 pubKeyRefund;
-        // timestamp before which Alice can call either `setReady` or `refund`
+        // timeout1 is the block timestamp before which Alice can call
+        // either `setReady` or `refund`.
         uint256 timeout1;
-        // timestamp after which Bob cannot claim, only Alice can refund
+        // timeout2 is the block timestamp after which Bob cannot claim, and
+        // after which Alice can refund if Bob did not claim.
```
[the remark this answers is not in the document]
dimalinux: I updated it to just:
```text
        // timeout2 is the block timestamp after which Bob cannot claim, only
        // Alice can refund.
```
The only change now is placing the field name first and clarifying that the timestamp is the block timestamp, as that is a question I get asked a lot.

ethereum/contracts/SwapCreator.sol
dimalinux: Done. I removed the references to ETH as well, as the contract can most likely be deployed as-is on BNB Smart Chain, and the Go changes required wouldn't be that extensive either. Infrastructure changes would be minimal once our bootnodes work across multiple chains.

ethereum/contracts/SwapCreator.sol
dimalinux: Give that we now define "Alice" and "Bob" at the start of the contract, but we haven't defined "swap initiator" yet, doesn't it flow more logically the way it is?

ethereum/contracts/SwapCreator.sol
dimalinux: Original:
```text
        // the asset being swapped: equal to address(0) for ETH, or an ERC-20 token address
        address asset;
```
Latest version:
```text
        // asset is address(0) for EVM native currency swaps, or it is the
        // address of the token that Alice is providing.
        address asset;
```
At the top of the contract, we clarify our definition of the word "token" for rest of the documentation. It could be a BEP-20 token. The field name was shifted to the first word for style consistency.

ethereum/contracts/SwapCreator.sol
dimalinux: Resolving. Change made as proposed.

ethereum/contracts/SwapCreator.sol
dimalinux: Updated to "swap funds". Resolving.

protocol/txsender/sender.go
```diff
@@ -127,10 +118,104 @@ func (s *privateKeySender) NewSwap(
...
+	if err = saveNewSwapTxCallback(tx.Hash()); err != nil {
+		return nil, err
+	}
+
+	receipt, err := block.WaitForReceipt(s.ctx, s.ethClient.Raw(), tx.Hash())
+	if err != nil {
+		return nil, fmt.Errorf("NewSwap tx %s failed waiting for receipt, %w", tx.Hash(), err)
+	}
+
+	log.Infof("NewSwap TX succeeded, %s", common.ReceiptInfo(receipt))
```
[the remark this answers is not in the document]
dimalinux: Done. Resolving.

ethereum/contracts/SwapCreator.sol
dimalinux: I rearranged it a little. The check to see if the address is zero is, arguably, more clear to see from the code than what any comment can provide.

ethereum/contracts/SwapCreator.sol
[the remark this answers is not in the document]
noot: can you add this?
`// note that the public key is encoded in uncompressed form before hashing.`

ethereum/contracts/SwapCreator.sol
noot: honestly i don't read code from top to bottom, so not really for me :p but not a huge deal

noot (approved):
looks good!

PR #486, opened 2023-06-15
updates for cross-chain bootnodes

dimalinux:
This PR depends on https://github.com/AthanorLabs/go-p2p-net/pull/9, which should get merged with a new release tag first.

We don't want to maintain separate sets of bootnodes for every supported environment. The `go-p2p-net` library, which was already adding a protocol ID prefix to stream names, was adjusted to add the same prefixes to advertised-in-the-DHT namespaces. Instead of advertising "XMR" in the namespace, `swapd` instance will now advertise `/atomic-swap/${chain_name}/${protocol_version}/XMR` instead.


common/config.go
```diff
@@ -26,6 +25,21 @@ const (
 var homeDir, _ = os.UserHomeDir()
 var baseDir = path.Join(homeDir, ".atomicswap")
 
+// publicBootnodes are bootnodes with public IP addresses and they used in every
+// environment other than development
+var publicBootnodes = []string{
+	"/ip4/109.105.198.218/tcp/9909/p2p/12D3KooWRbAV4AU75JXVUedmwakiM1yzzDzDCdmZWzGdXGQ4Yqdv",
```
dimalinux, line 31: There are 2 nodes on port 9909 that didn't have a 9900 entry on the same IP in the former stagenet list. All others that had 2 entries on the same IP were consolidated in favor of the entry on port 9900.

common/config.go
```diff
@@ -26,6 +25,22 @@ const (
 var homeDir, _ = os.UserHomeDir()
 var baseDir = path.Join(homeDir, ".atomicswap")
 
+// publicBootnodes are bootnodes with public IP addresses and they used in every
+// environment other than development.
+var publicBootnodes = []string{
```
dimalinux, line 30: This list combined the former stagenet and mainnet bootnodes. The `bootnode` executable uses port `9909` by default, mainly to reduce conflicts when `swapd` and a `bootnode` are running on the same host. If there were 2 entries from the same IP, I picked the one on port `9909`. Some of the ones on port `9900` are probably `swapd` instances, which is fine too.

common/config.go
```diff
@@ -26,6 +25,22 @@ const (
 var homeDir, _ = os.UserHomeDir()
 var baseDir = path.Join(homeDir, ".atomicswap")
 
+// publicBootnodes are bootnodes with public IP addresses and they used in every
+// environment other than development.
```
noot, line 29:
```suggestion
// publicBootnodes are bootnodes with public IP addresses that are used in every
// environment other than development.
```

common/config.go
```diff
@@ -186,16 +199,19 @@ func DefaultMoneroPortFromEnv(env Environment) uint {
...
 
-// ChainIDFromEnv returns the expected chainID that we should find on the
+// ChainNameFromEnv returns the expected chainID that we should find on the
 // ethereum endpoint when running int the passed environment.
-func ChainIDFromEnv(env Environment) *big.Int {
+func ChainNameFromEnv(env Environment) string {
 	switch env {
 	case Development:
-		return big.NewInt(GanacheChainID)
+		return "Ganache"
```
noot, line 207: i kind of prefer lower case as this is going in the protocol id with lowercase atomic-swap

net/host.go
```diff
@@ -26,9 +27,14 @@ import (
 )
 
 const (
-	// ProtocolID is the base atomic swap network protocol ID prefix. The full ID
-	// includes the chain ID at the end.
-	ProtocolID          = "/atomic-swap/0.3"
+	baseProtocolID = "/atomic-swap"
+
+	// p2pAPIVersion needs to be incremented every time:
+	// * fields that the p2p APIs exchange, like offers, change in a breaking way
```
noot, line 33:
```suggestion
	// * types that the p2p APIs exchange, such as offers, change in a breaking way
```

net/host.go
```diff
@@ -75,30 +81,38 @@ type Host struct {
...
+	DataDir   string
+	Port      uint16
+	KeyFile   string
+	Bootnodes []string
+	ListenIP  string
+	IsRelayer bool
+}
+
+// ChainProtocolID returns the chain versioned p2p protocol ID that includes the
+// Ethereum chain name being used. The streams that communicate between peers
```
noot, line 95:
```suggestion
// Ethereum chain name being used. The streams that are opened between peers
```

net/host.go
```diff
@@ -109,15 +123,14 @@ func NewHost(cfg *Config) (*Host, error) {
...
-		ProtocolID:               cfg.ProtocolID,
+		ProtocolID:               ChainProtocolID(cfg.Env),
 		ListenIP:                 cfg.ListenIP,
 		AdvertisedNamespacesFunc: h.advertisedNamespaces,
 	})
 	if err != nil {
 		return nil, err
 	}
 
-	log.Debugf("using base protocol %s", cfg.ProtocolID)
```
noot, line 120: no more logging the protocol ID? i think this is useful

noot (commented):
logic looks fine, however this PR is fully non-backwards-compatible, so we'll need to cut a new release and tell everyone to upgrade

noot (approved):

common/config.go
dimalinux: I made them all lower case.

net/host.go
dimalinux: Implemented verbatim. Resolving.

net/host.go
dimalinux: Implemented verbatim. Resolving.

common/config.go
dimalinux: Implemented verbatim. Resolving.

net/host.go
```diff
@@ -75,30 +81,38 @@ type Host struct {
...
+	ListenIP  string
+	IsRelayer bool
+}
+
+// ChainProtocolID returns the chain versioned p2p protocol ID that includes the
+// Ethereum chain name being used. The streams that communicate between peers
+// use this prefix. All provided values advertised in the DHT also use this
+// prefix.
+func ChainProtocolID(env common.Environment) string {
+	return fmt.Sprintf("%s/%s/%d", baseProtocolID, common.ChainNameFromEnv(env), p2pAPIVersion)
```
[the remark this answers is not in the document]
dimalinux: I added extra notes in the function description.

net/host.go
dimalinux: We were getting some logging of the protocol ID from the `go-p2p-net` package when the stream handlers are created, but I just added this one back a little further up.

PR #487, opened 2023-06-18
update mainnet bootnodes

noot:


dimalinux:
Can you just make this change directly in the other open PR so we don't get a merge conflict?

dimalinux:
> Can you just make this change directly in the other open PR so we don't get a merge conflict?

Actually, there is no need. I had already corrected both of those in the `dimalinux/cross-chain-bootnodes` PR that is waiting for review. The `cross-chain-bootnodes` PR adds a `TestPublicBootnodes` test that is skipped in CI, but is really useful for determining which bootnodes are bad and usually why. I'll just close this PR.

PR #489, opened 2023-06-20
doc updates

dimalinux:
Changes:
* Updates build.md with details on building release binaries
* Updates rpc.md with undocumented `swap_clearOffers` method
* Updates base docker images from bullseye to bookworm

Feel free to hit the edit button on these links to make direct changes to this PR.
* https://github.com/AthanorLabs/atomic-swap/blob/dimalinux/doc-updates/docs/build.md
* https://github.com/AthanorLabs/atomic-swap/blob/dimalinux/doc-updates/docs/rpc.md

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -2,7 +2,7 @@
 
 ## From source
 
-1. Install Golang
+### 1. Install Golang
 
 On Ubuntu, the easiest way to keep up-to-date with the latest stable version of
 Go is with snap:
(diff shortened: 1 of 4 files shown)


docs/rpc.md
```diff
@@ -45,9 +45,9 @@ curl -s -X POST http://127.0.0.1:5000 -H 'Content-Type: application/json' -d \
 Discover peers on the network via DHT that have active swap offers.
 
 Parameters:
-- `provides` (optional): one of `ETH` or `XMR`, depending on which offer you are searching
+- `provides`: (optional) one of `ETH` or `XMR`, depending on which offer you are searching
```
dimalinux, line 48: I was just harmonizing the style. I don't think it matters which side of the colon the `(optional)` is on, as long as they all do it the same way.

scripts/docker-bootnode/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN go install -tags=prod \
     github.com/athanorlabs/atomic-swap/cmd/swapcli@"${VERSION}"
 RUN /go/bin/bootnode --version
 
-FROM debian:bullseye-slim
+FROM debian:bookworm-slim
```
dimalinux, line 9: The Go docker image that we build the executables in is now based off of bookworm (debian 12) and glibc versions are only backwards compatible, not forwards. We could built the bootnode executable statically, but in this case I was updating a bootnode when I discovered the issue, so maybe keeping the `swapd`/`bootnode` images close to each other is more of a win.

docs/build.md
```diff
@@ -12,38 +12,70 @@ sudo snap install go --classic
...
 source .profile
 ```
 
-2. Clone the repo:
+### 2. Clone the repo
 ```bash
 git clone https://github.com/athanorlabs/atomic-swap.git
 cd atomic-swap
+
+# Optional: Check out the exact source code of the latest release
```
noot, line 27: I don't think this should be optional, since these instructions are directed to users

docs/build.md
```diff
@@ -12,38 +12,70 @@ sudo snap install go --classic
...
 
-This creates `swapd` and `swapcli` binaries in the `bin` folder.
+#### Option 2: Build release binaries without checking out the repo
 
-Note: if you wish to run a bootnode (see [here](./bootnode.md)), run `make build-all`.
+The make-target in "Option 1" uses "go install", a command that can be run
+without checking out the source code. If you just want the release binaries in a
+subfolder named "bin", you can use:
+```bash
+GOBIN=${PWD}/bin go install -tags=prod github.com/athanorlabs/atomic-swap/cmd/...@latest
```
noot, line 46: isn't this the exact command of `build-release`? this seems unneeded

dimalinux: It is, but these are instructions on how to install the binaries without checking out the repository, so they won't have access to the Makefile.

noot: maybe this should be option 1 then, before the cloning the repo option

dimalinux: Done, but that required a major rewrite of the top-level headings in `build.md`. The links in the description will be more helpful than the diffs at this point.

noot (approved):

PR #494, opened 2023-06-25
restore price feed

### LINKED ISSUE #492: Chainlink XMR price feed is gone
The `./bin/swapcli price-feed` command is returning an error right now (execution reverted). I looked into it and while the ETH/USD price feed is working, the `TestGetXMRUSDPrice_mainnet` test which calls `pricefeed.GetXMRUSDPrice` is broken.

Chainlink appears to have just removed the webpage for the feed without leaving any information on why/when it was taken down:
https://data.chain.link/ethereum/mainnet/crypto-usd/xmr-usd

The price feed still exists on Optimism if we want to pull it from there:
https://data.chain.link/optimism/mainnet/crypto-usd/xmr-usd

Optimism has a free https://mainnet.optimism.io RPC endpoint that is rate limited, but I imagine that the rate limit is high enough that users would not trigger it doing occasional price feed requests.

Other options would include Kraken's price feeds:
```text
curl -s "https://api.kraken.com/0/public/Ticker?pair=XMRUSD" | jq .
```
Descriptions of the response field names can be found here:
https://docs.kraken.com/rest/#tag/Market-Data/operation/getTickerInformation

dimalinux:
* Restores our suggested exchange rate ("price feed"), which broke when Chainlink removed the ETH mainnet feed for Monero. New version uses the Chainlink feed on Optimism.
* Minor tweaks to the swapcli flag aliases and help

Closes #492

### DIFF
--- a/cmd/swapcli/main.go
+++ b/cmd/swapcli/main.go
@@ -169,11 +169,13 @@ func cliApp() *cli.App {
 				Flags: []cli.Flag{
 					&cli.StringFlag{
 						Name:     flagMinAmount,
+						Aliases:  []string{"min"},
 						Usage:    "Minimum amount to be swapped, in XMR",
 						Required: true,
 					},
 					&cli.StringFlag{
 						Name:     flagMaxAmount,
+						Aliases:  []string{"max"},
 						Usage:    "Maximum amount to be swapped, in XMR",
 						Required: true,
 					},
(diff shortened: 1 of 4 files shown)


cmd/swapcli/main.go
```diff
@@ -184,15 +186,16 @@ func cliApp() *cli.App {
 					},
 					&cli.BoolFlag{
 						Name:  flagDetached,
-						Usage: "Exit immediately instead of subscribing to notifications about the swap's status",
+						Usage: "Exit immediately without subscribing to status notifications",
```
dimalinux, line 189: Adding new aliases lengthens the help lines, so I scanned the help for places where I could shorten the lines.

cmd/swapcli/main.go
```diff
@@ -308,8 +312,8 @@ func cliApp() *cli.App {
 				},
 			},
 			{
-				Name:    "suggested-exchange-rate",
-				Aliases: []string{"price-feed"},
+				Name:    "price-feed",
+				Aliases: []string{"suggested-exchange-rate"},
```
dimalinux, line 316: I'm open to other names. We do report both prices, not just the suggested exchange rate. `suggested-exchange-rate` was too much typing for me. It is kind of long to even keep around as an alias.

This is what we print:
```text
Exchange rate: 0.08537
XMR/USD Price: 162.21        (2023-06-25 18:12:03 -0500 CDT)
ETH/USD Price: 1900.08429046 (2023-06-25 18:20:39 -0500 CDT)
```
One positive thing about the switch to the Chainlink feed on Optimism for XMR is that it is updated at a much higher frequency. I think the old (now dead) feed was only getting updated like every 4 hours, while the new one appears to be every 15 minutes.

noot (approved):
looks good!

Issue #491, opened 2023-06-21
approved tokens list


dimalinux:
I find it too painful to pass `--token ...` every time to `./swapcli balances`. I find myself having to lookup the token address every time and I even forget which tokens I have. I'm going to propose a possible solution, but anyone who thinks they have a better idea please add commentary.

Solution: We have a tokens.yml file in `${DATA_DIR}/${ENV}`.  I'm picking YAML instead of JSON so users can add comments. In the file is a list of addresses of tokens that an individual user has verified. These tokens are always shown by the `balances` subcommand to `swapcli` and an entry is required if a user wants to take an offer for that token. This will prevent users from taking a token that has a deceptive symbol name, say USDT, but isn't at the official USDT address.

Open questions: Would the user always edit this file by hand, or would we have a `swapcli` subcommand to add an entry? If the file does not exist, we can always create a template version with no entries to help the user.

noot:
yeah, this relates to https://github.com/AthanorLabs/atomic-swap/issues/178, we should probably implement a whitelist for tokens

dimalinux:
> I would suggest that all offers should still be available but with proper warnings when not from verified tokens 

Yeah, in the CLI we could add the text  `(verified)` when the token is in the whitelist and allow the user to take the offer even if it is not verified. ... but nudge them towards adding any tokens they want to swap to their own whitelist in the documentation.

Issue #178, opened 2022-09-01
ERC20 support: add whitelist for tokens


noot:
see comment here: https://github.com/AthanorLabs/atomic-swap/pull/158/files#r954398151
we should add a whitelist that only allows specific ERC20s. potentially make the list update-able also.

code comment inside `newSwap()`:
```
            // transfer ERC-20 token into this contract
            // TODO: potentially check token balance before/after this step
            // and ensure the balance was increased by swap.value since fee-on-transfer
            // tokens are not supported
            IERC20(_asset).transferFrom(msg.sender, address(this), _value);
```

code comment inside `claim()`
```
            // TODO: this will FAIL for fee-on-transfer or rebasing tokens if the token
            // transfer reverts (i.e. if this contract does not contain _swap.value tokens),
            // exposing Bob's secret while giving him nothing
            //
            // potential solution: wrap tokens into shares instead of absolute values
            // swap.value would then contain the share of the token
            IERC20(_swap.asset).transfer(_swap.claimer, _swap.value - fee);
```

Issue #75, opened 2022-01-16
get xmr-holder to send a proof of funds before ether is locked


noot:
there is a potential for a bad actor to make a monero swap offer, but without actually having the funds. then some eth holding party may come, lock their funds, and then be forced to eventually refund.

getting the xmr-holder to send a proof of funds (https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#get_reserve_proof) in their initial message, which the counterparty will then verify, prevents the case where the xmr-holder didn't have the funds to begin with. however, it doesn't prevent against the case where they don't lock the funds. it also forces the xmr-holder to reveal the address that initially holds the funds, which is not ideal.

I might not end up implementing this due to the above drawbacks, needs some thought.

PR #495, opened 2023-07-02
add pairs

stubbrn:
Ability to select Pairs

**Just a mockup for now**, as the logic of the future "getPairs" RPC endpoint requires to know whether there's a "hard" whitelist or not (as discussed in #491 and #178).

pairs
pairs2


dimalinux:
Since this PR is not linked to a well documented issue, can you provide more documentation in the PR description.

Describe:
* The use case that the PR is trying to solve
* What a pair is
* It looks like there is some concept of bidirectional liquidity in the pairs. Can you describe what those liquidity values are, probably as part of the use case, and why the values should be trusted. (Or why the values are useful even if they can't be trusted.)
* What follow-up work is needed to make this PR useful. It sounds like it depends on a whitelist. Describe how you think we should solve that problem.

There are lots of images that are imported. Can you mention where they came from and what their copyright is?

dimalinux:
Let's say someone makes the following absurd offer with a 2 million USDT to XMR exchange rate and half of a monero being offered up.
```text
$ ./bin/swapcli make --token 0xdAC17F958D2ee523a2206206994597C13D831ec7 \
    --min 0.12 --max 0.5 --exchange-rate 2000000
Published:
  Offer ID:  0x20739776aae60ad00d7591165ef8efbbbec34b07d137f3a48169a1a234621194
  Peer ID:   12D3KooWQQRJuKTZ35eiHGNPGDpQqjpJSdaxEMJRxi6NWFrrvQVi
  Taker Min: 240000 "USDT"
  Taker Max: 1000000 "USDT"
```
Would this PR then report that there is a million dollars of USDT liquidity in the network?

dimalinux:
Let's see what @noot or someone else thinks. Are there any exchanges that print numbers like this or that have similar concept of what "liquidity" is? At least if end users are using our software, they can't offer up more XMR than they have. Offering up exchange rates that are extremely far from what anyone would take at the moment is 100% guaranteed to happen. I've never seen it not happen on any platform. I don't think having a single offer that is legit but absurd affect a network wide statistic is desirable behavior.

noot:
@stubbrn sorry for the late response. overall, I like the idea and I think it helps give a nice snapshot view of the network. but yeah, false liquidity reporting would be an issue. we could change the definition for now to say "reported liquidity" or something like that? I've thought a bit about how to add proof of XMR liquidity (#75) but there are still issues, as the maker can just transfer their funds to another account or something after sending the proof. I think for now, potentially changing the definition/adding a specific definition would be good.

noot (commented):
this looks good, however I just noticed the hundreds of token icons added lol. can you change this to have just the tokens currently used + common ERC20 tokens?

rpc/net.go
```diff
@@ -73,6 +88,71 @@ func (s *NetService) Peers(_ *http.Request, _ *interface{}, resp *rpctypes.Peers
 	return nil
 }
 
+// Pairs returns all currently available pairs from offers of all peers
+func (s *NetService) Pairs(_ *http.Request, req *rpctypes.PairsRequest, resp *rpctypes.PairsResponse) error {
+	if s.isBootnode {
+		return errUnsupportedForBootnode
+	}
+
+	peerIDs, err := s.discover(req)
```
noot, line 97: this only discovers peers advertising the coin inside `req.Provides`, but the function comment says it returns "returns all currently available pairs", but this will only return pairs that include the `req.Provides` coin

rpc/net.go
```diff
@@ -73,6 +88,71 @@ func (s *NetService) Peers(_ *http.Request, _ *interface{}, resp *rpctypes.Peers
...
+		Symbol:      "ETH",
+	}
+
+	for _, p := range peerIDs {
+		msg, err := s.net.Query(p)
+		if err != nil {
+			log.Debugf("Failed to query peer ID %s", p)
+			continue
+		}
+		if len(msg.Offers) > 0 {
```
noot, line 120: nit for readability, but can you change this to:
```
if len(msg.Offers) == 0 {
    continue
}

// the code that's currently in this block here
```

rpc/net.go
```diff
@@ -73,6 +88,71 @@ func (s *NetService) Peers(_ *http.Request, _ *interface{}, resp *rpctypes.Peers
...
+			for _, o := range msg.Offers {
+				address := o.EthAsset.Address()
+				index, exists := addrs[address]
+				pair := types.NewPair(o.EthAsset)
+				if !exists {
+					addrs[address] = index
```
noot, line 126: if the value does not exist in the map, isn't `index` always 0? this seems off, can you simplify this logic to just have one map `pairs := make(map[ethcommon.Address]*types.Pair)` where the key is the ETH-asset?

common/rpctypes/types.go
```diff
@@ -147,3 +147,11 @@ type AddressesResponse struct {
 type PeersResponse struct {
 	Addrs []string `json:"addresses" validate:"dive,required"`
 }
+
+// PairsRequest ...
+type PairsRequest = DiscoverRequest
```
noot, line 152: since the `Provides` field of the request is always set to `""`, can you define this as a new type:
```
type PairsRequest struct {
    SearchTime uint64 `json:"searchTime"` // in seconds
}
```

noot (approved):

PR #502, opened 2023-07-11
made optimism endpoint configurable and changed default

dimalinux:
The initial Optimism endpoint that I picked, `https://1rpc.io/op`, has degraded. It now, on a semi-regular basis, just returns an EOF error. (Worse than other endpoints, that give a message about being over the rate limit or a connection-refused error.)

This PR picks a different endpoint, from `blockpi`, and also makes the optimism endpoint configurable via an `ETH_OPTIMISM_ENDPOINT` environment variable. The naming of the variable was picked to match the unit test variables for Sepolia and Mainnet, and this new variable can also be set in the github secrets for CI.

Note that this variable name does not match the CLI environment variables that can be used in place of a CLI flag. If we don't end up finding a reliable endpoint, we may want to add full CLI support for setting the endpoint and using a variable name like `SWAPD_OPTIMISM_ENDPOINT`.

This PR also adds a fix that is needed due to `prettier` changes in the new 3.0.0 release.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,10 +21,8 @@ jobs:
           go-version: '1.20.x'
       - uses: actions/checkout@v3
 
-      # Temporarily locking prettier at 2.8.8. See:
-      # https://github.com/AthanorLabs/atomic-swap/issues/501
       - name: Install npm formatting/linting tools
-        run: npm install --global --save-dev prettier@2.8.8 prettier-plugin-solidity solhint
+        run: npm install --global --save-dev prettier prettier-plugin-solidity solhint
 
       # shellcheck should already be installed, but it doesn't hurt
       - name: Install apt formatting/linting tools
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,8 +21,10 @@ jobs:
           go-version: '1.20.x'
       - uses: actions/checkout@v3
 
+      # Temporarily locking prettier at 2.8.8. See:
+      # https://github.com/AthanorLabs/atomic-swap/issues/501
       - name: Install npm formatting/linting tools
-        run: npm install --global --save-dev prettier prettier-plugin-solidity solhint
+        run: npm install --global --save-dev prettier@2.8.8 prettier-plugin-solidity solhint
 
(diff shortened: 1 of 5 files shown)


.github/workflows/checks.yml
```diff
@@ -21,8 +21,10 @@ jobs:
           go-version: '1.20.x'
       - uses: actions/checkout@v3
 
+      # Temporarily locking prettier at 2.8.8. See:
+      # https://github.com/AthanorLabs/atomic-swap/issues/501
       - name: Install npm formatting/linting tools
-        run: npm install --global --save-dev prettier prettier-plugin-solidity solhint
+        run: npm install --global --save-dev prettier@2.8.8 prettier-plugin-solidity solhint
```
dimalinux, line 27: I didn't update the documentation. The Solidity plugin [has an open PR](https://github.com/prettier-solidity/prettier-plugin-solidity/pull/848) that presumably will fix the issue. They just need to merge it and make a release. Given that, I think we can remove the version peg before anyone will need to modify the Solidity code.

dimalinux: Thanks. I implemented it.

Makefile
```diff
@@ -1,4 +1,5 @@
 GOPATH ?= $(shell go env GOPATH)
+NPM_DIR = $(shell npm config get prefix)
```
dimalinux, line 2: Since we are only using `=`, not `:=`, the variable won't get evaluated if it is not used. You can still safely use most of the Makefile targets even if you don't have `npm` installed.

Makefile
```diff
@@ -30,7 +31,8 @@ format-shell:
 
 .PHONY: format-solidity
 format-solidity:
-	"$$(npm config get prefix)/bin/prettier" --print-width 100 --write \
+	"$(NPM_DIR)/bin/prettier" --print-width 100 --write \
+		--plugin="$(NPM_DIR)/lib/node_modules/prettier-plugin-solidity/src/index.js" \
```
dimalinux, line 35: With the 3.0.0 release of `prettier`, plugins are no longer loaded by default. You have to use this not-so-pretty flag or use a configuration file.

noot (approved):

PR #503, opened 2023-08-15
maintenance updates

dimalinux:
* updates to default monero nodes
* a bootnode had an IP change
* dependency updates
* linter update


.golangci.yml
```diff
@@ -53,7 +53,6 @@ linters:
   enable:
     - bidichk
     - bodyclose
-    - depguard
```
dimalinux, line 56: We would need to configure the package to leave it in, but until we have some concrete dependencies that we want to ban (or go to a strict whitelisting system), the plugin isn't helpful.

coins/test_support.go
```diff
@@ -30,7 +30,7 @@ func StrToDecimal(amount string) *apd.Decimal {
 // StrToETHAsset converts a string into the EthAssetAmount. Pass nil for the
 // token if the asset is ETH.
 func StrToETHAsset(amount string, optionalToken *ERC20TokenInfo) EthAssetAmount {
-	return NewEthAssetAmount(StrToDecimal(amount), nil)
+	return NewEthAssetAmount(StrToDecimal(amount), optionalToken)
```
dimalinux, line 33: Noticed the bug when it got flagged as an unused parameter by the linter. All callers are currently passing `nil`, so it didn't really matter.

common/config.go
```diff
@@ -65,21 +65,17 @@ func MainnetConfig() *Config {
 		EthEndpoint: "", // No mainnet default (permissionless URLs are not reliable)
 		MoneroNodes: []*MoneroNode{
 			{
-				Host: "node.sethforprivacy.com",
```
dimalinux, line 68: The sethforprivacy nodes, both stagenet and mainnet, have been offline for long enough that I don't think they are comming back up.

monero/wallet_client.go
```diff
@@ -43,20 +43,20 @@ const (
 // WalletClient represents a monero-wallet-rpc client.
 type WalletClient interface {
 	GetAccounts() (*wallet.GetAccountsResponse, error)
-	GetAddress(idx uint64) (*wallet.GetAddressResponse, error)
+	GetAddress(idx uint32) (*wallet.GetAddressResponse, error)
```
dimalinux, line 46: I did [a PR](https://github.com/MarinX/monerorpc/pull/18) on MarinX's repo to fix some features we don't use for the atomic swaps and adjusted these to match what `monero-wallet-rpc` is using to represent them on the server side. Matching them here too in order to update to the latest release.

tests/ganache.go
```diff
@@ -165,7 +165,7 @@ func getPackageKeys(t *testing.T, packageName string) []string {
 		startIndex += pkg.numKeys
 	}
 	t.Fatalf("Package %q does not have reserved test keys", packageName)
-	panic("unreachable code")
+	return nil // unreachable
```
dimalinux, line 168: Newer linter complained that the panic was unreachable, but was okay with returning `nil`.

noot (approved):
looks good!

PR #507, opened 2023-10-21
go version 1.21, libp2p and other updates, linter update

dimalinux:
Updates to go 1.21, which libp2p wasn't yet ready for in the last maintenance update. As part of the lib2p2 updates, we no longer use a persistent peer store. Linter is updated to v1.54.2.


protocol/swap/swap_manager.go
```diff
@@ -188,8 +188,8 @@ func (m *manager) GetOngoingSwapOfferIDs() ([]*types.Hash, error) {
 	defer m.RUnlock()
 
 	offerIDs := make([]*types.Hash, 0, len(m.ongoing))
-	for _, s := range m.ongoing {
-		offerIDs = append(offerIDs, &s.OfferID)
+	for i := range m.ongoing {
+		offerIDs = append(offerIDs, &m.ongoing[i].OfferID)
 	}
```
dimalinux, line 193: This change was only to make the linter happy. The original code looks safe to me.

rpc/server.go
```diff
@@ -150,20 +150,19 @@ func NewServer(cfg *Config) (*Server, error) {
...
 	r := mux.NewRouter()
 	r.Handle("/", rpcServer)
 	r.Handle("/ws", wsServer)
+
 	if !isBootnode {
+		reg, err := NewPrometheusRegistry()
+		if err != nil {
+			return nil, err
+		}
+		SetupMetrics(serverCtx, reg, cfg.Net, cfg.ProtocolBackend, cfg.XMRMaker)
```
dimalinux, line 162: This was just narrowing the scope NewPrometheusRegistry so bootnodes never create it and then not use it.

At the time I was reviewing this code, I was trying to figure out why the integration test bootnode was listening on a random 0.0.0.0 port. It turned out to be a combination of the latest libp2p code and the fact that we didn't configure the bootnode in the integration tests to not connect to the default bootnodes.

scripts/run-integration-tests.sh
```diff
@@ -43,6 +43,7 @@ CHARLIE_RPC_PORT=5002
 start-bootnode() {
 	local flags=(
 		"--rpc-port=${BOOTNODE_RPC_PORT}"
+		"--bootnodes="
```
dimalinux, line 46: This was the needed change that was holding up the pull request. The integration test bootnode was communicating with our production bootnodes. (Remember that bootnodes no longer have the concept of an environment.) We are setting the libp2p IP below to localhost which wouldn't be able to connect to the public bootnodes, and I don't believe the connections were happening with the older libp2p versions.

go.mod
```diff
@@ -1,111 +1,110 @@
...
-	github.com/athanorlabs/go-p2p-net v0.3.0
+	github.com/athanorlabs/go-p2p-net v0.3.1-0.20231210033046-13ef86a8318f
 	github.com/btcsuite/btcd/btcutil v1.1.3
-	github.com/cockroachdb/apd/v3 v3.2.0
+	github.com/cockroachdb/apd/v3 v3.2.1
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0
-	github.com/ethereum/go-ethereum v1.12.2
-	github.com/fatih/color v1.15.0
-	github.com/go-playground/validator/v10 v10.15.0
+	github.com/ethereum/go-ethereum v1.13.4
```
dimalinux, line 15: There is a v1.13.5 of go-ethereum, but it is adversely affecting unit test `TestSwapCreator_Refund_afterT2`.  This PR is for libp2p updates, so I didn't look into why.

go.mod
```diff
@@ -1,111 +1,110 @@
...
+go 1.21
 
 require (
 	filippo.io/edwards25519 v1.0.0
 	github.com/ChainSafe/chaindb v0.1.6
 	github.com/MarinX/monerorpc v1.0.7
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/athanorlabs/go-dleq v0.1.0
-	github.com/athanorlabs/go-p2p-net v0.3.0
+	github.com/athanorlabs/go-p2p-net v0.3.1-0.20231210033046-13ef86a8318f
```
dimalinux, line 11: Will update to a tagged release before merging.

noot: approved the go-p2p-net PR!

rpc/server.go
noot: bootnodes should probably have metrics as well - however yeah in the previous logic, that was not the case. maybe we can re-add in another PR

noot (approved):

